Issue #5, opened 2015-01-26
onProgress has no state at the end of a download


shaunmulligan:
The onProgress callback from `resin.models.os.download` doesn't return `state` when it reaches the end of the download.

Using the example `resin.models.os.download` code:

```
parameters =
		network: 'ethernet'
		appId: 91

resin.models.os.download parameters, '/opt/os.zip', (error) ->
		throw error if error?
	, (state) ->
		console.log "Total: #{state.total}"
		console.log "Received: #{state.received}"
```
gives the error:

```
Total: 107564591
Received: 106094176
Total: 107564591
Received: 106946144

/Users/hobochild/Documents/resin/autoTestPi/app.coffee:41
          console.log("Total: " + state.total);
                                       ^
TypeError: Cannot read property 'total' of undefined
  at Request.<anonymous> (/Users/hobochild/Documents/resin/autoTestPi/app.coffee:27:38)
  at Request.emit (events.js:117:20)
  at IncomingMessage.<anonymous> (/Users/hobochild/Documents/resin/autoTestPi/node_modules/resin-sdk/node_modules/request/request.js:1266:12)
  at IncomingMessage.emit (events.js:117:20)
  at _stream_readable.js:938:16
  at process._tickCallback (node.js:419:13)
```

jviotti:
An easy fix is to add `return if not state?` at the beginning of the `onProgress()` callback, however we should prevent `onProgress()` to get called with no state in the last call.

jviotti:
The issue is fixed now. `return if no state?` is not necessary anymore.

PR #9, opened 2015-02-04
Make sure status is never undefined in onProgress server callback #5

jviotti:


lib/server.coffee
```diff
@@ -111,7 +113,6 @@ exports.request = (options = {}, outerCallback, onProgress) ->
 				progress(request(options))
 					.on('progress', onProgress)
 					.on('error', outerCallback)
-					.on('end', onProgress)
```
Page-: Is it worth it/useful to still expose the `end` event?

Page-:
Is there anything that can be done about the docs update noise?  It makes these PRs a lot harder to look at since there's a huge number of trivially changed files

jviotti:
@Page- I agree as well. JSDoc seems to update every file at every change (it's frustrating when developing as well, as it makes very difficult to commit part of my changes). It seems JSDoc updated the footer of every file with the new date. I'll investigate if I can prevent that from happening.

lib/server.coffee
jviotti: @Page- I've did some testing, at `close` handles all cases. Having the `end` listener call `onProgress` results in `onProgress` having a `null` state argument, and calling it with `outerCallback` made such callback be called twice (by `end` and `close`).

Page-: Doesn't `close` get called on `error` as well, which would result in a double call there?

jviotti: After some experimentation, `end` is not even being called by `node-request-progress`. Also by checking their example, they use `close` to determine the final callback, so I think we're good like it is (I wasn't able to reproduce a case where `outerCallback` was called twice neither).

Issue #7, opened 2015-01-29
Resin OS Download state


jviotti:
Extend `resin.models.os.download()` state to provide `eta`, `remaining`, etc.

Ideally, we can create a `ProgressState` class that provides an uniform interface to the state, and we can then make Resin CLI Visuals progress bar accept a `ProgressState` as an argument.

jviotti:
The implementation of the ProgressState class is complete. I'll move the Resin CLI Visuals progress bar accept a `ProcessState` as an argument to the relevant repository.

PR #10, opened 2015-02-04
Abstract progress state interface with ProgressState class

jviotti:
Partially completes #7.


lib/progress-state.coffee
```diff
@@ -0,0 +1,132 @@
...
+#		.on('progress', ProgressState.createFromNodeRequestProgress(onProgressCallback))
+#		.on('error', ...)
+#		.on('close', ...)
+###
+ProgressState.createFromNodeRequestProgress = (callback) ->
+	time = getCurrentTime()
+	received = 0
+
+	return (state) ->
+		newTime = getCurrentTime()
```
Page-: I'd change this to:
```coffee-script
newTime = getCurrentTime()
timeDelta = newTime - time
time = newTime
```
in order to keep the whole delta/update logic together, I was looking for other references to the old `time` in between and had to double check I didn't miss any.

jviotti: It indeed increases readability. Fixed in https://github.com/resin-io/resin-sdk/commit/16d6049fdad45314ae7579b3eb054d644e134c30.

Page-: Thanks

PR #11, opened 2015-02-05
Create a standardised set of Error subclasses

### LINKED ISSUE #6: Create a standardised set of Error subclasses
Use them throughout the SDK and document them accordingly.

jviotti:
- Use them in trough the SDK.
- Includes documentation.

Closes #6.


lib/errors.coffee
```diff
@@ -1,165 +1,956 @@
...
+
+		###*
+		# @name message
+		# @type String
+		# @memberof errors.ResinInvalidDeviceType
+		# @instance
+		# @constant
+		# @default Invalid device type: this.type
+		###
+		@message = "Invalid device type: #{@type}"
```
Page-: You should change this to `super("Invalid device type: #{@type}")` so that you call the `typed-error` constructor (which handles stack trace construction and the like) - when you have no explicit constructor then it defaults to just calling the parent constructor with all arguments.

jviotti: @Page- Please take a look at https://github.com/resin-io/resin-sdk/pull/13.

PR #13, opened 2015-02-06
Call TypedError constructor with the message

jviotti:
As noted in https://github.com/resin-io/resin-sdk/pull/11#discussion_r24239432.


Page-:
LGTM

PR #33, opened 2015-03-20
Reference applications and devices by name instead of id

jviotti:


lib/models/device.coffee
```diff
@@ -96,23 +130,46 @@ exports.getAllByApplication = (applicationId, callback) ->
...
+
+	username = token.getUsername()
+
+	if not username?
+		return callback(new errors.ResinNotLoggedIn())
+
 	return pine.get
 		resource: 'device'
-		id: deviceId
+		name: name
```
Page-: This needs to be in a filter

lib/models/device.coffee
```diff
@@ -131,17 +188,33 @@ exports.get = (deviceId, callback) ->
...
+	if not _.isFunction(callback)
+		throw new errors.ResinInvalidParameter('callback', callback, 'not a function')
+
+	if not token.getUsername()?
+		return callback(new errors.ResinNotLoggedIn())
+
 	return pine.delete
 		resource: 'device'
-		id: id
+		name: name
```
Page-: This needs to be in a filter

lib/models/device.coffee
```diff
@@ -193,12 +282,34 @@ exports.identify = (uuid, callback) ->
...
+	if not _.isFunction(callback)
+		throw new errors.ResinInvalidParameter('callback', callback, 'not a function')
+
+	if not token.getUsername()?
+		return callback(new errors.ResinNotLoggedIn())
+
 	return pine.patch
 		resource: 'device'
-		id: id
+		name: name
```
Page-: This needs to be in a filter

lib/models/device.coffee
```diff
@@ -212,19 +323,41 @@ exports.rename = (id, name, callback) ->
...
```
Page-: This needs to be in a filter

lib/models/device.coffee
jviotti: Fixed.

lib/models/device.coffee
jviotti: Fixed.

lib/models/device.coffee
jviotti: Fixed.

jviotti:
@Page- Can you re check that the filters are applied correctly on the places you commented please?

PR #35, opened 2015-04-27
Check that device uuid is valid in logs

jviotti:


jviotti:
Fixes https://github.com/resin-io/resin-cli/issues/13.

PR #36, opened 2015-04-27
Implement resin.models.application.has()

jviotti:
This will be used to fix https://github.com/resin-io/resin-cli/issues/30.

### DIFF
--- a/tests/models/application.spec.coffee
+++ b/tests/models/application.spec.coffee
@@ -192,6 +192,92 @@ describe 'Application Model:', ->
 					expect(application).to.not.exist
 					done()
 
+	describe '.has()', ->
+
+		it 'should throw if no name', ->
+			expect ->
+				application.has(null, _.noop)
+			.to.throw(errors.ResinMissingParameter)
+
+		it 'should throw if name is not a string', ->
+			expect ->
+				application.has([ 'MyApp' ], _.noop)
+			.to.throw(errors.ResinInvalidParameter)
+
+		it 'should throw if not callback', ->
+			expect ->
+				application.has('MyApp', null)
+			.to.throw(errors.ResinMissingParameter)
+
+		it 'should throw if callback is not a function', ->
+			expect ->
+				application.has('MyApp', [ _.noop ])
+			.to.throw(errors.ResinInvalidParameter)
+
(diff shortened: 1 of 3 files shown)


tests/models/application.spec.coffee
```diff
@@ -192,6 +192,92 @@ describe 'Application Model:', ->
...
+				it 'should return false', (done) ->
+					application.has 'MyApp', (error, has) ->
+						expect(error).to.not.exist
+						expect(has).to.be.false
+						done()
+
+			describe 'given an application', ->
+
+				beforeEach ->
+					@application =
```
emirotin: why `@` here?

tests/models/application.spec.coffee
```diff
@@ -192,6 +192,92 @@ describe 'Application Model:', ->
...
+						__metadata: { uri: '/ewa/application(999)', type: '' }
+
+					@pineGetStub = sinon.stub(pine, 'get')
+					@pineGetStub.returns(Promise.resolve([ @application ]))
+
+				afterEach ->
+					@pineGetStub.restore()
+
+				it 'should return true', (done) ->
+					application.has 'MyApp', (error, has) ->
```
emirotin: does it pass? the name in the stub is `App1`, and here you search for `MyApp`

jviotti: @emirotin Yeah, sinon ignores the actual argument to the function, so it will return the `App1` object for any string here.

tests/models/application.spec.coffee
jviotti: I copy-pasted a chunk from the application.get tests, and this was used in the `it` clause. Safe to remove here.

tests/models/application.spec.coffee
emirotin: Ah I see. But seems that at least you should return a single object instead of an arrayfrom the stubbed method?

emirotin:
Besides the purist note lgtm

tests/models/application.spec.coffee
jviotti: @emirotin It should, ideally, but the API returns an array containing a single application instead of the object directly, so I mimic it here.

PR #40, opened 2015-05-04
Add resin.models.device.getByUUID

pcarranzav:
This PR adds a method to get a device by specifying UUID instead of name. Especially useful for apps that rely on the fact that devices know the UUID via env vars.
Not sure if the camelCasing is correct (getByUUID or getByUuid?).

### DIFF
--- a/lib/models/device.coffee
+++ b/lib/models/device.coffee
@@ -183,6 +183,58 @@ exports.get = (name, callback) ->
 	.nodeify(callback)
 
 ###*
+# @summary Get a single device by UUID
+# @public
+# @function
+#
+# @param {String} uuid - device UUID
+# @param {module:resin.models.device~getCallback} callback - callback
+#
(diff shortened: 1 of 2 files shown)


lib/models/device.coffee
```diff
@@ -183,6 +183,58 @@ exports.get = (name, callback) ->
 	.nodeify(callback)
 
 ###*
+# @summary Get a single device by UUID
+# @public
+# @function
+#
+# @param {String} uuid - device UUID
+# @param {module:resin.models.device~getCallback} callback - callback
```
jviotti: I think you forgot to annotate the callback. This is still referencing `getCallback` and there is no callback annotation above.

jviotti:
@pcarranzav Looks great, the camel casing is good as well, I'll merge once the callback annotation issue is addressed.

PR #44, opened 2015-05-12
Refactor Keys model to use PineJS

jviotti:
This PR should be tested and merged once the corresponding backend changes are in production.

### DIFF
--- a/lib/models/key.coffee
+++ b/lib/models/key.coffee
@@ -2,16 +2,16 @@
 # @module resin.models.key
 ###
 
-_ = require('lodash-contrib')
+_ = require('lodash')
 errors = require('resin-errors')
-request = require('resin-request')
+token = require('resin-token')
+pine = require('resin-pine')
+auth = require('../auth')
 
 ###*
 # A Resin API key
 # @typedef {Object} Key
 ###
-
-# TODO: Do this with pinejs once it's exposed as an OData API
 
 ###*
 # getAll callback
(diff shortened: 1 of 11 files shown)


lib/models/key.coffee
```diff
@@ -33,16 +32,23 @@ request = require('resin-request')
...
+	username = token.getUsername()
+
+	if not username?
+		return callback(new errors.ResinNotLoggedIn())
+
+	return pine.get
+		resource: 'user__has__public_key'
+		options:
+			filter:
+				user: { username }
```
emirotin: why not by user ID?

jviotti: @emirotin Much better. Do you mind taking another look?

lib/models/key.coffee
```diff
@@ -33,16 +33,21 @@ request = require('resin-request')
...
 		return callback(error) if error?
 
-		if _.isEmpty(keys)
-			return callback(new errors.ResinNotAny('keys'))
+		return pine.get
+			resource: 'user__has__public_key'
+			options:
+				filter:
+					user: { id }
+		.then (keys) ->
```
Page-: If you use `.tap` then you don't need the `return keys` below

lib/models/key.coffee
```diff
@@ -65,15 +70,22 @@ exports.getAll = (callback) ->
...
+		return pine.get
+			resource: 'user__has__public_key'
+			id: id
+			options:
+				filter:
+					user: { id: userId }
+		.then (key) ->
+			if _.isEmpty(key)
+				throw new errors.ResinKeyNotFound(id)
+			return _.first(key)
```
Page-: You could replace this with:
```
.tap (key) ->
	if _.isEmpty(key)
		throw new errors.ResinKeyNotFound(id)
.get(0)
```

lib/models/key.coffee
jviotti: Fixed.

lib/models/key.coffee
jviotti: Fixed.

Page-:
LGTM

PR #50, opened 2015-05-26
Support device environment variables

jviotti:


lib/models/environment-variables.coffee
```diff
@@ -143,3 +143,126 @@ exports.remove = (id, callback) ->
...
+	return pine.get
+		resource: 'device_environment_variable'
+		options:
+			filter:
+				device: deviceId
+			expand: 'device'
+			orderby: 'env_var_name asc'
+	.tap (environmentVariables) ->
+		if _.isEmpty(environmentVariables)
+			throw new errors.ResinNotAny('device environment variables')
```
Page-: I don't know if this is the standard way of handling an empty list in the SDK, but I'd expect to just get an empty array returned since I'm asking for all the ones that exist - whereas if I was asking for one specific instance then I'd expect an error for it to be missing (since I'm looking for that particular one by name/id)

lib/models/environment-variables.coffee
```diff
@@ -143,3 +143,126 @@ exports.remove = (id, callback) ->
...
+# @param {String} name - environment variable name
+# @param {String} value - environment variable value
+# @param {module:resin.models.environment-variables.device~createCallback} callback - callback
+#
+# @example
+#	resin.models.environmentVariables.device.create 51, 'EDITOR', 'vim', (error) ->
+#		throw error if error?
+###
+exports.device.create = (deviceId, name, value, callback) ->
+	return pine.post
```
jviotti: I'm getting 500 here:

```sh
POST https://api.resin.io/ewa/device_environment_variable -> 500
```

Any ideas?

Page-: @emirotin ?

lib/models/environment-variables.coffee
jviotti: I guess is debatable, but I'll keep the current behaviour since all the other SDK "getAll" functions behave the same way, but definitely something to consider in the future.

lib/models/environment-variables.coffee
jviotti: Fixed. I needed to use `body` instead of `data`.

emirotin: Was writing this comment :) It's a difference in PineJS client version afaik

jviotti: @emirotin :) Thanks for your help. I implement support for "data" as well in resin-pine, to avoid a similar issue in the future: https://github.com/resin-io/resin-pine/pull/3.

Page-:
LGTM

emirotin:
lgtm2

PR #51, opened 2015-05-28
Accept device names instead of ids in device env vars

jviotti:


lib/models/environment-variables.coffee
```diff
@@ -158,26 +159,29 @@ exports.device = {}
...
-			filter:
-				device: deviceId
-			expand: 'device'
-			orderby: 'env_var_name asc'
-	.tap (environmentVariables) ->
-		if _.isEmpty(environmentVariables)
-			throw new errors.ResinNotAny('device environment variables')
-	.nodeify(callback)
+exports.device.getAll = (deviceName, callback) ->
+	deviceModel.get deviceName, (error, device) ->
```
Page-: Can this be:
```coffee
deviceModel.get(deviceName)
.then (device) ->
	pine.get
		...
.tap ...
.nodeify(callback)
```

lib/models/environment-variables.coffee
```diff
@@ -190,23 +194,26 @@ exports.device.getAll = (deviceId, callback) ->
...
-exports.device.create = (deviceId, name, value, callback) ->
-	return pine.post
-		resource: 'device_environment_variable'
-		body:
-			device: deviceId
-			env_var_name: name
-			value: value
-	.nodeify(callback)
+exports.device.create = (deviceName, name, value, callback) ->
+	deviceModel.get deviceName, (error, device) ->
```
Page-: Same here

lib/models/environment-variables.coffee
jviotti: @Page- . Done, can you check again please?

lib/models/environment-variables.coffee
jviotti: Done as well.

lib/models/environment-variables.coffee
```diff
@@ -158,22 +159,23 @@ exports.device = {}
...
-exports.device.getAll = (deviceId, callback) ->
-	return pine.get
-		resource: 'device_environment_variable'
-		options:
-			filter:
-				device: deviceId
-			expand: 'device'
-			orderby: 'env_var_name asc'
+exports.device.getAll = (deviceName, callback) ->
+	deviceModel.get(deviceName).then (device) ->
```
Page-: Is it possible to ask for just specific field in `.get` to save on traffic?  (particularly noticeable for `.getAll`s)

Page-:
LGTM

lib/models/environment-variables.coffee
jviotti: @Page- `.get` still needs to be able to get the whole device as it's a public function. The only thing I can think of is implementing a `.getIDByDeviceName()` function or something like that, and reuse that one in these two functions. What do you think?

Page-: I was thinking of an optional array parameter to specify the files you want, ie:
```coffee
deviceModel.get(deviceName, ['id'])
```
```coffee
deviceModel.get(deviceName, ['id', 'name])
```
or potentially:
```coffee
deviceModel.get(deviceName, 'id')
```
```coffee
deviceModel.get(deviceName, 'id', 'name')
```
however whilst the second version is nicer to type, it would be harder to distinguish since you already have potentially multiple parameters like that for filtering right?  
Could also maybe do it via an options object (which would be usable for other things as well in the future):
```coffee
deviceModel.get(deviceName, select: 'id')
```
```coffee
deviceModel.get(deviceName, select: ['id', 'name'])
```

jviotti: The idea is cool, however I think it makes there interface a little bit complex, we will have to support this in the rest of the functions (which would be a major change) and I guess users will not bother to select the exact things they need (and it implies the users know the name of the property in advance), and therefore will be a feature we will be using only internally. 

I suggest implementing a custom function that only returns the id (we can keep it private as well), or wait until the SDK becomes too network consuming, so we don't optimise prematurely.

Page-: I'm fairly sure if someone wanted to use the SDK to query the status of 10,000 devices that they'd appreciate being able to specify just the fields they're interested in.  Also it would be an optional thing so the default interface would remain the same, as for implementing it could be done using shared code to minimise the overhead, eg:
```
defaults = (options, callback) ->
	if _.isFunction(options)
		return { callback: options, options: {} }
	return { callback, options }

optionedGet = (options, query) ->
	if options?.select?
		if query.options?.select?
			throw new Error('Cannot specify select for this command')
		query.options ?= {}
		query.options.select = options.select
	return pine.get(query)

exports.device.getAll = (deviceName, options, callback) ->
	{ options, callback } = defaults(options, callback)
	deviceModel.get(deviceName).then (device) ->
		optionedGet options,
			resource: 'device_environment_variable'
			options:
				filter:
					device: device.id
				expand: 'device'
				orderby: 'env_var_name asc'
```

jviotti: Yeah, the benefits are more apparent when thinking about that many devices. Can you move your comment to a GitHub issue? I'll merge this PR as is and will send a new one with this feature soon.

PR #52, opened 2015-06-01
Expose promise interface in auth module

jviotti:
This introduces a small breaking change:

- auth.authenticate() doesn't yield a username anymore.

### DIFF
--- a/lib/auth.coffee
+++ b/lib/auth.coffee
@@ -2,8 +2,9 @@
 # @module resin.auth
 ###
 
+Promise = require('bluebird')
 errors = require('resin-errors')
-request = require('resin-request')
+request = Promise.promisifyAll(require('resin-request'))
 token = require('resin-token')
 
 ###*
@@ -32,19 +33,13 @@ token = require('resin-token')
 #			console.log("My username is: #{username}")
 ###
 exports.whoami = (callback) ->
-	try
-		username = token.getUsername()
-	catch error
-		return callback(error)
-
-	return callback(null, username)
+	Promise.try(token.getUsername).nodeify(callback)
 
 ###*
 # authenticate callback
 # @callback module:resin.auth~authenticateCallback
 # @param {(Error|null)} error - error
 # @param {String} token - session token
-# @param {String} username - username
 ###
 
 ###*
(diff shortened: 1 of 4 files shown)


lib/auth.coffee
```diff
@@ -64,20 +59,19 @@ exports.whoami = (callback) ->
...
-	request.request
-		method: 'POST'
-		url: '/login_'
-		json: credentials
-	, (error, response) ->
-		return callback(error) if error?
-		savedToken = response?.body
-		return callback(null, savedToken, credentials.username)
+	Promise.fromNode (callback) ->
+		request.request
```
emirotin: I think you can `promisifyAll` request instead

Page-: You can, and then `request.postAsync().spread (response, body) ->`

jviotti: Sadly  `.spread' did not work on this function:

```sh
TypeError: expecting an array, a promise or a thenable. See http://goo.gl/s8MMhc
```

So I had to use `.then` here, but it worked in `.register()`.

lib/auth.coffee
```diff
@@ -269,20 +258,10 @@ exports.logout = (callback) ->
...
-
-	request.request
+	request.requestAsync
 		method: 'POST'
 		url: '/user/register'
 		json: credentials
-	, (error, response, body) ->
-		return callback(error) if error?
-		return callback(null, body)
+	.spread (response, body) ->
```
Page-: This can be `.get(1)`

lib/auth.coffee
```diff
@@ -64,20 +59,18 @@ exports.whoami = (callback) ->
...
-	request.request
+	request.requestAsync
 		method: 'POST'
 		url: '/login_'
 		json: credentials
-	, (error, response) ->
-		return callback(error) if error?
-		savedToken = response?.body
-		return callback(null, savedToken, credentials.username)
+	.then (data) ->
```
Page-: This can be `.get('body')`

lib/auth.coffee
Page-: That is very weird, since they are both POSTs :S

lib/auth.coffee
jviotti: Fixed.

lib/auth.coffee
jviotti: Fixed.

Page-:
LGTM

PR #53, opened 2015-06-01
Expose promise interface in models/config module

jviotti:


lib/models/config.coffee
```diff
@@ -24,12 +25,13 @@ request = require('resin-request')
 #		console.log(config)
 ###
 exports.getAll = (callback) ->
-	request.request
-		method: 'GET'
-		url: '/config'
-	, (error, response, config) ->
-		return callback(error) if error?
-		return callback(null, config)
+	Promise.fromNode (callback) ->
```
Page-: It would be nicer to `Promise.promisifyAll(request)` and then use the async versions

jviotti: @Page- Fixed, can you have another look please?

lib/models/config.coffee
```diff
@@ -24,12 +25,12 @@ request = require('resin-request')
...
 ###
 exports.getAll = (callback) ->
-	request.request
+	request.requestAsync
 		method: 'GET'
 		url: '/config'
-	, (error, response, config) ->
-		return callback(error) if error?
-		return callback(null, config)
+	.spread (response, config) ->
```
Page-: This can be `.get(1)`

Page-:
LGTM

lib/models/config.coffee
jviotti: Fixed.

PR #56, opened 2015-06-08
Make use of promises in models.key

jviotti:


lib/models/key.coffee
```diff
@@ -33,19 +33,13 @@ auth = require('../auth')
...
-		return callback(error) if error?
-
+	auth.getUserId().then (id) ->
 		return pine.get
 			resource: 'user__has__public_key'
 			options:
 				filter:
 					user: { id }
-		.tap (keys) ->
-			if _.isEmpty(keys)
```
Page-: Is removing this check intentional?

jviotti: Yeah, I'm returning an empty array instead of throwing an error in `getAll` functions.

Page-:
LGTM

PR #57, opened 2015-06-08
Make use of promises in models.device

jviotti:


lib/models/device.coffee
```diff
@@ -2,10 +2,11 @@
 # @module resin.models.device
 ###
 
+Promise = require('bluebird')
 _ = require('lodash-contrib')
 pine = require('resin-pine')
 errors = require('resin-errors')
-request = require('resin-request')
+request = Promise.promisifyAll(require('resin-request'))
```
Page-: It'd be nice to not have to do the `promisifyAll`

lib/models/device.coffee
```diff
@@ -144,35 +114,25 @@ exports.getAllByApplication = (name, callback) ->
...
+					user: { username }
 
 	.then (device) ->
 		if _.isEmpty(device)
 			throw new errors.ResinDeviceNotFound(name)
-
-		device = _.first(device)
-
-		# TODO: Move to server
+		return device
```
Page-: Can just do `.tap (device) ->` and remove the need for this return

lib/models/device.coffee
```diff
@@ -144,35 +114,25 @@ exports.getAllByApplication = (name, callback) ->
...
+	.get(0)
+	.then (device) ->
```
Page-: Can also `.tap` here

lib/models/device.coffee
```diff
@@ -197,41 +157,25 @@ exports.get = (name, callback) ->
...
+					user: { username }
 
 	.then (device) ->
 		if _.isEmpty(device)
 			throw new errors.ResinDeviceNotFound(uuid)
-
-		device = _.first(device)
-
-		# TODO: Move to server
+		return device
```
Page-: Can `.tap` this

lib/models/device.coffee
```diff
@@ -197,41 +157,25 @@ exports.get = (name, callback) ->
...
+	.get(0)
+	.then (device) ->
 		device.application_name = device.application[0].app_name
-
 		return device
```
Page-: Can `.tap` this

lib/models/device.coffee
```diff
@@ -256,19 +200,11 @@ exports.getByUUID = (uuid, callback) ->
...
-	if not _.isFunction(callback)
-		throw new errors.ResinInvalidParameter('callback', callback, 'not a function')
-
-	exports.get name, (error) ->
-		if error instanceof errors.ResinDeviceNotFound
-			return callback(null, false)
-
-		return callback(error) if error?
-		return callback(null, true)
+	exports.get(name).then ->
```
Page-: This can use `.return(true)`

lib/models/device.coffee
```diff
@@ -291,16 +227,9 @@ exports.has = (name, callback) ->
...
-		throw new errors.ResinMissingParameter('callback')
-
-	if not _.isFunction(callback)
-		throw new errors.ResinInvalidParameter('callback', callback, 'not a function')
-
-	exports.get name, (error, device) ->
-		return callback(error) if error?
-		return callback(null, !!device.is_online)
+	exports.get(name).then (device) ->
+		return !!device.is_online
```
Page-: There shouldn't be any need for the `!!`

lib/models/device.coffee
jviotti: Yeah, I'll be refactoring each of the modules to use promises.

lib/models/device.coffee
jviotti: Fixed.

lib/models/device.coffee
jviotti: Fixed.

lib/models/device.coffee
jviotti: Fixed.

lib/models/device.coffee
jviotti: Fixed.

lib/models/device.coffee
jviotti: Fixed.

lib/models/device.coffee
jviotti: I remember needing this because I was getting `is_online: 1`, `is_online: 0`, as well as `is_online: true`, `is_online: false`. I can remove the `!!` if you confirm that we're using only integers or booleans now.

Page-: It should always give `true`/`false` (and the check I just did showed `true`/`false`), I did fix a bug ages ago where it would return `0`/`1` in a $expand though, which might be what you experienced on the ui at some point - if do you find any other cases where it happens then let me know

jviotti: @Page- Cool, I'll remove the check. I also noticed that `is_online: undefined` in some cases (I think it happened before the device went online for the first time). Is that still the case?

Page-: That should be impossible, it is never allowed to be null - at a guess what happened was you requested for a device that doesn't exist and tried getting the property of an empty object (which returned undefined)

Page-:
LGTM

PR #61, opened 2015-06-12
Upgrade Resin Request to v1.3.0

jviotti:


package.json
```diff
@@ -48,7 +48,7 @@
     "resin-errors": "^1.0.2",
     "resin-network-config": "^1.0.0",
     "resin-pine": "^1.1.1",
-    "resin-request": "^1.2.5",
+    "resin-request": "^1.3.0",
```
Page-: This was a breaking change so really should have been bumped to `2.0.0` - if you've published it to `1.3.0` then anyone installing with `^1.2.5` will be broken

PR #62, opened 2015-06-12
Use promises in os module

jviotti:


lib/models/os.coffee
```diff
@@ -35,29 +41,27 @@ OSParams = require('./os-params')
...
 ###
-exports.download = (parameters, destination, callback, onProgress) ->
+exports.download = (options, destination, callback) ->
+	Promise.try ->
+		if not token.getUsername()?
+			throw new errors.ResinNotLoggedIn()
 
-	if not token.getUsername()?
-		return callback(new errors.ResinNotLoggedIn())
+		osParams = _.omit(options, 'onProgress')
```
Page-: It would make more sense to whitelist the options you expect rather than blacklist specific ones you don't expect (because they could always pass an option that is invalid under any context, rather than a valid one that just isn't expected for this branch)

jviotti: @Page- `os-params` already whitelists the options and throws an error if you're passing an extra one: https://github.com/resin-io/resin-sdk/blob/master/lib/models/os-params.coffee#L62. Therefore the `omit` here is to prevent an error being thrown for `onProgress`.

Page-: This seems pretty awkward, it means every time resin-request support a new option it will have to be added to this blacklist manually, just a bit of a hassle to say the least

jviotti:
Closing this PR, as I'm making use of new promised `resin-request` which makes this change irrelevant.

PR #63, opened 2015-06-21
Move device register from resin-device-register to the sdk

jviotti:
It was unnecessary to have this functionality as a separate module, when a simpler interface could be exposed by the SDK.


abresas:
LGTM

PR #65, opened 2015-06-22
Make use of promised dependencies

jviotti:


lib/models/application.coffee
```diff
@@ -274,10 +248,11 @@ exports.remove = (name, callback) ->
 ###
 exports.restart = (name, callback) ->
 	exports.get(name).then (application) ->
-		return request.requestAsync
+		return request.send
 			method: 'POST'
 			url: "/application/#{application.id}/restart"
-	.nodeify(_.unary(callback))
+	.return(undefined)
```
abresas: why undefined?

tests/models/device.spec.coffee
```diff
@@ -14,276 +13,180 @@ describe 'Device Model:', ->
...
 
-				it 'should return false', (done) ->
-					device.has 'MyDevice', (error, hasDevice) ->
-						expect(error).to.not.exist
-						expect(hasDevice).to.be.false
-						done()
+			it 'should add application_name', (done) ->
+				device.get 'MyDevice', (error, device) =>
+					expect(device.application_name).to.equal(@device.application[0].app_name)
+					done()
```
abresas: I usually do .nodeify(done) on tests

abresas:
lgtm, left some notes

lib/models/application.coffee
jviotti: To prevent resolving the promise with the request response, just as a way to avoid polluting the API with unnecessary information.

tests/models/device.spec.coffee
jviotti: I plan to move the SDK tests to use `chai-as-promised` as the other modules, but plan to do it in another PR.

lib/auth.coffee
```diff
@@ -173,8 +170,7 @@ exports.isLoggedIn = (callback) ->
 #		console.log(token)
 ###
 exports.getToken = (callback) ->
-	Promise.try ->
-		savedToken = token.get()
+	token.get().then (savedToken) ->
```
Page-: This could be a `.tap`

lib/auth.coffee
```diff
@@ -201,8 +197,7 @@ exports.getToken = (callback) ->
 #		console.log(id)
 ###
 exports.getUserId = (callback) ->
-	Promise.try ->
-		id = token.getUserId()
+	token.getUserId().then (id) ->
```
Page-: This could be a `.tap`

PR #66, opened 2015-06-22
Remove JSDoc HTML generation

jviotti:
We will generate markdown documentation files like the other modules, which also makes it easy for adding to docs.resin.io.


abresas:
we cannot check every file here, I guess you can just merge

PR #70, opened 2015-06-23
Implement resin.settings

jviotti:


abresas:
code-wise lgtm, still trying to understand the feature

jviotti:
@abresas It exposes the settings object provided by https://github.com/resin-io/resin-settings-client by the SDK, so we don't encourage people to require low level modules directly.

PR #72, opened 2015-06-23
Make device.identify take a device name instead of a uuid

jviotti:


emirotin:
Just want to mention the duplicate names issue again. As the SDK is essentially the API I'm pretty sure it's ok for it to use reliable and low-level identifiers (UUIDs). The CLI on the other side should be user-friendly and operate on names, then find the UUID through the SDK and show an interactive menu if there are multiple, finally call the SDK identify function passing the proper UUID.

All IMO

jviotti:
@emirotin Sure, I made a note about discussing the usage of uuids/ids in the SDK, which is quite necessary given this issue, but wanted to merge the branch since it was blocking me with other things. This problem definitely will be solved before publishing next version to NPM.

emirotin:
Yeah, just mentioned it as I saw another usage of names

PR #73, opened 2015-06-23
Check that a device exist before trying to rename

jviotti:


lib/models/device.coffee
```diff
@@ -275,23 +275,20 @@ exports.identify = (uuid, callback) ->
...
 ###
 exports.rename = (name, newName, callback) ->
-	return pine.patch
-		resource: 'device'
-		body:
-			name: newName
-		options:
-			filter:
-				name: name
+	exports.get(name).then ->
```
emirotin: and get throws in case the device doesn't exist, right?

Also the generic question: device name is not unique. You can even have multiple device of the same name in one app.

jviotti: Right. Well, that's a major issue. I knew device names were not unique, but I thought there couldn't be duplicated devices per user.

In that case, the only reliable way to refer to a device would be with an id, which was how the SDK worked initially, and everybody complained.

The other solution I can think of is disallowing duplicated device names from within an application. That way we can refer to devices as `<application>/<device>`, in a Github manner, for example: `resin device MyApp/MyDevice`.

Thoughts?

emirotin: Makes sense, but we can potentially already have duplicates, I'm afraid. You can also switch to interactive list menu with UUIDs and links to the dashboard when you see multiple candidates?

emirotin: Ah, it's SDK...
How about the SDK (being the API) requires UUID, but the CLI (which is UI) uses names and falls back to the list menu in case of duplicates?

jviotti: It should work, however using uuids/ids in the SDK makes it very hard to use the SDK and it'd be much nicer to handle readable names instead. 

I could port the SDK to uuids, but that'd also mean going one step back, since there were various discussions about needing the SDK to handle names as well.

I wonder why no one noticed this on time.

emirotin: Just saw this comment. I don't think SDK has to support something user-friendly. It's a programming interface, right? Nobody's using it by hand. If you search for your devices through the API you can get UUIDs as easy as the name. If you have the name in your app you can still do the same as the CLI — search the API through the SDK and pick the desired device then.

jviotti: Does the duplication issue only happens on devices? Are users allowed to have multiple apps with the same names?

emirotin: It's prohibited for apps in a hook https://bitbucket.org/rulemotion/resin-api/src/413cdc75c4d372dd23b719c4427da3da04de1f02/src/models/hooks/index.coffee?at=master#cl-39

PR #81, opened 2015-06-29
Add basic unit tests for key model

jviotti:
Only function that have functionality other than the pine request are
tested.


tests/models/key.spec.coffee
```diff
@@ -0,0 +1,40 @@
+m = require('mochainon')
+Promise = require('bluebird')
```
jviotti: This is not being used in the tests.

PR #84, opened 2015-07-03
Implement models.device.getLocalIPAddresses()

jviotti:


lib/models/device.coffee
```diff
@@ -201,6 +201,30 @@ exports.isOnline = (uuid, callback) ->
...
+# @example
+# resin.models.device.getLocalIPAddress('7cf02a62a3a84440b1bb5579a3d57469148943278630b17e7fc6c4f7b465c9').then (localIPAddress) ->
+# 	console.log(localIPAddress)
+###
+exports.getLocalIPAddress = (uuid, callback) ->
+	exports.get(uuid).then (device) ->
+		if not device.is_online
+			throw new Error("The device is offline: #{uuid}")
+
+		return device.ip_address?.split(' ')[1]
```
Page-: This is not a sensible way to do it, there is no guarantee on the ordering of the ip addresses, nor that the device will only have 1 local ip address.  The better way to do this is to remove only the ip address that matches `vpn_address`

package.json
```diff
@@ -39,7 +39,7 @@
     "run-sequence": "^1.1.0"
   },
   "dependencies": {
-    "bluebird": "^2.9.27",
+    "bluebird": "2.9.30",
```
jviotti: I added a note to myself to remember unlocking this dependency once the issue is fixed: https://github.com/petkaantonov/bluebird/issues/689.

lib/models/device.coffee
jviotti: You're right. Fixed, please take a look.

build/models/device.js
```diff
@@ -237,6 +237,36 @@ THE SOFTWARE.
...
+   * @example
+   * resin.models.device.getLocalIPAddresses('7cf02a62a3a84440b1bb5579a3d57469148943278630b17e7fc6c4f7b465c9').then (localIPAddresses) ->
+   * 	for localIP in localIPAddresses
+   * 		console.log(localIP)
+   */
+
+  exports.getLocalIPAddresses = function(uuid, callback) {
+    return exports.get(uuid).then(function(device) {
+      var ips;
+      if (!device.is_online) {
```
Page-: This checks if it's online on the vpn, however the device can still have connectivity and be accessible without being on the vpn - although it's an edge case (and suffers reduced functionality).  The ui has the same issue with the difference not being the clearest though, so I'm not sure what we can do to limit the impact

build/models/device.js
```diff
@@ -237,6 +237,36 @@ THE SOFTWARE.
...
+        throw new Error("The device is offline: " + uuid);
+      }
+      ips = device.ip_address.split(' ');
+      return _.difference(ips, [device.vpn_address]);
```
Page-: It would be slightly cleaner as `_.without(ips, device.vpn_address)`

lib/models/device.coffee
```diff
@@ -201,6 +201,32 @@ exports.isOnline = (uuid, callback) ->
...
+# @throws Will throw if the device is offline.
+#
+# @example
+# resin.models.device.getLocalIPAddresses('7cf02a62a3a84440b1bb5579a3d57469148943278630b17e7fc6c4f7b465c9').then (localIPAddresses) ->
+# 	for localIP in localIPAddresses
+# 		console.log(localIP)
+###
+exports.getLocalIPAddresses = (uuid, callback) ->
+	exports.get(uuid).then (device) ->
+		if not device.is_online
```
Page-: This checks if it's online on the vpn, however the device can still have connectivity and be accessible without being on the vpn - although it's an edge case (and suffers reduced functionality). The ui has the same issue with the difference not being the clearest though, so I'm not sure what we can do to limit the impact

lib/models/device.coffee
```diff
@@ -201,6 +201,32 @@ exports.isOnline = (uuid, callback) ->
...
+			throw new Error("The device is offline: #{uuid}")
+
+		ips = device.ip_address.split(' ')
+		return _.difference(ips, [ device.vpn_address ])
```
Page-: It would be slightly cleaner as _.without(ips, device.vpn_address)

jviotti: Fixed!

lib/models/device.coffee
jviotti: I think we shouldn't worry about that edge case. `is_online` here means that the device is online with Resin.io, more than the device is online on it's own.

Page-: Yeah, it's just that it can be confusing to users (and the ip address reported will be correct in the case of having connectivity but no vpn)

Page-:
LGTM

PR #85, opened 2015-07-03
Fix error that caused has promise to be fulfilled with an empty object

jviotti:


jviotti:
The bug was fixed by the bluebird team.

PR #92, opened 2015-07-16
Accept uuids instead of device names in logs module

jviotti:


lib/logs.coffee
```diff
@@ -40,29 +40,26 @@ deviceModel = require('./models/device')
...
 # 		console.log(line)
 ###
-exports.subscribe = (deviceName, callback) ->
-	Promise.props
-		uuid: deviceModel.get(deviceName).get('uuid')
-		pubNubKeys: configModel.getPubNubKeys()
-	.then (results) ->
-		return logs.subscribe(results.pubNubKeys, results.uuid)
+exports.subscribe = (uuid, callback) ->
+	deviceModel.get(uuid).then(configModel.getPubNubKeys).then (pubNubKeys) ->
```
Page-: There's no need for the `deviceModel.get(uuid)`?

lib/logs.coffee
```diff
@@ -72,24 +69,21 @@ exports.subscribe = (deviceName, callback) ->
...
 # 		console.log(line)
 ###
-exports.history = (deviceName, callback) ->
-	Promise.props
-		uuid: deviceModel.get(deviceName).get('uuid')
-		pubNubKeys: configModel.getPubNubKeys()
-	.then (results) ->
-		return logs.history(results.pubNubKeys, results.uuid)
+exports.history = (uuid, callback) ->
+	deviceModel.get(uuid).then(configModel.getPubNubKeys).then (pubNubKeys) ->
```
Page-: There's no need for the `deviceModel.get(uuid)`?

lib/logs.coffee
jviotti: @Page- It is used in both cases to determine the existence of the uuid (since given an invalid uuid to `.get()` throws an error). I could use `.has()` which should express the intention clearer, but that would require me to check the return value of `.has()`, and it's false, throw the error myself. What do you think?

jviotti:
@Page- I've updated the code to use `device.has()` for clarity purposes.

Page-:
Ok, a comment could also have been ok, but it's nice to be explicit rather than relying on magic - I would never have clicked the behaviour without knowing the ins and outs of every single module (also if the behaviour of that module were to change it could be tricky)

Page-:
LGTM

jviotti:
@Page- Yeah true. I've tried adding a comment initially, but decided to go the clarity way.

PR #93, opened 2015-07-17
Upgrade Resin Request to v2.2.1

jviotti:
This change includes token refresh logic, which is critical to avoid users getting an invalid token after a while.


Page-:
LGTM

PR #107, opened 2015-08-03
Fix key.get() undefined issue. Fix #102.

### LINKED ISSUE #102: Returned result is undefined "resin.models.key.get()"
In both input cases (string|number) the result is undefined. The id number that i use comes from resin.models.key.getAll().

jviotti:


Page-:
LGTM - when you specify an id it returns that as a singular (because it knows that there can only be either 0 or 1 results - never more)

PR #113, opened 2015-08-04
Update key.create due to API change

jviotti:


lib/models/key.coffee
```diff
@@ -126,8 +128,14 @@ exports.create = (title, key, callback) ->
 	# Avoid ugly whitespaces
 	key = key.trim()
 
-	return pine.post
-		resource: 'user__has__public_key'
-		body: { title, key }
-	.get('id')
+	Promise.props
+		userId: auth.getUserId()
```
jviotti: Since we're only getting the userId, this should be done with a `.then()` instead of `.props`.

PR #125, opened 2015-08-21
Set RESIN and USER as system environment variables

jviotti:
Previously, only environment variables starting with `RESIN_` were
considered system environment variables, however, the UI also defines
`RESIN` and `USER` as reserved names.

This PR modified `resin.models.environmentVariables.isSystemVariable()`
to account for those cases.

Fixes: https://github.com/resin-io/resin-sdk/issues/124

### DIFF
--- a/lib/models/environment-variables.coffee
+++ b/lib/models/environment-variables.coffee
@@ -154,7 +154,7 @@ exports.remove = (id, callback) ->
 # > false
 ####
 exports.isSystemVariable = (variable) ->
-	return /^RESIN_/.test(variable.name)
+	return /^RESIN_|^RESIN$|^USER$/.test(variable.name)
 
 ###*
 # @namespace resin.models.environment-variables.device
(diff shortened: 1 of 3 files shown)


lib/models/environment-variables.coffee
```diff
@@ -154,7 +154,7 @@ exports.remove = (id, callback) ->
 # > false
 ####
 exports.isSystemVariable = (variable) ->
-	return /^RESIN_/.test(variable.name)
+	return /^RESIN(_.*)?|USER/.test(variable.name)
```
dfunckt: This regex will also catch strings like RESINATOR, SOME_USER.

The correct regex would be: /^RESIN_|^RESIN$|^USER$/

jviotti: Fixed and added unit tests, thanks for the catch!

dfunckt:
LGTM :)

emirotin:
LGTM

PR #141, opened 2015-09-11
Make generateUUID() async

jviotti:
The change is propagated from resin-io/resin-register-device. See
commit:

	https://github.com/resin-io/resin-register-device/commit/e1858f00d8659f5ae8fcb07f6dcd245a217e1bcc


pcarranzav:
LGTM

PR #145, opened 2015-10-02
Make getPubNubKeys and getMixpanelToken private

jviotti:
Currently this functions are publicly exposed in the documentation.
Given that there is no need for a user to interact with them directly,
we mark them as *private*.

Notice that the functions still exist and are available to the client,
but they don't appear in the public generated documentation.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -3,6 +3,12 @@
 All notable changes to this project will be documented in this file.
 This project adheres to [Semantic Versioning](http://semver.org/).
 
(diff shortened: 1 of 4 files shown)


abresas:
LGTM

abresas:
changelog?

jviotti:
@abresas I write the changelog when releasing by checking the merged commits, but may be a good opportunity to improve that practice!

jviotti:
@abresas Done!

PR #152, opened 2015-10-22
Simplify resin.auth.2fa.isPassed()

jviotti:
We return `true` if 2FA is not enabled, which means that
`twoFactorRequired` is `undefined`.

Later on, we attempt to negate the value of `twoFactorRequired`.

Notice that `!undefined === true`, therefore if 2FA is not enabled, the
second check will return `true` already.


nghiant2710:
LGTM :+1:

PR #178, opened 2016-01-21
Fix error when fetching a device with a number uuid

jviotti:
Take a device uuid like this:

	728046334d1fcbb956ae240381d68ba3283ce8b1a870f512665271134f7a19

Notice the first nine characters are numbers, therefore we expect
something like:

	resin.models.device.get(728046334)

But passing a number directly throws the following error:

	Expected null/string/number/obj/array, got: undefined

on `uuid.length`.


Page-:
LGTM

PR #188, opened 2016-02-09
Update env vars restricted error message

jviotti:


jviotti:
Turns out the error message change is only in staging, so we will merge this once staging gets deployed.

PR #193, opened 2016-02-18
Decouple SDK from environment settings

jviotti:
This is a big, breaking change. Currently the SDK reads settings from the filesystem and environment variables using `resin-settings-client`. With this PR, the SDK only maintain runtime settings set by `resin.settings.set()`.


jviotti:
@emirotin Can we close this PR?

emirotin:
Yeah, I'll send my own version of it once we have all the deps merged.

PR #196, opened 2016-02-27
Implement resin.models.device.reboot()

jviotti:
This function allows the user to perform a hard-reboot on the device by
using the Supervisor API:

https://github.com/resin-io/resin-supervisor/blob/master/docs/API.md#post-v1reboot

This function, as well as `resin.models.device.restart()` is not tested in
the integration suite since its not a behaviour we can easily assert on.

Fixes: https://github.com/resin-io/resin-sdk/issues/195

Misc changes:

- Clarify that `device.restart()` restarts the application

### DIFF
--- a/DOCUMENTATION.md
+++ b/DOCUMENTATION.md
@@ -36,6 +36,7 @@ If you feel something is missing, not clear or could be improved, please don't h
             * [.note(uuid, note)](#resin.models.device.note) ⇒ <code>Promise</code>
             * [.move(uuid, application)](#resin.models.device.move) ⇒ <code>Promise</code>
             * [.restart(uuid)](#resin.models.device.restart) ⇒ <code>Promise</code>
+            * [.reboot(uuid)](#resin.models.device.reboot) ⇒ <code>Promise</code>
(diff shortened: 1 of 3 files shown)


DOCUMENTATION.md
```diff
@@ -729,8 +732,12 @@ resin.models.device.move('7cf02a6', 'MyApp', function(error) {
 ```
 <a name="resin.models.device.restart"></a>
 ##### device.restart(uuid) ⇒ <code>Promise</code>
+This function restarts the Docker container running
+the application on the device, but doesn't reboot
+the device itself.
+
 **Kind**: static method of <code>[device](#resin.models.device)</code>  
-**Summary**: Restart device  
+**Summary**: Restart device application
```
pcarranzav: Might be better to rephrase this to `Restart application on device`

pcarranzav:
LGTM (besides that one comment)

DOCUMENTATION.md
jviotti: @pcarranzav Good suggestion, thanks!

PR #198, opened 2016-03-09
Make use of new Bluebird catch utilities

jviotti:


lib/models/os.coffee
```diff
@@ -49,9 +49,9 @@ exports.getLastModified = (deviceType, callback) ->
 		method: 'HEAD'
 		url: getImageMakerUrl(deviceType)
 	.catch (error) ->
```
Page-: This can be `.catch { name: 'ResinRequestError', statusCode: 404 }, ...`

jviotti: Oh nice, I have another module that has a similar pattern as well. Thanks!

PR #206, opened 2016-04-20
Implement application container related supervisor API endpoints

lekkas:
As with #205 , there are no integration tests since we need a real device for testing.

The supervisor API endpoints in question are

* `/v1/apps/<appId>`
* `/v1/apps/<appId>/stop`
* `/v1/apps/<appId>/start`

and can be found here: https://github.com/resin-io/resin-supervisor/pull/126/

---
This change is [Reviewable](https://reviewable.io/reviews/resin-io/resin-sdk/206)

### DIFF
--- a/DOCUMENTATION.md
+++ b/DOCUMENTATION.md
@@ -37,6 +37,7 @@ If you feel something is missing, not clear or could be improved, please don't h
             * [.rename(uuid, newName)](#resin.models.device.rename) ⇒ <code>Promise</code>
             * [.note(uuid, note)](#resin.models.device.note) ⇒ <code>Promise</code>
             * [.move(uuid, application)](#resin.models.device.move) ⇒ <code>Promise</code>
+            * [.startApplication(uuid)](#resin.models.device.startApplication) ⇒ <code>Promise</code>
             * [.stopApplication(uuid)](#resin.models.device.stopApplication) ⇒ <code>Promise</code>
             * [.restart(uuid)](#resin.models.device.restart) ⇒ <code>Promise</code>
             * [.reboot(uuid)](#resin.models.device.reboot) ⇒ <code>Promise</code>
@@ -137,6 +138,7 @@ If you feel something is missing, not clear or could be improved, please don't h
         * [.rename(uuid, newName)](#resin.models.device.rename) ⇒ <code>Promise</code>
         * [.note(uuid, note)](#resin.models.device.note) ⇒ <code>Promise</code>
         * [.move(uuid, application)](#resin.models.device.move) ⇒ <code>Promise</code>
+        * [.startApplication(uuid)](#resin.models.device.startApplication) ⇒ <code>Promise</code>
         * [.stopApplication(uuid)](#resin.models.device.stopApplication) ⇒ <code>Promise</code>
         * [.restart(uuid)](#resin.models.device.restart) ⇒ <code>Promise</code>
         * [.reboot(uuid)](#resin.models.device.reboot) ⇒ <code>Promise</code>
@@ -429,6 +431,7 @@ resin.models.application.getApiKey('MyApp', function(error, apiKey) {
     * [.rename(uuid, newName)](#resin.models.device.rename) ⇒ <code>Promise</code>
     * [.note(uuid, note)](#resin.models.device.note) ⇒ <code>Promise</code>
     * [.move(uuid, application)](#resin.models.device.move) ⇒ <code>Promise</code>
+    * [.startApplication(uuid)](#resin.models.device.startApplication) ⇒ <code>Promise</code>
     * [.stopApplication(uuid)](#resin.models.device.stopApplication) ⇒ <code>Promise</code>
     * [.restart(uuid)](#resin.models.device.restart) ⇒ <code>Promise</code>
     * [.reboot(uuid)](#resin.models.device.reboot) ⇒ <code>Promise</code>
@@ -808,6 +811,31 @@ resin.models.device.move('7cf02a6', 'MyApp');
 ```js
 resin.models.device.move('7cf02a6', 'MyApp', function(error) {
 	if (error) throw error;
+});
+```
+<a name="resin.models.device.startApplication"></a>
+
+##### device.startApplication(uuid) ⇒ <code>Promise</code>
+**Kind**: static method of <code>[device](#resin.models.device)</code>  
+**Summary**: Start application on device  
+**Access:** public  
+**Fulfil**: <code>String</code> - application container id  
+
+| Param | Type | Description |
+| --- | --- | --- |
+| uuid | <code>String</code> | device uuid |
+
+**Example**  
+```js
+resin.models.device.startApplication('7cf02a6').then(function(containerId) {
+	console.log(containerId);
+});
+```
+**Example**  
+```js
+resin.models.device.startApplication('7cf02a6', function(error, containerId) {
+	if (error) throw error;
+	console.log(containerId);
 });
 ```
 <a name="resin.models.device.stopApplication"></a>
--- a/DOCUMENTATION.md
+++ b/DOCUMENTATION.md
@@ -37,6 +37,7 @@ If you feel something is missing, not clear or could be improved, please don't h
             * [.rename(uuid, newName)](#resin.models.device.rename) ⇒ <code>Promise</code>
             * [.note(uuid, note)](#resin.models.device.note) ⇒ <code>Promise</code>
             * [.move(uuid, application)](#resin.models.device.move) ⇒ <code>Promise</code>
+            * [.stopApplication(uuid)](#resin.models.device.stopApplication) ⇒ <code>Promise</code>
             * [.restart(uuid)](#resin.models.device.restart) ⇒ <code>Promise</code>
             * [.reboot(uuid)](#resin.models.device.reboot) ⇒ <code>Promise</code>
             * [.shutdown(uuid)](#resin.models.device.shutdown) ⇒ <code>Promise</code>
@@ -136,6 +137,7 @@ If you feel something is missing, not clear or could be improved, please don't h
         * [.rename(uuid, newName)](#resin.models.device.rename) ⇒ <code>Promise</code>
         * [.note(uuid, note)](#resin.models.device.note) ⇒ <code>Promise</code>
         * [.move(uuid, application)](#resin.models.device.move) ⇒ <code>Promise</code>
+        * [.stopApplication(uuid)](#resin.models.device.stopApplication) ⇒ <code>Promise</code>
         * [.restart(uuid)](#resin.models.device.restart) ⇒ <code>Promise</code>
         * [.reboot(uuid)](#resin.models.device.reboot) ⇒ <code>Promise</code>
         * [.shutdown(uuid)](#resin.models.device.shutdown) ⇒ <code>Promise</code>
@@ -427,6 +429,7 @@ resin.models.application.getApiKey('MyApp', function(error, apiKey) {
     * [.rename(uuid, newName)](#resin.models.device.rename) ⇒ <code>Promise</code>
     * [.note(uuid, note)](#resin.models.device.note) ⇒ <code>Promise</code>
     * [.move(uuid, application)](#resin.models.device.move) ⇒ <code>Promise</code>
+    * [.stopApplication(uuid)](#resin.models.device.stopApplication) ⇒ <code>Promise</code>
     * [.restart(uuid)](#resin.models.device.restart) ⇒ <code>Promise</code>
     * [.reboot(uuid)](#resin.models.device.reboot) ⇒ <code>Promise</code>
     * [.shutdown(uuid)](#resin.models.device.shutdown) ⇒ <code>Promise</code>
@@ -805,6 +808,31 @@ resin.models.device.move('7cf02a6', 'MyApp');
 ```js
 resin.models.device.move('7cf02a6', 'MyApp', function(error) {
 	if (error) throw error;
+});
+```
+<a name="resin.models.device.stopApplication"></a>
+
+##### device.stopApplication(uuid) ⇒ <code>Promise</code>
+**Kind**: static method of <code>[device](#resin.models.device)</code>  
+**Summary**: Stop application on device  
+**Access:** public  
+**Fulfil**: <code>String</code> - application container id  
+
+| Param | Type | Description |
+| --- | --- | --- |
+| uuid | <code>String</code> | device uuid |
+
+**Example**  
+```js
+resin.models.device.stopApplication('7cf02a6').then(function(containerId) {
+	console.log(containerId);
+});
+```
+**Example**  
+```js
+resin.models.device.stopApplication('7cf02a6', function(error, containerId) {
+	if (error) throw error;
+	console.log(containerId);
 });
 ```
 <a name="resin.models.device.restart"></a>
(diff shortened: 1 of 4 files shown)


lib/models/device.coffee
```diff
@@ -249,6 +249,41 @@ exports.getApplicationName = (uuid, callback) ->
 	exports.get(uuid).get('application_name').nodeify(callback)
 
 ###*
+# @summary Get application container id
```
jviotti: I think this summary is not accurate. The `containerId` is one of the properties that can be retrieved from this endpoint.

The summary should be more generic and in line with the function name (e.g: "Get application container information"), or the function should be edited to only return the container id, and be renamed to something like `getContainedId()`.

lib/models/device.coffee
```diff
@@ -509,6 +544,76 @@ exports.move = (uuid, application, callback) ->
...
+# 	console.log(containerId);
+# });
+#
+# @example
+# resin.models.device.start('7cf02a6', function(error, containerId) {
+# 	if (error) throw error;
+# 	console.log(containerId);
+# });
+###
+exports.start = (uuid, callback) ->
```
jviotti: I think the function name is a bit too vague. "start" can mean "start the device" (e.g: power it on), "start the application", "start the supervisor", etc. What about `.startContainer()`?

lib/models/device.coffee
```diff
@@ -509,6 +544,76 @@ exports.move = (uuid, application, callback) ->
...
 
 ###*
+# @summary Start application on device
+# @name start
+# @public
+# @function
+# @memberof resin.models.device
+#
+# @param {String} uuid - device uuid
+# @fulfil {String} - application container id
```
jviotti: The function resolves the response body, which is an object containing a `containerId` property, but not the property itself. You can add `.get('containerId')` after `.get('body')` so the functionality works according to this claim.

lib/models/device.coffee
```diff
@@ -509,6 +544,76 @@ exports.move = (uuid, application, callback) ->
...
+
+###*
+# @summary Stop application on device
+# @name stop
+# @public
+# @function
+# @memberof resin.models.device
+#
+# @param {String} uuid - device uuid
+# @fulfil {String} - application container id
```
jviotti: Same as above. This function resolves and object containing a `containerId` property, but not the property itself.

lib/models/device.coffee
```diff
@@ -509,6 +544,76 @@ exports.move = (uuid, application, callback) ->
...
+# 	console.log(containerId);
+# });
+#
+# @example
+# resin.models.device.stop('7cf02a6', function(error, containerId) {
+# 	if (error) throw error;
+# 	console.log(containerId);
+# });
+###
+exports.stop = (uuid, callback) ->
```
jviotti: What about `.stopContainer()`, for the same reasons outlined above?

lib/models/device.coffee
```diff
@@ -509,6 +544,76 @@ exports.move = (uuid, application, callback) ->
...
+	exports.get(uuid).then (device) ->
+		appId = device.application[0].id
+		return request.send
+			method: 'GET'
```
jviotti: This should be a `POST`, according to the supervisor docs.

lib/models/device.coffee
```diff
@@ -509,6 +544,76 @@ exports.move = (uuid, application, callback) ->
...
+# resin.models.device.start('7cf02a6', function(error, containerId) {
+# 	if (error) throw error;
+# 	console.log(containerId);
+# });
+###
+exports.start = (uuid, callback) ->
+	exports.get(uuid).then (device) ->
+		appId = device.application[0].id
+		return request.send
+			method: 'GET'
```
jviotti: This should be a `POST`, according to the supervisor docs.

lib/models/device.coffee
lekkas: Good catch Juan, I originally had a `getContainerId()` endpoint and missed this part when refactoring.

lib/models/device.coffee
lekkas: Right, so there is already a `.restart()` method, which has the semantics of restarting the application, so I kept `.start()` and `.stop()` for consistency. Also, at this point we assume single-container apps, since the supervisor APIs do not support taking a containerId as an argument. I am fine with renaming the function if you feel strongly about it though. 

We could also have `.restartApplication()`, `.stopApplication()`, `.startApplication()`

lib/models/device.coffee
lekkas: Good catch 👍

lib/models/device.coffee
lekkas: Fixed by adding `.get('containerId')`

lib/models/device.coffee
lekkas: Fixed!

lib/models/device.coffee
lekkas: Let's continue the naming discussion in the above thread

lib/models/device.coffee
jviotti: I see. I like having `Application` as prefix for all functions, including `.restart()`, however that'd involve a breaking change for little benefit. 

I say let's keep them like they are (`.start()` and `.stop()`), but can you make a GitHub issue about renaming all of them, including `.restart()` to include the `Application` suffix?

lekkas: Issue created here: https://github.com/resin-io/resin-sdk/issues/207

DOCUMENTATION.md
```diff
@@ -590,6 +599,31 @@ resin.models.device.getApplicationName('7cf02a6', function(error, applicationNam
 	console.log(applicationName);
 });
 ```
+<a name="resin.models.device.getApplicationInfo"></a>
+
+##### device.getApplicationInfo(uuid) ⇒ <code>Promise</code>
+**Kind**: static method of <code>[device](#resin.models.device)</code>
+**Summary**: Get application info
```
jviotti: I believe you forgot to re-create the docs.

lib/models/device.coffee
```diff
@@ -249,6 +249,41 @@ exports.getApplicationName = (uuid, callback) ->
...
+# });
+###
+exports.getApplicationInfo = (uuid, callback) ->
+	exports.get(uuid).then (device) ->
+		appId = device.application[0].id
+		return request.send
+			method: 'GET'
+			url: "/supervisor/v1/apps/#{appId}"
+			baseUrl: settings.get('apiUrl')
+			body:
```
Page-: The body can be just `{ uuid }` instead

lib/models/device.coffee
```diff
@@ -249,6 +249,41 @@ exports.getApplicationName = (uuid, callback) ->
...
+# });
+#
+# @example
+# resin.models.device.getApplicationInfo('7cf02a6', function(error, appInfo) {
+# 	if (error) throw error;
+# 	console.log(appInfo);
+# });
+###
+exports.getApplicationInfo = (uuid, callback) ->
+	exports.get(uuid).then (device) ->
```
Page-: Since we're fetching all the device info it'd be worth checking the supervisor version here imo, to see if it's up to date enough to have the endpoints (and give a nice message if not)

lib/models/device.coffee
Page-: Can we do:
```
exports.restart = exports.application.restart
```
or the equivalent, in order to have the new actions have the api we want moving forwards, but keep backwards compatibility until a major version where we bundle up more breaking changes?

lib/models/device.coffee
lekkas: Done. Added supervisor version checks to all `/v1/apps/<appId>` endpoints.

lib/models/device.coffee
lekkas: Done, added `uuid: uuid` instead to be consistent with the `/blink` endpoint request.

lib/models/device.coffee
lekkas: A small note on this, the supervisor version is checked on each `/v1/apps` endpoint separately, maybe a better way would be to have minimum supervisor versions for every endpoint - I could open an gh issue for this.

lib/models/device.coffee
lekkas: Can you explain this a bit more? I'm confused because right now both the `device` and the `application` model have a restart method, the former restart the application, while the latter restarts the app container on the device.

lib/models/device.coffee
jviotti: @Page- @lekkas I believe the logic to check the minimum supervisor version and show a warning message should live in the API (the supervisor proxy) instead of here. We're hardcoding a specific minimum supervisor version in the SDK (which its not ideal) and it still doesn't protect users using the API directly, or the Python SDK for example.

If this logic is in the supervisor proxy, the endpoint can return a custom HTTP header to determine a warning notice (e.g: `X-Notice`, or something like that), that all SDKs can use to generically display this.

lib/models/device.coffee
jviotti: @Page- Sounds like a good plan.

@lekkas What Page means is that you can rename `.start()` to `.startApplication()`, `.stop()` to `.stopApplication()` and edit the definition of `.restart()` to be `exports.restart = exports.restartApplication = ( ... ) ->`, so we maintain backwards compatibility.

My only worry is how this will play with the generated documentation. If you have a way to display both `.restar()` and `.restartApplication()` in `DOCUMENTATION.md` in a way that they share the function details, but the former is marked a "deprecated", we should do it.

lekkas: Ok, added the `Application` suffix to all `/v1/apps` endpoints, also added a deprecation notice for `.restart()`.

lib/models/device.coffee
lekkas: I've already added a supervisor check, let's wait for @Page- and I can happily revert the commit if we eventually decide not to have a check on the SDK. 

Imo it makes sense to me to have the API supervisor proxy as simple as possible, which I would expect to transparently forward requests to the device.

The supervisor API is, after all, documented and the SDK has the purpose of facilitating its use, so I wouldn't mind adding supervisor version checks here (and to other SDKs, like the python one). Users who choose to directly use the API (e.g. with curl) aren't protected from making a wrong API call anyway. - By the way, how does the Supervisor API proxy handle non-existent endpoints?

lib/models/device.coffee
lekkas: @jviotti please check the deprecation comments in `DOCUMENTATION.md` and `devices.coffee` and let me know if you think its an abomination :)

lib/models/device.coffee
```diff
@@ -541,6 +659,21 @@ exports.restart = (uuid, callback) ->
 	.nodeify(callback)
 
 ###*
+# @summary Restart application on device. *Deprecated*: Use
```
jviotti: Can you remove the deprecation notice and link from the summary? JSDoc is already marking it as deprecate because of the `@deprecated` tag, and instead of linking to the other function directly, you can make use of `@see`.

lib/models/device.coffee
jviotti: Well, the proxy has the purpose of being the intermediary between the client and the supervisor running on the device. 

The benefit of having it there is that this logic, albeit minimum, will not be duplicated in various projects (e.g: the Python SDK), and gives the proxy the freedom to add/remove/edit warning/deprecation notices at will, without needing to force clients to upgrade their SDKs, or having to commit and publish new versions for little benefit.

We can either hardcode a warning in a single SDK, or, with a bit more work, have a robust and scalable way to report warnings/deprecations to all clients, which would be valuable in the future.

lib/models/device.coffee
lekkas: Done. Also `@deprecated` should support a notice text after the tag (http://usejsdoc.org/tags-deprecated.html), yet it didn't seem to appear in the generated `DOCUMENTATION.md`

lib/models/device.coffee
Page-: It's a tricky one, because the idea was to work nicely with custom supervisor builds (which might report a non-semver version or whatever they want) and also have custom endpoints or different behaviour, so forwarding dumbly works the best with that.  The SDK however is linked to the official resin supervisor, as it expects the endpoints that exist to match our supervisor and to behave in the same way, so we already have a dependency on our supervisor in the SDK and so can use that existing requirement to add additional linking.

lekkas: The way I see it, the supervisor API proxy is a transparent intermediate and therefore it's the SDK's responsibility to be updated with the supervisor API.

I also don't think there is a dichotomy here, we can have the SDKs track supervisor API versions *and* have a robust/scalable way to report deprecations.

jviotti: Okay, fair enough. Let's create a private helper function to display this warning so we don't repeat ourselves, and let's make sure this warning gets added to the Python SDK as well for consistency, at least with a reported GH issue.

lekkas: Added a private helper function in resin.models.device. Though not asynchronous, the function used Promises for consistency.

lekkas: And the Python SDK ticket - https://github.com/resin-io/resin-sdk-python/issues/30

lib/models/device.coffee
```diff
@@ -27,6 +28,38 @@ configModel = require('./config')
...
+# 	console.log(isCompatible);
+# });
+#
+# @example
+# resin.models.device.isCompatible(version, MIN_VERSION, function(error, isCompatible) {
+# 	if (error) throw error;
+# 	console.log(isCompatible);
+# });
+###
+isCompatible = (version, minVersion) ->
```
Page-: This would be nicer (imo) as:
```coffee
isCompatible = Promise.method (version, minVersion) ->
	if semver.lt(version, minVersion)
		throw new Error("Incompatible supervisor version: #{version} - must be >= #{minVersion}")
	return true
```

lib/models/device.coffee
```diff
@@ -27,6 +28,38 @@ configModel = require('./config')
...
+# @memberof resin.models.device
+#
+# @param {String} version - version under check
+# @param {String} minVersion - minimum accepted version
+# @fulfil {Boolean} - is compatible
+# @reject {Error} Will reject if the given version is < than the given minimum version
+# @returns {Promise}
+#
+# @example
+# resin.models.device.isCompatible(version, MIN_VERSION).then(function(isCompatible) {
```
Page-: This example makes no sense - you can't call it externally at all

lekkas: exported the `isCompatible()` method in the new PR..

lib/models/device.coffee
lekkas: Nice! New promise tool :)

Page-:
LGTM

lib/models/device.coffee
```diff
@@ -27,6 +28,37 @@ configModel = require('./config')
...
+# 	console.log(isCompatible);
+# });
+#
+# @example
+# resin.models.device.isCompatible(version, MIN_VERSION, function(error, isCompatible) {
+# 	if (error) throw error;
+# 	console.log(isCompatible);
+# });
+###
+exports.isCompatible = Promise.method (version, minVersion) ->
```
jviotti: I think we should not export this. In the case of the example, you can just say `isCompatible()`. Also, the fact that the function signature contains `is` implies that the function will return either `true` or `false`, without any side effect, like an error.

What about `ensureSupervisorCompability()`? Clarifying "Supervisor" also make the type of compatibility that we want to check more obvious.

lekkas: The `export` was made with the assumption that this method could potentially be reused by other modules - like, for instance, `getPubNubKeys` from the `config` model, which is also private.

I agree that `is` implies a `true` or `false` return value, yet I tried to use a short function name and couldn't really think of something better - the function's documentation also clearly states that the method refers to the supervisor version. 

I'm pushing a change with the proposed `ensureSupervisorCompability` name to be consistent with the SDK coding conventions.

jviotti: I see, what about creating a new file called "compatibility.coffee" or something like that that includes and exports this function?

lekkas: Isn't supervisor compatibility device-specific? Why have it outside the `device` model?

lekkas:
@jviotti , can I have a concluding lgtm or comments please?

jviotti:
@lekkas LGTM

PR #210, opened 2016-05-16
Change supervisor semver check to >= 1.8.0-alpha.0 for /apps endpoint

lekkas:
One alternative is to check if version is greater than all the versions possible in the `1.7` range with `semver.gtr(detected_version, 1.7)`, changing the semantics of `ensureSupervisorCompatibility()` accordingly and having a `SUPERVISOR_APPS_API_MAX_INCOMPATIBLE_RANGE` constant (or similar) to hold `1.7`. 

A benefit of the `gtr()` alternative is that users will get a friendlier message ("*version should be > 1.7*" instead of this PR's "*version should be >= 1.8.0-alpha.0*"), the drawback is that it complicates the semantics of `ensureSupervisorCompatibility()`- @shaunmulligan @jviotti @Page- ideas?

### DIFF
--- a/build/models/device.js
+++ b/build/models/device.js
@@ -44,7 +44,7 @@ limitations under the License.
 
   auth = require('../auth');
 
-  MIN_SUPERVISOR_APPS_API = '1.8.0';
+  MIN_SUPERVISOR_APPS_API = '1.8.0-alpha.0';
 
 
   /**
--- a/lib/models/device.coffee
+++ b/lib/models/device.coffee
@@ -28,7 +28,11 @@ configModel = require('./config')
 applicationModel = require('./application')
 auth = require('../auth')
 
-MIN_SUPERVISOR_APPS_API = '1.8.0'
+# The min version where /apps API endpoints are implemented is 1.8.0 but we'll
+# be accepting >= 1.8.0-alpha.0 instead. This is a workaround for a published 1.8.0-p1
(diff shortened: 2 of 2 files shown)


Page-:
SGTM

jviotti:
I think it looks good. Let's merge once Appveyor tests pass.

lekkas:
@jviotti Is it possible to re-run the appveyor tests somehow? The test failures do not seem related to this PR.

jviotti:
@lekkas Hm, we recently upgraded our Appveyor plan to include parallel builds, which is causing conflicts between SDK builds since they re-use the same testing account.

We can either create different accounts for each Node version (although this still has the possibility of failure if another PR is built in parallel), or somehow tell Appveyor to not run parallel builds for this project.

Let me check if we can configure the latter.

jviotti:
@lekkas I've fixed the issue. Once you rebase `master` on this branch, the issue should be gone.

lekkas:
@jviotti A single test fails in ` continuous-integration/appveyor/pr `:

```
1) SDK Integration Tests given a logged in fresh user given a single application with two offline devices that share the same uuid root "before each" hook for "should be rejected with an error if there is an ambiguation between shorter uuids":
```

jviotti:
I'll publish to NPM in a minute.

jviotti:
@lekkas v5.3.1 is out with the fix.

lekkas:
Great 👍 ! So what was the AppVayor failed test about?

PR #218, opened 2016-08-12
return full uuid for getDeviceUrl

craig-mulligan:
connects to https://github.com/resin-io/resin-cli/issues/388 (does waffle work cross repos?)

I didn't have time to run the tests, cause I have to run, but pushed in case you have a chance to look at it today @jviotti

### DIFF
--- a/lib/models/device.coffee
+++ b/lib/models/device.coffee
@@ -1121,9 +1121,9 @@ exports.getDeviceUrl = (uuid, callback) ->
 		if not hasDeviceUrl
 			throw new Error("Device is not web accessible: #{uuid}")
 
-		return configModel.getAll().get('deviceUrlsBase')
-	.then (deviceUrlsBase) ->
-		return "https://#{uuid}.#{deviceUrlsBase}"
+		configModel.getAll().get('deviceUrlsBase').then (deviceUrlsBase) ->
+			exports.get(uuid).get('uuid').then (uuid) ->
+				return "https://#{uuid}.#{deviceUrlsBase}"
 	.nodeify(callback)
 
 ###*
--- a/tests/integration.coffee
+++ b/tests/integration.coffee
@@ -946,6 +946,10 @@ describe 'SDK Integration Tests', ->
 
 					describe 'resin.models.device.getDeviceUrl()', ->
 
+						it 'should eventually return the correct device url given a shorter uuid', ->
+							promise = resin.models.device.getDeviceUrl(@device.uuid.slice(0, 7))
+							m.chai.expect(promise).to.eventually.match(/[a-z0-9]{62}/)
+
 						it 'should eventually be an absolute url', (done) ->
 							resin.models.device.getDeviceUrl(@device.uuid).then (deviceUrl) ->
 								return request.getAsync(deviceUrl)
--- a/build/models/device.js
+++ b/build/models/device.js
@@ -1244,9 +1244,11 @@ limitations under the License.
       if (!hasDeviceUrl) {
         throw new Error("Device is not web accessible: " + uuid);
       }
-      return configModel.getAll().get('deviceUrlsBase');
-    }).then(function(deviceUrlsBase) {
-      return "https://" + uuid + "." + deviceUrlsBase;
+      return configModel.getAll().get('deviceUrlsBase').then(function(deviceUrlsBase) {
+        return exports.get(uuid).get('uuid').then(function(uuid) {
+          return "https://" + uuid + "." + deviceUrlsBase;
+        });
+      });
     }).nodeify(callback);
   };
 


lib/models/device.coffee
```diff
@@ -1117,13 +1117,16 @@ exports.hasDeviceUrl = (uuid, callback) ->
...
 exports.getDeviceUrl = (uuid, callback) ->
-	exports.hasDeviceUrl(uuid).then (hasDeviceUrl) ->
+	exports.hasDeviceUrl(uuid).bind({}).then (hasDeviceUrl) ->
 		if not hasDeviceUrl
 			throw new Error("Device is not web accessible: #{uuid}")
 
 		return configModel.getAll().get('deviceUrlsBase')
 	.then (deviceUrlsBase) ->
-		return "https://#{uuid}.#{deviceUrlsBase}"
+		this.deviceUrlsBase = deviceUrlsBase
```
Page-: Can you do
```coffee
.then (deviceUrlsBase) ->
	exports.get(uuid).get('uuid')
	.then (uuid) ->
		return "https://#{uuid}.#{deviceUrlsBase}"
```
instead, to avoid the magic with binding and `this` (which is actually a pretty nifty hack btw, but the simpler solution is preferable for readability :P )

jviotti:
Looks good, I'll merge once the following items are completed:

- Re-build the JavaScript to include latest changes.
- Some tests are added. There are a few that specifically test shorter uuid vs longer uuid stuff, so you can adapt one of those.

Let me know if I can help with anything!

lib/models/device.coffee
craig-mulligan: Awesome thanks! I've just read so many things about nested promises that I try an avoid them completely, but one level down is fine :) Cheers!

tests/integration.coffee
```diff
@@ -946,8 +946,12 @@ describe 'SDK Integration Tests', ->
 
 					describe 'resin.models.device.getDeviceUrl()', ->
 
+						it 'should eventually be a url containing full length uuid', ->
```
craig-mulligan: Unsure if this should just be included in the following `'should eventually be an absolute url'` test or it's own? @jviotti

jviotti: Better to have a separate test IMO.

tests/integration.coffee
```diff
@@ -946,8 +946,12 @@ describe 'SDK Integration Tests', ->
...
+							promise = resin.models.device.getDeviceUrl(@device.uuid.slice(0, 7))
+							m.chai.expect(promise).to.eventually.match(/[a-z0-9]{62}/)
+
 						it 'should eventually be an absolute url', (done) ->
-							resin.models.device.getDeviceUrl(@device.uuid).then (deviceUrl) ->
+							resin.models.device.getDeviceUrl(@device.uuid.slice(0, 7)).then (deviceUrl) ->
```
jviotti: Why is this being changed to use a shorter uuid if the above test is already handling the shorter uuid?

tests/integration.coffee
jviotti: Maybe the title can be more like "should eventually return the correct device url given a shorter uuid", or something like that that highlights the important point of the test is the shorter uuid behaviour?

tests/integration.coffee
craig-mulligan: To be double sure... no you are right I'll update.

lib/models/device.coffee
```diff
@@ -1121,10 +1121,10 @@ exports.getDeviceUrl = (uuid, callback) ->
...
 			throw new Error("Device is not web accessible: #{uuid}")
 
-		return configModel.getAll().get('deviceUrlsBase')
-	.then (deviceUrlsBase) ->
-		return "https://#{uuid}.#{deviceUrlsBase}"
-	.nodeify(callback)
+		configModel.getAll().get('deviceUrlsBase').then (deviceUrlsBase) ->
+			exports.get(uuid).get('uuid').then (uuid) ->
+				return "https://#{uuid}.#{deviceUrlsBase}"
+			.nodeify(callback)
```
jviotti: This `.nodeify()` call should be more indented to the left, to match line 1120, otherwise, this means we're only "nodeifying" the most inner promise.

craig-mulligan:
Good to merge @jviotti?

jviotti:
Yeah, looks great, will merge in a second. Do you need me to publish this ASAP? Thanks a lot!

craig-mulligan:
Nope, I was just running through my open PRs and wanted get them merged. No rush at all :)

jviotti:
Ok, amazing! I'll try to batch it together with other SDK changes, but if you ever need this, just ping me and I'll publish.

PR #219, opened 2016-08-31
Add a warning to the readme about the test account getting reset.

Page-:
I'm not 100% if this is the case, but I believe so - I'm glad I noticed this before I tried running the tests on my user!


jviotti:
Right, this should have been documented! You luckily double checked for your account! Everything of course looks fine. I do wonder if a CHANGELOG entry is worth. Maybe we should omit it?

Page-:
I don't mind, it's your call, I just added it from habit

PR #220, opened 2016-08-31
WIP: Update to use the latest device registration method.

Page-:
Depends on https://github.com/resin-io-modules/resin-register-device/pull/22

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -3,6 +3,11 @@
 All notable changes to this project will be documented in this file.
 This project adheres to [Semantic Versioning](http://semver.org/).
 
+### Breaking Changes
+
+- `resin.models.device.generateUUID()` is now known as `resin.models.device.generateUniqueKey()` and is synchronous.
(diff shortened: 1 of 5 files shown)


jviotti:
As far as I understand, all tests are failing due to the PR you're referring to above is not yet merged, right?

Page-:
Yeah, the dependency needs to be merged/published before the tests here can succeed

Page-:
@pimterry can you check that this is already implemented in the SDK - it looks to me like it is (resin-register-device is updated and the function renamed) and if so then this can be closed

pimterry:
@Page- very similar changes have been made, but they're not exactly the same as this PR. We do now provide the correct `generateUniqueKey` and we do register with a provisioning key, but we don't pass a deviceApiKey, unlike this PR. Judging from the API code that's fine, as the API will generate one and send it back, which should be sufficient. Is that right?

Generally I don't feel like I totally understand the flow here - do we have any docs on the full device provisioning & api key process, and how we think it _should_ work? Why would users ever want to specify the device's api key for themselves?

PR #227, opened 2016-09-27
resin.models.device.ensureSupervisorCompatibility(): add tests and handle invalid semver versions gracefully

### LINKED ISSUE #222: Graciously handle invalid semver versions in resin.models.device.getApplicationInfo
If for any reason HostOS or Supervisor versions are incorrectly set, the `semver` module will fail with a cryptic 'Invalid Version: null' message.

lekkas:
Closes #222

### DIFF
--- a/build/models/device.js
+++ b/build/models/device.js
@@ -75,6 +75,9 @@ limitations under the License.
    */
 
   exports.ensureSupervisorCompatibility = ensureSupervisorCompatibility = Promise.method(function(version, minVersion) {
+    if (!semver.valid(version)) {
+      throw new Error("Invalid supervisor version: " + version);
+    }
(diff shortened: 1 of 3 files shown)


lekkas:
Tests fail with
```
 AssertionError: expected promise to be rejected with an error including 'No such device type' but got 'Request error: Device type "foo-bar-baz" not found'
```

TIL that `resin-sdk` actually uses the real thing during testing.

Page-:
Yeah, I PRed https://github.com/resin-io/resin-sdk/pull/219 when I found out that :P

lekkas:
If it wasn't for the warning I'd definitely have used my own account :P

pcarranzav:
@Page- @lekkas I PRd #230 to make the tests work again

PR #234, opened 2016-11-24
update nodejs test versions

emirotin:


emirotin:
Same thing

emirotin:
AHA! **Similarly, we do not provide these values to untrusted builds, triggered by pull requests from another repository.**

jviotti:
Oh, I see. That makes sense, since you're PRing from a fork. I'll see if I can enable that somehow anyway, since the same issue might be hit by external contributors at some point.

PR #239, opened 2016-12-14
Make the tests stable

### LINKED ISSUE #237: Make the tests stable
See https://github.com/resin-io/resin-sdk/pull/236#issuecomment-266483336.

Currently the tests frequently (maybe 2/3rds of the time) fail due to ETIMEDOUT on Windows when making requests (of no consistent type) to the Resin API. The tests should reliably pass.

pimterry:
This fixes #237, building on top of #236.

This change is very simple, just fixes a Karma config bug (arguably a Karma bug, see karma-runner/karma#2505), brings in the new resin-request and puts resin-io-modules/resin-request#78 and resin-io-modules/resin-request#79 to use.

End result is that we now output some debug info during the tests about the requests being made, and retry failing requests. The debug output is just so we can debug this better if it does start to go wrong later - it's a bit noisy for now, but we can remove it soon once we're sure this change makes the tests stable.

### DIFF
--- a/karma.conf.coffee
+++ b/karma.conf.coffee
@@ -12,7 +12,7 @@ module.exports = (config) ->
 		captureConsole: true
 
 	karmaConfig.plugins.push(require('karma-env-preprocessor'))
-	karmaConfig.preprocessors['**/*.spec.coffee'] = [ 'env' ]
+	karmaConfig.preprocessors['**/*.spec.coffee'] = [ 'browserify', 'env' ]
 	karmaConfig.envPreprocessor = [
 		'RESINTEST_EMAIL'
 		'RESINTEST_PASSWORD'
--- a/package.json
+++ b/package.json
@@ -59,7 +59,7 @@
     "resin-errors": "^2.4.0",
     "resin-pine": "^4.0.0",
     "resin-register-device": "^2.1.1",
-    "resin-request": "^6.1.0",
+    "resin-request": "^6.2.0",
     "resin-token": "^3.0.0",
     "semver": "^5.1.0"
   }
--- a/tests/integration.spec.coffee
+++ b/tests/integration.spec.coffee
@@ -41,6 +41,8 @@ _.assign opts,
 	apiKey: null
 	isBrowser: IS_BROWSER,
 	isTest: true
+	debug: true
+	retries: 3
 
 pine = getPine(opts)
 resinRequest = getResinRequest(opts)
--- a/tests/integration.spec.coffee
+++ b/tests/integration.spec.coffee
@@ -41,7 +41,6 @@ _.assign opts,
 	apiKey: null
 	isBrowser: IS_BROWSER,
 	isTest: true
-	debug: true
 	retries: 3
 
 pine = getPine(opts)
--- a/tests/util.coffee
+++ b/tests/util.coffee
@@ -1,2 +1,2 @@
 exports.loadEnv = ->
-	require('dotenv').config()
+	require('dotenv').config(silent: true)


jviotti:
I see the following in the CI server logs:

```
{ Error: ENOENT: no such file or directory, open '.env'
  at Error (native)
  at Object.fs.openSync (fs.js:640:18)
  at Object.fs.readFileSync (fs.js:508:33)
  at Object.config (/home/travis/build/resin-io/resin-sdk/node_modules/dotenv/lib/main.js:30:37)
  at exports.loadEnv (/home/travis/build/resin-io/resin-sdk/tests/util.coffee:2:20)
  at module.exports (/home/travis/build/resin-io/resin-sdk/karma.conf.coffee:6:2)
  at Object.parseConfig (/home/travis/build/resin-io/resin-sdk/node_modules/karma/lib/config.js:377:5)
  at new Server (/home/travis/build/resin-io/resin-sdk/node_modules/karma/lib/server.js:56:20)
  at Object.exports.run (/home/travis/build/resin-io/resin-sdk/node_modules/karma/lib/cli.js:280:7)
  at Object.<anonymous> (/home/travis/build/resin-io/resin-sdk/node_modules/karma/bin/karma:3:23)
  at Module._compile (module.js:570:32)
  at Object.Module._extensions..js (module.js:579:10)
  at Module.load (module.js:487:32)
  at tryModuleLoad (module.js:446:12)
  at Function.Module._load (module.js:438:3)
  at Module.runMain (module.js:604:10)
  at run (bootstrap_node.js:394:7)
  at startup (bootstrap_node.js:149:9)
  at bootstrap_node.js:509:3
 errno: -2, code: 'ENOENT', syscall: 'open', path: '.env' }
```

Is that expected?

jviotti:
Notice that the debug logs are currently causing the authentication token to be printed in the CI servers, which means that anyone looking can get access to our testing accounts.

emirotin:
Can you edit the PR and make a PR _to my branch_ please?

emirotin:
> I see the following in the CI server logs:

Heh, it's because of `dotenv` module. I didn't expect it to do so and expected it to silently do nothing if the env file is missing. And also I didn't see this message before.

karma.conf.coffee
```diff
@@ -12,7 +12,7 @@ module.exports = (config) ->
 		captureConsole: true
 
 	karmaConfig.plugins.push(require('karma-env-preprocessor'))
-	karmaConfig.preprocessors['**/*.spec.coffee'] = [ 'env' ]
+	karmaConfig.preprocessors['**/*.spec.coffee'] = [ 'browserify', 'env' ]
```
emirotin: aha

pimterry:
@emirotin Base branch changed.

@jviotti Yes, you're right, that's not good.

What should we do about this? Since it does initially look like this is now stable I'm no longer that worried about further debugging, so I could turn debugging off, and we can change the password. That doesn't totally solve the preexisting problem though - presumably anybody could submit a PR to this project that turns on debug, and immediately grab any new password we use?

jviotti:
@pimterry PRs from forks will not have access to the secure environment variables we put in the CI servers, and we can safely trust people with write access to these repos (whose PRs can read the secure environment variables), since they are the resin.io team.

emirotin:
Let's remove `debug` from this PR, merge it (it just goes to my branch), then figure out how to hide the token from the debug output.

emirotin:
Ah, and also wrap the `dotenev.config` in try/catch probably?

emirotin:
Hm, it shouldn't actually throw, it should be a warning, and

> Dotenv outputs a warning to your console if missing a .env file. Suppress this warning using silent.
> require('dotenv').config({silent: true});

pimterry:
Debug removed, dotenv silenced.

pimterry:
Also @jviotti you're totally right about the hiding secure vars thing [for Travis](https://docs.travis-ci.com/user/environment-variables/#Defining-encrypted-variables-in-.travis.yml), but the docs for Appveyor don't mention anything about that. They make it sound like it's all PRs or none for them, and you should only turn on vars in PRs for private repos: https://www.appveyor.com/docs/build-configuration/#secure-variables. Haven't tested though, is it just that the docs don't tell the whole story?

jviotti:
@pimterry Looks like a docs-induced confusion :) I use secure env vars in Appveyor for some projects, and under settings, I see:

screenshot 2016-12-14 13 02 12

Which is exactly what we need.

emirotin:
CI tests seem to be hung? No activity in Travis

pimterry:
Not sure why Travis has hung, but it's queued in their infrastructure, not our code, so there's not much we can do. We're in the queue. There were a few pushes in quick succession for this, and for each one we seem to trigger two builds (the push on the branch and the PR itself), on both Travis & Appveyor, each of which runs two jobs (for two node versions), each of which runs the entire test suite for Node and the Browser, twice (because of `prepublish`). They're currently taking 50+ minutes, per job, so our backlog might take a while.

I could kill jobs for all commits but the very latest batch, but I quite want to let the builds all run, just to confirm they're definitely now stable.

I'm not sure why we're notifying on branches and PRs and duplicating these builds. Could we stop doing that?

It'd also be good to find a way to not run the whole 20 minute test suite twice with each build (as discussed on Flowdoc). Sounds like that's not practical for now, but we need to be a little careful - if the builds go from the current ~50 minutes to 60 then we'll hit the Travis time limit, and the builds won't run at all.

pimterry:
Interestingly, one of the builds has failed: https://ci.appveyor.com/project/resin-io/resin-sdk/build/1.0.938/job/4w6t2nxcoxtcyg4g#L2414. No longer because of ETIMEDOUT, but because:

```
PhantomJS 2.1.1 (Windows 8 0.0.0) SDK Integration Tests given a logged in fresh user given a single application with two offline devices that share the same uuid root "before each" hook for "should be rejected with an error if there is an ambiguation between shorter uuids" FAILED
Request error: Data is referenced by uuid.
c:/Users/appveyor/AppData/Local/Temp/1/node_modules/resin-request/build/request.js:160:0
```

I'm pretty sure I actually know what this is: I'm running tests locally, and that test just failed randomly for me too, I think because it uses fixed device uuids.

Presumably device uuids must be unique across all devices, not just per-user? That would mean these tests will fail if anybody happens to be running those same few tests at the same time anywhere, which isn't great. Separate to the ETIMEDOUT issue, and unlikely to come up often normally, but still pretty messy.

emirotin:
Yeah, tests are not isolated and it's not great at all. Ideally we'd like to run each of them against of a separate fresh DB :(
UUIDs should be globally unique, and it looks like the tests try to delete a record that's FKed by something else.

pimterry:
> UUIDs should be globally unique, and it looks like the tests try to delete a record that's FKed by something else.

Autogenerated ones will be, but the test explicitly provides fixed UUIDs: https://github.com/resin-io/resin-sdk/blob/af90835dee327bb450239bc768aeea56b09cdc6d/tests/integration.spec.coffee#L1129-L1130. I was assuming that that's why they're conflicting: both create the same device, and end up failing at some point because of the existing FK relationships (not sure where, maybe when you create new a device with an existing UUID but within a different application it explodes? Sounds like it should imo). Does that sound plausible?

I think we can find a way to isolate these though, at the cost of some extra complexity (assuming the above is right). I'll take a look tomorrow - let me know if I'm totally off track or if this isn't worth fixing.

emirotin:
The "data is referenced" is when you try to delete as far as I remember, but maybe it's also thrown when you violate the unique constraint? @Page- any ideas?
Generally, we could just generate a unique session ID and prepend it to every unique field we inject into the DB.

I'll merge this one now into my branch.

Page-:
The "data is referenced" is when you try to delete something that is referenced by a foreign key, as far as I'm aware that's the only case

PR #240, opened 2016-12-14
Update to support the new device registration flow

### LINKED ISSUE #232: Decide on the new device registration process
It (starting from `resin-register-device@3`) needs a provisioning key which will make it unusable for the ghost devices creation used by the CLI.

Decision to be made after we publish the browser-compat SDK.

NB: it's easy to prohibit this method in the browser, but
a) we need the provisioning keys for node.js tests
b) this method is used for inserting fixtures for other tests valid for the browser, like moving the device

pimterry:
This PR fixes #232 by pulling in resin-register-device@3 (see resin-io/resin-register-device#22 and resin-io/resin-register-device#25 for background).

This PR brings the resin-sdk API in line with resin-register-device, and that adds the various changes to the tests required to make this work. Device registration now works in Node and browsers. Also clarifies some tests and reenables some tests that were [temporarily disabled 2 months ago](https://github.com/resin-io/resin-sdk/blob/af90835dee327bb450239bc768aeea56b09cdc6d/tests/integration.spec.coffee#L1210), and now pass.

I've favoured internal consistency for this change over backward compatibility here, which is debatable. Breaking changes are:

* generateUUID is now generateUniqueKey, and is no longer async
* register no longer returns the device, now just `{ id, uuid, api_key }`).

These are more or less the same APIs as resin-register-device. We could transform these here to make this backward compatible (wrapping generateUniqueKey to make it async; doing an extra request after every register call to get the device and then attaching api_key to that result) but both of those changes are a little messy internally and have downsides. _If_ we're happy breaking compatibility, I think this PR is the better end result.

Other minor point is that I've called the new `{ id, uuid, api_key }` return type "device registration info", but I haven't described the format anywhere, so it's not easy for users to know exactly what'll be in that. Can't see any examples of how we document return object types in JSDoc, let me know if there's a clear way I should add more detail.

This PR is on top of #239, because otherwise the tests won't pass, and I've used that as a base for this PR so you can see the changes here independently. I'll can shift it to `sdk-browser` once the tests PR has been merged.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -3,6 +3,11 @@
 All notable changes to this project will be documented in this file.
 This project adheres to [Semantic Versioning](http://semver.org/).
 
+### Changed
+
+- **Breaking!** Changed `register()` to work with the new device registration flow. `register()` now return device registration information (`{ id: "...", uuid: "...", api_key: "..." }`), but not the full device object.
+- **Breaking!** Changed `generateUUID()` to `generateUniqueKey()` to reflect that it should now be used for both generating a uuid and an api key.
+
 ## [5.4.0] - 2016-10-27
 
 ### Added
--- a/lib/models/device.coffee
+++ b/lib/models/device.coffee
@@ -43,7 +43,7 @@ CONTAINER_ACTION_ENDPOINT_TIMEOUT = 50000
 
 getDeviceModel = (deps, opts) ->
 	{ pine, request } = deps
-	{ apiUrl, isBrowser, isTest } = opts
+	{ apiUrl } = opts
 
 	configModel = once -> require('./config')(deps, opts)
 	applicationModel = once -> require('./application')(deps, opts)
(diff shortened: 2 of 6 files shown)


emirotin:
The tests PR is merged, good to rebase.

CHANGELOG.md
```diff
@@ -3,6 +3,11 @@
 All notable changes to this project will be documented in this file.
 This project adheres to [Semantic Versioning](http://semver.org/).
 
+### Changed
+
+- **Breaking!** Changed `register()` to work with the new device registration flow. `register()` now return device registration information, but not the full device object.
```
emirotin: It's good to mention the format at least here, like `device registration information ({ ... })`

lib/models/device.coffee
```diff
@@ -1052,40 +1045,36 @@ getDeviceModel = (deps, opts) ->
 	# @param {String} applicationName - application name
 	# @param {String} uuid - device uuid
 	#
-	# @fulfil {Object} - device
+	# @fulfil {Object} - device registration info
```
emirotin: again can be documented at least here after the `info`

lib/models/device.coffee
```diff
@@ -1052,40 +1045,36 @@ getDeviceModel = (deps, opts) ->
...
-	#
-	# 	resin.models.device.register('MyApp', uuid, function(error, device) {
-	# 		if (error) throw error;
-	#
-	# 		console.log(device);
-	# 	});
+	# 	console.log(registrationInfo);
 	# });
 	###
-	exports.register = onlyIf(not isBrowser or isTest) (applicationName, uuid, callback) ->
```
emirotin: As far as I remember I've introduced `isTest` only to workaround this specific limitation. If true good to remove it from everywhere

emirotin:
We should remember to bring the registration-related changes in the CLI when we upgrade the SDK version there.
I think the CLI doesn't have tests ATM.

emirotin:
Can you please rebase and resolve conflicts as I've squashed the commits in the sdk-browser branch?

pimterry:
I've pulled in all your comments and rebased.

emirotin:
LGTM, please squash before merging.

emirotin:
Oh BTW you need to rebuild the docs with `npm run docs`

pimterry:
Squashed down to just two commits: the actual changes of this PR, and a separate one for the re-enabling the CORS test, since that's a totally separate change, and it'd be good to be able to follow that back in future.

PR #242, opened 2016-12-15
Safely stop tests and build running on npm install

pimterry:
This halves the number of tests we run, working towards #241.

This script makes NPM `prepublish` never run the tests or CS build. Instead, those are run in NPM@4's new `prepublishOnly` step, which _only_ happens on an actual publish, not on simple `npm install`'s.

To make this safe, we have to catch any users that aren't using NPM@4, but are trying to publish - `require-NPM4-to-publish` does that, and gets triggered by any actual publishes when they trigger `prepublish`.

Tested locally with NPM 4.0.3 and NPM 3.10.10.

### DIFF
--- a/package.json
+++ b/package.json
@@ -23,8 +23,8 @@
     "test-node": "gulp test",
     "test-browser": "karma start",
     "build": "gulp build",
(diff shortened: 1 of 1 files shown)


pimterry:
Sure, done.

PR #247, opened 2016-12-21
Switch to using a module for requiring npm4, which means we can publish on windows again

Page-:


package.json
```diff
@@ -23,7 +23,7 @@
     "test-node": "gulp test",
     "test-browser": "karma start",
     "build": "gulp build",
-    "prepublish": "(not-in-publish && echo 'Skipping prepublish') || npm run require-npm4-to-publish",
+    "prepublish": "require-npm4-to-publish",
     "require-npm4-to-publish": "semver -r '>=4.0.0' $(npm --version) || (echo 'NPM 4+ required to publish, run: sudo npm install -g npm' && exit 1)",
```
lurch: Does this line need to be removed as well, now that the logic has been moved to an external module?

Page-: Definitely, fixed

package.json
```diff
@@ -39,14 +38,14 @@
...
     "jsdoc-to-markdown": "^2.0.1",
     "karma": "^1.3.0",
     "karma-env-preprocessor": "^0.1.1",
     "mocha": "^3.0.0",
     "mochainon": "^1.0.0",
     "request": "^2.67.0",
+    "require-npm4-to-publish": "^1.0.0",
     "resin-config-karma": "^1.0.4",
     "resin-settings-client": "^3.5.2",
     "rindle": "^1.2.0",
```
Page-: The semver dependency already existed before your PR and is a runtime dependency so I think it's needed - although I didn't grep for it

Page-: Just checked, it's used for checking if the supervisor of a device is compatible with certain features

pimterry: Yes you're totally right: https://github.com/resin-io/resin-sdk/blob/2f39bdb9386b440120d97cb65e885a8b46bbb174/lib/models/device.coffee#L20

PR #259, opened 2017-01-12
return resin-settings-client to dependencies

### LINKED ISSUE #256: Browser version fails throws exception with webpack HMR.
I'm giving the new browser version a spin with a react and running into an exception when Hot module replacement is active. 

```
Error in ./~/resin-sdk/build/resin.js
Module not found: 'resin-settings-client' in /Users/gaudi/work/resin-1984/node_modules/resin-sdk/build

 @ ./~/resin-sdk/build/resin.js 85:15-47
```

Still trying to debug, but if you have any ideas, send em my way.

@pimterry @emirotin

emirotin:
Fix #256

It's safe to have it there. The module is not used internally but exposed as part of the SDK in Node.js. 
For the browser builds we do drop it explicitly.

UPD: I've found a bunch of device methods that were expecting the presence of `settings` instead of using the `apiUrl` passed in. Fixed in the 2nd commit.
These methods are not in tests so we didn't capture the errors.


yarn.lock
```diff
@@ -1092,7 +1092,13 @@ debug@0.7.4:
   version "0.7.4"
```
Page-: Is it possible to get rid of the noise of this file?

build/resin-browser.js
```diff
@@ -1599,7 +1599,7 @@ getDeviceModel = function(deps, opts) {
         return request.send({
```
Page-: Is it possible to get rid of the noise of this file?

yarn.lock
emirotin: You mean make the github hide the contents? I'm not sure. It shouldn't be edited by hand. So looks like the only way is to ask GH to update their library to detect it and hide the diff by default.

build/resin-browser.js
emirotin: Need to check again how GH detects the generated files and add some line that will please it.

emirotin:
Will be published as v6.0.0-beta2 in about an hour (tests running locally atm).

Issue #262, opened 2017-01-31
Support all the PineJS options


emirotin:
Like extra filtering, ordering, selecting specific fields, eager-loading, etc.
This makes sense to all the methods that _get_ multiple resources, like `application.getAll`.
I'd just allow passing the entire options as a single object and merging them properly (1 thing to be careful with is the $filter)

emirotin:
Looks like resolved?

PR #263, opened 2017-02-01
Allow ids in addition to names

### LINKED ISSUE #243: Switch from names to IDs
Right now the SDK operates on many resources by using human-readable things like app names and device uuids / names.

That makes sense for the CLI because they're way easier to type and recognize for the end users.
But doesn't make much sense to the programmatic usage. For example, the UI already operates on IDs and using the names means extra request round-trips.

We should make the SDK target the programmatic usage and accepts the IDs everywhere as the primary way to identify things.

The name -> ID conversion could live in a separate layer inside of the SDK, or can be moved to the CLI code (potentially as a compatibility wrapper around the SDK instead of patching all the existing method calls).

This should be handled as a separate major version after we merge the initial support for the browser.

pimterry:
This fixes #243, by overloading every method that takes an device uuid or application name as a string to also take a device or application id if a number is passed.

This is quite a large PR. I've added new tests so far as possible, but there's quite a few touched methods that don't have any, sadly, so careful review would be very helpful.

This has basically been a review of the entire SDK codebase, and there's a few notes that have come up:

- The current docs are a little messy (they duplicate the list of all the models, for example). Way worse though, running `npm run docs` now generates a totally different file for me to the existing docs, both on this branch but also on master.

  The new file duplicates the whole api over and over again, generating very broken docs that about 20x the size of the current ones. Compare: [commited](https://raw.githubusercontent.com/resin-io/resin-sdk/master/DOCUMENTATION.md) / [new](https://raw.githubusercontent.com/resin-io/resin-sdk/243-use-ids/DOCUMENTATION.md).

  This definitely needs fixing. We did [recently bump](https://github.com/resin-io/resin-sdk/commit/2b7d79420f5a0a1dac8a392f4510d63af653ea7d) to jsdoc-to-markdown 2.0, but downgrading doesn't seem to fix this. I don't know anything about jsdoc-to-markdown - any idea what could have caused this? Happy to investigate further if nothing springs to mind.

- The overloading strategy works fine, and doesn't break what the docs said, but [one of the tests](https://github.com/resin-io/resin-sdk/blob/a2beb8988285c9e8cac39a4d81cd035226b1585d/tests/integration.spec.coffee#L1087-L1089) was actually explicitly tested that passing as a number also worked. Is that a concern?

- Quite a few methods do a seemingly unnecessary `application/device.get`. E.g. application.remove, restart, almost every environmentVariable method, etc. See [Application.remove](https://github.com/resin-io/resin-sdk/blob/243-use-ids/lib/models/application.coffee#L274-L279) for a good example.

  In cases like these, we could put together a filter for pine for both name and id cases fairly easily using just the `nameOrId` argument, to skip the upfront request entirely, and directly do the delete from there. That'd be nice once we get this in the UI, and for API load generally. Is there any reason why we need to keep this? It might return a different format of error if the device doesn't exist, but we should be able to work around that.

- Device.restart is deprecated. Is the upcoming real 6.0.0 release a good opportunity to kill it?

- The docs and tests for `.getDeviceStatus` have been wrong since they were originally implemented. Docs, tests and examples say it takes a device uuid, but actually it takes a full device object. The tests happened to pass because a uuid string is duck-type equivalent to an offline device object :smile:.

  I've now fixed this - is there anything else we need to check is using this correctly? I've already looked at the cli, which does ignore the docs and use this correctly.


jviotti:
> I don't know anything about jsdoc-to-markdown - any idea what could have caused this? Happy to investigate further if nothing springs to mind.

Oh, that's weird. Did you downgrade using npm's `^`/`~`? Maybe there is a minor version increment that broke this and npm is automatically installing the latest one that matches the ranges.

> The docs and tests for .getDeviceStatus have been wrong since they were originally implemented. Docs, tests and examples say it takes a device uuid, but actually it takes a full device object. The tests happened to pass because a uuid string is duck-type equivalent to an offline device object 

lol, good catch!

jviotti:
> This fixes #243, by overloading every method that takes an device uuid or application name as a string to also take a device or application id if a number is passed.

Any need to keep supporting ids for devices? Seems like just uuids is the most reasonable way to go unless I'm missing some use case for ids?

pimterry:
> Oh, that's weird. Did you downgrade using npm's ^/~? Maybe there is a minor version increment that broke this and npm is automatically installing the latest one that matches the ranges.

@jviotti I've been manually installing fixed versions. I've tried 1.1.1 (our previous dep), 1.3.8 (latest 1.X) and 2.0.1 (latest), and none of them come up with anything like our current docs. 1.1.1 actually just produces an empty file. I still don't really know why.

What version have you got installed at the moment? If you build the docs yourself on master, can you reproduce this? What about if you delete node_modules, npm install, and build them?

> Any need to keep supporting ids for devices? Seems like just uuids is the most reasonable way to go unless I'm missing some use case for ids?

I'll have to hand points on motivation to @emirotin. My understanding is that the UI already has ids for everything everywhere, and it's inconvenient to have to go off and get application names/device uuids for any API requests we want to do anywhere.

jviotti:
@pimterry I'm indeed able to reproduce in `master`. I went back to v5.4.0, which the last version I released myself and couldn't reproduce it there. Bisecting led me to the following commit: https://github.com/resin-io/resin-sdk/commit/4e73a2c2c371514fc990c90e55a845a9d0259c38

This commit builds the whole SDK into `build/resin-browser.js` and `build/resin-browser.min.js`. Notice how the `docs` npm script looks like:

```js
"docs": "jsdoc2md build/{,**/}*.js > DOCUMENTATION.md"
```

So the issue is that `jsdoc2md` is grabbing those files as well (the non minified version precisely, which includes all the JSDoc annotations).

emirotin:
> I'll have to hand points on motivation to @emirotin. My understanding is that the UI already has ids for everything everywhere, and it's inconvenient to have to go off and get application names/device uuids for any API requests we want to do anywhere.

Yeah, IDs are the "true" PK/FKs in the API and the UI is built around this assumption.
As we use them for every other resource (for operations like getting an entry by PK or loading the related entry by FK) it makes sense to uniformly support them for the devices as well.

UUIDs (and names) are good for the end users as the brain can parse them and find some recognizable patterns (use last 3 letters of the hex UUID for example), so they make sense for the manual entry. They're also good because making a typo (or losing 1 char) will likely lead to not found error instead of affecting a different object.
But for automated systems IDs are a natural choice.

lib/models/application.coffee
```diff
@@ -237,24 +257,25 @@ getApplicationModel = (deps, opts) ->
...
+	#
+	# @example
 	# resin.models.application.remove('MyApp', function(error) {
 	# 	if (error) throw error;
 	# });
 	###
-	exports.remove = (name, callback) ->
-		exports.get(name).then ->
+	exports.remove = (nameOrId, callback) ->
+		exports.get(nameOrId).then (application) ->
```
emirotin: Can save a round-trip (would be beneficial for the UI):

```
Promise.try ->
  if isId(nameOrId) return nameOrId
  exports.get(nameOrId).get('id')
.then (appId) ->
  pine.delete ...
.asCallback(callback)
```

emirotin: Same for all the methods that call `get` to get the ID.
Can of course keep it DRY and move these lines to a separate method `getId(nameOrId)` and just call it everywhere

emirotin: Another note: to reduce the load on the API and the amount of data flying over the network we should add support for extra PineJS options (mentioned already in a different thread) and in this specific case pass `$select: 'id'` which will return an object with ID only.
We should intro this optimization to other places that only need the specific props of the resource.
Of course this can go into a separate PR that will implement the custom props in the first place.

lib/models/build.coffee
```diff
@@ -40,23 +40,25 @@ getBuildModel = (deps, opts) ->
...
 	#		console.log(builds);
 	# });
 	###
-	exports.getAllByApplication = (name, callback) ->
-		applicationModel().has(name).then (hasApplication) ->
-
-			if not hasApplication
-				throw new errors.ResinApplicationNotFound(name)
+	exports.getAllByApplication = (nameOrId, callback) ->
+		applicationModel().get(nameOrId).then (application) ->
```
emirotin: The for non-existent app is dropped because it's handled by the `get` method, right?

lib/models/device.coffee
```diff
@@ -1305,25 +1430,25 @@ getDeviceModel = (deps, opts) ->
 	# @public
 	# @function
 	# @memberof resin.models.device
-	#
-	# @param {String} uuid - device uuid
-	# @fulfil {String} - device statud
+	#π
```
emirotin: pi?

lib/models/application.coffee
pimterry: Yep, see my original PR notes above. Sounds like there isn't a good reason for this then - I'll remove it.

I'm going to do the PineJS options part as a second PR, because this one is enormous enough already.

lib/models/build.coffee
pimterry: Yes, exactly. `has` is actually just `get()`ing the application and then suppressing the exact same exception to return a boolean, so we might as well do it more directly.

lib/models/device.coffee
pimterry: I... That's very weird. It appears alt-p on my machine prints pi, and I must've pressed that by accident? Now fixed.

🍰

pimterry:
Thanks @jviotti! Good find. I've now fixed this by excluding the browser builds from JSDoc, and rebuilt the docs. This now works correctly - if we'd prefer we could alternatively move the browser builds somewhere other than `build/` instead, but I'm not going to unless anybody shouts because that's more work.

lib/models/application.coffee
emirotin: Oh sorry skipped that. Unless Juan knows a good reason — good to remove I think.
Different errors for not found are probably OK.

But there's a separate question about the multiple found entries.
Both app names and device UUIDs are unique for the given user. But the user can have access (and thus see) two apps of the same name if one of them is shared with the user. What will the SDK do in this case? According to the code (`.get(0)`) it will just randomly pick the first of them which sounds like an issue to me.

pimterry: Ah, I wondered how that case could happen. It should be ok though, in both the previous code and this PR, `.get` checks for multiple results first, and throws a `ResinAmbiguousApplication` exception before the `.get(0)`.

emirotin: Ah, good.

pimterry:
Phew, all done. To reiterate and summarize, this PR now:

* Fixes the doc builds
* Corrects the docs for .getDeviceStatus
* Removes `models.device.restart`
* Overloads every method that takes an uuid or app name to also take a device/app id
* Avoids making unnecessary requests just to get ids in the cases where that's possible (there are a few ones where extra request check is necessary though, because we get `[]` from the api for a bad id, not an error, e.g. [builds.getAllByApplication](https://github.com/resin-io/resin-sdk/blob/2b9fe1fb86ec3ebcdee4ef924faac33a7c56a5cd/lib/models/build.coffee#L53))

If somebody could review and comment/approve when they're happy, I'll get this merged.

alexandrosm:
@pimterry -- is this last one perhaps a bug you should file with pine.js?

-- 

*Alexandros Marinos*

Founder & CEO, Resin.io


+1 206-637-5498

@alexandrosm

emirotin:
Yeah, in my understanding Pine's filter roughly translates to `WHERE application = x` and the DB happily returns an empty response.

pimterry:
Yeah, it's that we're filtering the set of builds (for example) by application id, not finding an application and then finding the builds from it. In that model it makes perfect sense that we get `[]` instead of a 404, and I'm not sure if pine supports a different approach to this, or what the implications of that would be. I think it's fine as is for now.

pimterry:
I'm still waiting for somebody to approve this if you're happy with all the above. @emirotin @Page- @jviotti

lib/models/application.coffee
```diff
@@ -237,24 +267,26 @@ getApplicationModel = (deps, opts) ->
...
-		exports.get(name).then ->
+	exports.remove = (nameOrId, callback) ->
+		getId(nameOrId).then (applicationId) ->
 			return pine.delete
 				resource: 'application'
-				options:
-					filter:
-						app_name: name
+				id: applicationId
+		.catch(treat404AsMissingApplication(nameOrId))
```
Page-: I think this would be nicer as something like:
```coffee
exports.catch404 =
	code: 'ResinRequestError'
	statusCode: 404

exports.treatAsMissingApplication= (nameOrId) ->
	return (err) ->
		replacementErr = new errors.ResinApplicationNotFound(nameOrId))
		replacementErr.stack = err.stack
		throw replacementErr
...
.catch(catch404, treatAsMissingApplication(nameOrId))
```

This does a few things
* it separates the predicate for what you're catching from the conversion, which means they can be composed separately
* it lets us use bluebird sugar for the predicate, including avoiding the need to rethrow
* it lets us create the error object lazily on demand rather than immediately, meaning if it never happens we don't have to pay the cost of creating that error - notably the cost of constructing a stack trace.

lib/models/build.coffee
```diff
@@ -40,23 +40,25 @@ getBuildModel = (deps, opts) ->
...
 
 			return pine.get
 				resource: 'build'
 				filter:
 					application: $any:
```
Page-: I think it you can change this to `application: application.id` now that it's using id

lib/models/application.coffee
```diff
@@ -28,6 +31,15 @@ getApplicationModel = (deps, opts) ->
 
 	exports = {}
 
+	# Internal method for name/id disambiguation
+	# Note that this throws an exception for missing names, but not missing ids
+	getId = (nameOrId, callback) ->
+		(if isId(nameOrId)
```
Page-: This might be nicer as
```coffee
Promise.try ->
	if isId(nameOrId)
		return nameOrId
	else
		exports.get(nameOrId).get('id')
.asCallback(callback)

Comment (pimterry) [inline lib/models/application.coffee]: Agreed, done.

File: lib/models/build.coffee (hunk 5 above)
Comment (pimterry) [inline lib/models/build.coffee]: Agreed, done.

File: lib/models/application.coffee (hunk 4 above)
Comment (pimterry) [inline lib/models/application.coffee]: Agreed, done.

File: lib/models/application.coffee [hunk 7]
```diff
@@ -28,6 +31,16 @@ getApplicationModel = (deps, opts) ->
 
 	exports = {}
 
+	# Internal method for name/id disambiguation
+	# Note that this throws an exception for missing names, but not missing ids
+	getId = (nameOrId, callback) ->
```
emirotin: It's private so the callback is likely not needed?

lib/models/application.coffee
```diff
@@ -301,13 +338,20 @@ getApplicationModel = (deps, opts) ->
...
 	# 	if (error) throw error;
 	# 	console.log(apiKey);
 	# });
 	###
-	exports.getApiKey = (name, callback) ->
-		exports.get(name).then (application) ->
+	exports.getApiKey = (nameOrId, callback) ->
+		# Do a full get, not just getId, because the actual api endpoint doesn't fail if the id
+		# doesn't exist. TODO: Can use getId once https://github.com/resin-io/resin-api/issues/110 is resolved
+		exports.get(nameOrId).then (application) ->
```
emirotin: Would be nice to pass the `$select: 'id'` option to speedup things.
Would also rename the method to `generateApiKey` or `createApiKey`, the name is confusing.

pimterry: I've renamed the method, and added a deprecated `getApiKey` which we can kill off later.

I'm going to leave tweaks like `$select: id` out for now. There are (and were before this PR) quite a few places here where you could potentially make optimisations like that to reduce the data transfer, but I don't think they're necessary immediately, and they'll be easier to add once #262 is supported - I'll do this along with that. Tracked as #267.

lib/models/application.coffee
pimterry: Good point, done.

PR #270, opened 2017-02-22
Expose request and token instances

pimterry:
In the UI, we want to be able to set the token used inside the SDK, and to repeat individual failed requests using resin-request. This is possible by depending directly on `resin-request` and `resin-token`, and instantiating instances of that in exactly the same way as the SDK, but it's error-prone and messy. Being able to directly share an instance makes this much cleaner.

I think this is the best way to do that (the other option would be to optionally take a resin-request/resin-token instance in the SDK constructor, and build them all separately downstream), but it's a little weird, and definitely debatable.

I haven't exposed `.pine`, only `.token` and `.request`. Mostly just because I don't need resin-pine, but I could do that too if people think it's worthwhile.


emirotin:
`.pine` might be useful for quickly doing some very custom requests without (or actually before) sending a PR to the SDK

pimterry:
I can't really actually merge this yet, because it's dependent on https://github.com/resin-io-modules/resin-request/pull/94 which still needs review. @emirotin can you take a look at that please?

PR #271, opened 2017-02-24
Add getAllByApplication and normalize device envvar name formatting

pimterry:
This adds `models.environmentVariables.device.getAllByApplication` to the SDK, taking it from the UI which had previously implemented this itself standalone.

I think this is worth being a built-in member of the SDK because it's a fairly useful operation in itself, and because it's nicely consistent with `models.environmentVariable.device.getAllByApplication`, and the many `getAllByApplication` methods elsewhere too.

This PR also changes the return format from `environmentVariables.device.create` and `.update` slightly. Previously `.getAll` returned variables with with `.name` rather than `.env_var_name` (as returned by the API), but `.create` and `.update` didn't. With this change it's now always `.name` everywhere. This is a little tenuous, and maybe we should be doing this in the API instead? Open to that, but I think this is an easier first step for now regardless, especially since we're already in beta for a breaking SDK release soon.


emirotin:
The API has some limitation that didn't allow the property to be named `name`. So it's OK to normalize here for now IMO.

lib/models/environment-variables.coffee
```diff
@@ -214,16 +224,48 @@ getEnvironmentVariablesModel = (deps, opts) ->
...
+	# 	console.log(environmentVariables)
+	# });
+	###
+	exports.device.getAllByApplication = (nameOrId, callback) ->
+		applicationModel().get(nameOrId).get('id').then (applicationId) ->
+			return pine.get
+				resource: 'device_environment_variable'
+				options:
+					filter:
+						device:
```
emirotin: This way of filtering is deprecated, see https://github.com/resin-io/pinejs-client-js/blob/master/core.coffee#L32

pimterry: :+1: Now fixed.

PR #273, opened 2017-03-03
Add a method that can merge default and extra Pine query options

pimterry:
This is the complex part of #262, which is the next big step towards https://github.com/resin-io/resin-ui/issues/400.

I'm opening this PR with just the standalone merging part before putting it into use, since it's a fiddly first step, and big changes to this might affect later work. Obviously this means I haven't tested it fully yet, since I haven't integrated it, but I'm reasonably confident this works as is.

Comments and tests should hopefully explain the behaviour pretty well. The core thing that makes this complicated is the different field-by-field semantics, as I've been trying to chase intuitive sensible behaviour for various extra provided options. I think that, for example:

* `builds.getAllByApplication('myApp', select: ['id'])` means 'only return the id', not 'in addition to the normal fields, make sure you also return the id'
* `builds.getAllByApplication('myApp', filter: id: 4)` means 'in addition to the normal filtering, make sure the id is 4' not 'ignore all filtering, and only check that the id is 4'
* `builds.getAllByApplication('myApp', expand: 'application')` means 'in addition to the other relationed entities returned, also return the application' not 'dont return any related entities except the application'

Apparently unlimited depth of `expand` with nested further `$expand` and `$select` options makes all this quite fiddly, but I think the end result is _fairly_ intuitive. Opinions extremely welcome here, as this is very subjective.

I've included rules for all the relevant pine options I can find or those that I'm aware of, and done my best to work out how `$expand` etc work based on the code I can find using it, but do shout if there's something there I've missed.


lib/util.coffee
```diff
@@ -23,3 +27,73 @@ exports.treatAsMissingDevice = (uuidOrId) ->
...
+				if defaults.filter
+					result.filter = $and: [defaults.filter, value]
+				else
+					result.filter = value
+
+			when 'expand'
+				result.expand = mergeExpandOptions(defaults.expand, value)
+
+			else
+				throw new Error("Unknown pine option: #{option}")
```
Page-: It might be better to just let pinejs-client handle unknown options, that way we can automatically support new options without having to make changes here

lib/util.coffee
```diff
@@ -23,3 +27,73 @@ exports.treatAsMissingDevice = (uuidOrId) ->
...
+
+# Converts a valid expand object in any format into a new object
+# containing (at most) a $expand and a $select key
+convertExpandToObject = (expandOption) ->
+	if not expandOption?
+		return {}
+	else if isString(expandOption)
+		return "#{expandOption}": {}
+	else if isArray(expandOption)
+		return fromPairs(expandOption.map((key) -> [key, {}]))
```
Page-: Do we usually rely on `Array.map` or `lodash/map`?

tests/util.spec.coffee
```diff
@@ -0,0 +1,119 @@
...
+
+	it 'combines expand options for separate single relationships', ->
+		result = mergePineOptions
+			expand: 'device'
+		,
+			expand: 'application'
+
+		m.chai.expect(result).to.deep.equal
+			expand:
+				device: {}
```
Page-: Is it actually valid to pass an empty `expand` object to pinejs-client (I honestly don't remember - it's not something you'd normally do by hand, but makes sense in this auto-generated form)

pimterry: I haven't tested this in depth, but yes, it seems to at first glance.

lib/util.coffee
pimterry: I think we're trying to avoid pulling in too much lodash, and every target runtime should have array.map now

Page-: Yeah, it's more if we're already pulling it in - if not then that's fine

lib/util.coffee
pimterry: My concern is that an arbitrary default for merging of new options (presumably overwriting?) might not be correct and fixing it, if it's not, will be a breaking change later on. If we reject unknown options now then we're free to later add them with the appropriate handling as a minor bump instead, which will be nicer for everybody.

Page-: The way I saw it was that for brand new options resin-sdk won't be using them yet at all, so the correct merge approach would be a simple assign.  And then at the point we update the minimum dep version and start using them then we would add any special case handling needed here as well

pimterry: > at the point we update the minimum dep version and start using them then we would add any special case handling needed here as well

But when we do this, if we do need any special case handling, it's always a breaking change. As we start using the sdk more and more (e.g. in the ui, and in etcher), breaking changes will be a reasonable size annoyance in quite a few places (not to mention for sdk end users).

Feels like that's unnecessary - we're allowing things in the api that may or may not come back to cause problems for us, to save the tiny bit of effort to add options to the whitelist here if we want them (if we do ever add any at all). I'd rather be cautious. If we don't have to worry about breaking changes, we can add new options here with impunity and all downstream clients will immediately get those for free, so everybody gets less friction all round.

Also has the benefit of catching you if you try to use options that don't exist, e.g. if you pass in something with a typo or similar.

I think in practice adding more pine options here looks unlikely to come up very often, and I don't hate this, so if you do feel strongly about it we can go down this road, but I'd definitely prefer the cautious option by default.

emirotin:
The assumptions listed in the PR description make total sense to me (and yes, they do highlight some unclearness about the "right" behavior).

lib/util.coffee
```diff
@@ -23,3 +27,73 @@ exports.treatAsMissingDevice = (uuidOrId) ->
...
+
+# Merging two sets of pine options sensibly is more complicated than it sounds.
+#
+# The general rules are:
+# * select, orderby, top and skip override (select this, instead of the default)
+# * filters are combined (i.e. both filters must match)
+# * expands are combined (include both expansions), and this recurses down.
+#   * That means $expands within expands are combined
+#   * And $selects within expands override
+# * Any unknown options 'extra' options throw an error. Unknown 'default' options are ignored.
```
emirotin: NIT: extra `options` word here?

lib/util.coffee
```diff
@@ -23,3 +27,73 @@ exports.treatAsMissingDevice = (uuidOrId) ->
...
+	result = cloneDeep(defaults)
+
+	for own option, value of extras
+		switch option
+			when 'select', 'orderby', 'top', 'skip'
+				result[option] = value
+
+			when 'filter'
+				if defaults.filter
+					result.filter = $and: [defaults.filter, value]
```
emirotin: NIT: we normally do put extra spaces inside of the square brackets

lib/util.coffee
```diff
@@ -23,3 +27,73 @@ exports.treatAsMissingDevice = (uuidOrId) ->
...
+	for own expandKey, extraExpandOptions of extraExpand
+		expandOptions = defaultExpand[expandKey] ||= {}
+
+		if extraExpandOptions.$select
+			expandOptions.$select = extraExpandOptions.$select
+
+		if extraExpandOptions.$expand
+			expandOptions.$expand = mergeExpandOptions(expandOptions.$expand, extraExpandOptions.$expand)
+
+	defaultExpand
```
emirotin: NIT: I prefer explicit `return` for non-one-liners

lib/util.coffee
emirotin: We do already use native methods where they have the proper support, like map, and filter, and reduce.

lib/util.coffee
pimterry: Done

lib/util.coffee
pimterry: Done

lib/util.coffee
pimterry: Done

Issue #274, opened 2017-03-03
resin.models.device.has() now breaks if an 'undefined' parameter is passed


lekkas:
sdk version: `resin-sdk@^6.0.0-beta2`

This is a breaking change that breaks the unpublished [master branch](https://github.com/resin-io/resin-cli/commit/9214cb0d90610968362c567a00eb19af256f611a) of `resin-cli`

Stack:

```
kostasl@macbook:~/resin/github/resin-io-modules/resin-sync (master) $ DEBUG=1 resinstaging ssh
TypeError: Cannot read property 'length' of undefined
    at Object.exports.get (/usr/local/lib/node_modules/resin-cli/node_modules/resin-sdk/build/models/device.js:230:31)
    at Object.exports.has (/usr/local/lib/node_modules/resin-cli/node_modules/resin-sdk/build/models/device.js:425:20)
    at Command.action (/usr/local/lib/node_modules/resin-cli/build/actions/ssh.js:68:34)
    at /usr/local/lib/node_modules/resin-cli/node_modules/capitano/build/command.js:96:21
    at tryCatcher (/usr/local/lib/node_modules/resin-cli/node_modules/bluebird/js/release/util.js:16:23)
    at Promise.successAdapter [as _fulfillmentHandler0] (/usr/local/lib/node_modules/resin-cli/node_modules/bluebird/js/release/nodeify.js:22:30)
    at Promise._settlePromise (/usr/local/lib/node_modules/resin-cli/node_modules/bluebird/js/release/promise.js:564:21)
    at Promise._settlePromise0 (/usr/local/lib/node_modules/resin-cli/node_modules/bluebird/js/release/promise.js:612:10)
    at Promise._settlePromises (/usr/local/lib/node_modules/resin-cli/node_modules/bluebird/js/release/promise.js:691:18)
    at Async._drainQueue (/usr/local/lib/node_modules/resin-cli/node_modules/bluebird/js/release/async.js:133:16)
    at Async._drainQueues (/usr/local/lib/node_modules/resin-cli/node_modules/bluebird/js/release/async.js:143:10)
    at Immediate.Async.drainQueues (/usr/local/lib/node_modules/resin-cli/node_modules/bluebird/js/release/async.js:17:14)
    at runCallback (timers.js:637:20)
    at tryOnImmediate (timers.js:610:5)
    at processImmediate [as _immediateCallback] (timers.js:582:5)
```

pimterry:
Fixed in 6.0.0.

PR #275, opened 2017-03-03
Ensure application/device .has and .get handle null/undefined arguments clearly & consistently

pimterry:
Connects to #274

In the current v6 beta both `device.has` and `device.get` throw a confusing error if you passed null or undefined. Now, device and application .has and .get (used internally by many methods) throw a clear error (e.g. "Application not found: null"), and this automatically allows `.has` to handle this and still return a boolean correctly everywhere.

This now isn't a breaking change for 5.0.0, I believe, because previously it'd return a device not found error anyway. Before v6, we used to turn the uuid into a string (literally `String(uuid)`) and then look for uuid `'null'` or `'undefined'`, and get pretty much exactly the same error but from the api itself.


lib/models/device.coffee
```diff
@@ -192,40 +192,43 @@ getDeviceModel = (deps, opts) ->
...
+			else
+				pine.get
+					resource: 'device'
+					options:
+						expand: 'application'
+						filter:
+							# Handle shorter uuids by asserting
+							# that it is a substring of the device
+							# uuid starting at index zero.
+							$eq: [
```
Page-: I think this can be replaced with `uuid: $startswith: uuidOrId`

lib/models/application.coffee
```diff
@@ -108,27 +108,31 @@ getApplicationModel = (deps, opts) ->
...
-					filter:
-						app_name: nameOrId
-			.tap (applications) ->
-				if isEmpty(applications)
-					throw new errors.ResinApplicationNotFound(nameOrId)
+		Promise.try ->
+			if not nameOrId?
+				throw new errors.ResinApplicationNotFound(nameOrId)
+
+			else if isId(nameOrId)
```
emirotin: NIT: this `else` can be dropped

lib/models/device.coffee
pimterry: Done

lib/models/application.coffee
pimterry: Done

pimterry:
All done, and I've prepended a "// Generated by browserify" comment to the browser build, which makes github hide the diff there too.

PR #278, opened 2017-03-09
Allow pinejs options in models.key.getAll

pimterry:
First test run of adding pine options to an SDK method (because it's required in the UI for https://github.com/resin-io/resin-ui/issues/400), to make sure everybody's happy with the approach and result before I dig in too far.

This is intended to be backward compatible: if you only provide one argument, and it's a function, it'll be assumed that that's the callback.


DOCUMENTATION.md
```diff
@@ -2559,7 +2564,7 @@ The `logs` object yielded by this function emits the following events:
 
 - `line`: when a log line is received.
 - `clear`: when the logs are cleared.
-- `error`: when an error happens.
+- `error`: when an error has happened.
```
emirotin: Huh I thought I've fixed it. In a different place maybe :)

pimterry: This just came in when I rebased and rebuilt - I expect it's just because the documentation wasn't regenerated.

PR #279, opened 2017-03-10
Add build.get, include update_timestamp, and add a lot more pine options

pimterry:
This does most of the hard work for the last supported parts of the UI SDK transition (https://github.com/resin-io/resin-ui/issues/400). It:

* Adds an optional `options` parameter to every method that we need one for in the UI, plus a few more for consistency. 
* Includes `update_timestamp` in the list of fields we select for in `getAllByApplication`
* Adds a new `build.get` method to load individual builds by id.

This depends (via github in package.json) on https://github.com/resin-io/pinejs-client-js/pull/28 - I'll bump that, and then bump the deps in resin-pine and here to pull it in once that's merged.


pimterry:
Released pinejs-client, updated to make this depend on that, and rebased. I'll merge in a sec once at least one of the tests passes.

PR #283, opened 2017-03-16
Add device.getAllByParentDevice and a parent device param to app.create

pimterry:
Connects to https://github.com/resin-io/resin-ui/issues/400. This is a prerequisite for features required in https://github.com/resin-io/resin-ui/pull/535.


jackmurray:
Hi there, just a heads up – we’ll be moving support for free users to the [community forums](https://forums.resin.io/c/troubleshooting) 

this Monday, March 20. The main difference is that your questions and our developers' answers will be public instead of private, building a knowledge base for everyone.

You can take this question to the forums now, I’ll meet you there to answer it. If not, I’ll help you out here, but know that future questions should be directed to the forums. 

P.S. if you’d like to continue accessing private support after the change, we can upgrade you to a paid plan and you won’t see any difference.

PR #288, opened 2017-03-21
a set of OS downloading utils

emirotin:
This implements a set of functions for versioned OS downloads (and also for the OS image size estimate).

More tests are needed and also this is not tested in integration yet (will be done here https://github.com/resin-io/resin-image-manager/issues/36).

But please do the generic review.

P.S. Many tests are skipped for the sake of the testing speed, will un-skip them before merging of course.


CHANGELOG.md
```diff
@@ -3,6 +3,12 @@
 All notable changes to this project will be documented in this file.
 This project adheres to [Semantic Versioning](http://semver.org/).
 
+### Changed
+
+- Add support for the versioned OS downloads, see
+`resin.models.os.{getSupportedVersions, getMaxSatisfyingVersion, formatVersion}`.
+`resin.models.os.{getLastModified, download}` now support the `version` param.
```
pimterry: I think this third line is supposed to be a separate bullet point?

lib/util/device-types.coffee
```diff
@@ -0,0 +1,19 @@
+# copied from the @resin-io/device-types
+# because that has the run-time dependency on coffee-script
+
+find = (array, predicate) ->
+	for el in array
+		return el if predicate(el)
+
+includes = (array, el) -> array and array.indexOf(el) >= 0
```
LucianBuzzo: This can be simplified to `includes = (array, el) -> array and el in array`

lib/util/device-types.coffee
```diff
@@ -0,0 +1,19 @@
...
+
+dtPredicate = (slug) ->
+	return (deviceType) ->
+		deviceType.slug == slug or includes(deviceType.aliases, slug)
```
LucianBuzzo: NIT prefer `is` to `==`

lib/models/os.coffee
```diff
@@ -14,23 +14,223 @@ See the License for the specific language governing permissions and
...
+	getDownloadSize = imgMakerHelper.buildApiRequester
+		url: '/size_estimate'
+		withVersion: true
+		postProcess: ({ body }) -> body.size
+
+	getOsVersions = imgMakerHelper.buildApiRequester
+		buildUrl: ({ deviceType }) ->
+			"/image/#{deviceType}/versions"
+		postProcess: ({ body }) ->
+			{ versions, latest } = body
```
pimterry: Nit: you could extract these directly above: `({ body: { versions, latest } }) ->`

lib/models/os.coffee
```diff
@@ -14,23 +14,223 @@ See the License for the specific language governing permissions and
...
+	# });
+	###
+	exports.getDownloadSize = (deviceType, callback) ->
+		callback = findCallback(arguments)
+
+		return isValidDeviceType(deviceType)
+			.then (isValid) ->
+				if not isValid
+					throw new errors.ResinRequestError('No such device type')
+			.then ->
```
pimterry: Don't need a separate `.then` here. Also true in other functions below.

lib/models/os.coffee
```diff
@@ -14,23 +14,223 @@ See the License for the specific language governing permissions and
...
+	#
+	# @example
+	# console.log(resin.models.os.formatVersion('1.2.3', osVersions)) // => 'v1.2.3 (recommended)'
+	###
+	exports.formatVersion = (v, osVersions) ->
+		v = normalizeVersion(v)
+		result = "v#{v}"
+		if v is osVersions.recommended
+			result += ' (recommended)'
+		return result
```
pimterry: This feels like a low-level utility method - it feels very strange to me that this is in the SDK. Could we move this downstream to where we use it instead?

lib/models/os.coffee
```diff
@@ -14,23 +14,223 @@ See the License for the specific language governing permissions and
...
+	# 	console.log('The OS download size for raspberry-pi', size);
+	# });
+	###
+	exports.getDownloadSize = (deviceType, callback) ->
+		callback = findCallback(arguments)
+
+		return isValidDeviceType(deviceType)
+			.then (isValid) ->
+				if not isValid
+					throw new errors.ResinRequestError('No such device type')
```
pimterry: This (and the other similar cases below) shouldn't be throwing `ResinRequestError`. That's for HTTP errors, it comes with status code and body properties, this isn't anything like that. We've got a `ResinInvalidDeviceType` error already - we should use that instead.

lib/models/os.coffee
```diff
@@ -39,24 +239,33 @@ getOsModel = (deps, opts) ->
...
+			.then ->
+				return normalizeVersion(version)
+			.then (version) ->
+				imgMakerHelper.request
+					method: 'HEAD'
+					url: "/image/#{deviceType}/?version=#{version}"
+			.catch notFoundResponse, ->
+				throw new errors.ResinRequestError('No such version for the device type')
```
pimterry: Shouldn't really be a `ResinRequestError` either. I'd just make this `Error` for now, we could add more specific cases later, but this is forward compatible, and I don't think it matters much.

lib/util/device-types.coffee
```diff
@@ -0,0 +1,19 @@
+# copied from the @resin-io/device-types
+# because that has the run-time dependency on coffee-script
+
+find = (array, predicate) ->
+	for el in array
+		return el if predicate(el)
```
pimterry: Why don't we use lodash's `find` here?

lib/util/device-types.coffee
pimterry: Or replaced with `_.includes`.

lib/util/img-maker.coffee
```diff
@@ -0,0 +1,74 @@
...
+
+IMG_MAKER_API_VERSION = '1'
+IMG_MAKER_API_PREFIX = "/api/v#{IMG_MAKER_API_VERSION}"
+
+DEFAULT_RESULTS_CACHING_INTERVAL = 10 * 60 * 1000 # 10 minutes
+
+getImgMakerHelper = (imageMakerUrl, request) ->
+	exports = {}
+
+	exports.buildOptions = buildOptions = (options) ->
```
pimterry: This doesn't need to be exported.

tests/integration.spec.coffee
```diff
@@ -125,7 +125,7 @@ describe 'SDK Integration Tests', ->
 				mockResin = getSdk(opts)
 				m.chai.expect(mockResin).to.include.keys(validKeys)
 
-	describe 'given a not logged in user', ->
+	describe.skip 'given a not logged in user', ->
```
pimterry: It's probably easier in general to just put `.only` on each of the sets of tests you do want to run, rather than excluding every other batch.

CHANGELOG.md
emirotin: Didn't mean that but can separate of course

lib/models/os.coffee
emirotin: Yeah I know but dislike such deep desturcturings.

lib/models/os.coffee
emirotin: My bad, copy-pasted it from a different place.

lib/models/os.coffee
emirotin: Copied it from the UI where it's used but plan to make the same usage in the CLI. Think it makes more sense to duplicate it there?

lib/models/os.coffee
emirotin: Used to be this way, but can change of course

lib/util/device-types.coffee
emirotin: I'm just a bit cautious of bringing extra deps when the necessity is such trivial

tests/integration.spec.coffee
emirotin: Yeah forgot about it :)

lib/models/os.coffee
pimterry: Ah, ok, in that case this I'm not that worried. I'm cautious about using the wrong exceptions just because later fixes to those are breaking changes, but if they're already wrong then there's no need to block this on fixing them.

lib/models/os.coffee
pimterry: I'd rather not. This is a user-facing set of tools to write code that interacts with Resin, we shouldn't turn it into a place for general purpose internal utilities. We could build a separate library for internal utility functions? For now I'd move it downstream into the CLI instead of here, and we can very easily iterate towards nicer solutions later on.

lib/models/os.coffee
pimterry: Fair enough, I'm not too bothered either way.

lib/models/os.coffee
emirotin: OK

CHANGELOG.md
```diff
@@ -3,6 +3,12 @@
 All notable changes to this project will be documented in this file.
 This project adheres to [Semantic Versioning](http://semver.org/).
 
+### Changed
+
+- Add support for the versioned OS downloads, see
+`resin.models.os.{getSupportedVersions, getMaxSatisfyingVersion, formatVersion}`.
```
pimterry: `formatVersion` should go from here, now that it's out of the code.

Docs need updating too.

emirotin:
Travis builds has failed consistently with the `method is not defined` error from resin-request in the application.getAll() tests for custom pine options. Locally both tests pass so this is a glitch, but surprisingly consistent.
I suppose it may be related to the fact I forgot to update the yarn lock

pimterry:
That's an actual broken install - it happens if you don't have [this pinejs change](https://github.com/resin-io/pinejs-client-js/pull/28) included, which is required to support the options merging. It's in PineJS 3.1.0, which should get pulled in automatically. Which version is being installed?

pimterry:
Ah, wait, I see it's now passing after updating the yarn.lock. All good then :+1:

emirotin:
In this case we should bump the resin-pine version and the dependency here to indicate we don't work with the earlier versions.

emirotin:
Created a separate issue but will merge this (once tested in the CLI) without waiting for it

emirotin:
> Request error: "uuid" must be unique..

I've seen this error before. This is very strange. I think this is another implication of some race conditions.

pimterry:
@emirotin I think it's still part of #252, and I still have no idea why that happens. I never see it locally, but it happens remarkably frequently in CI. Doesn't seem to be caused by test runs in parallel, it's race conditions inside the test suite itself. Ideas would be very welcome!

emirotin:
memoizee's maxAge is holding process from exiting: https://github.com/medikoo/memoizee/issues/25

Have to resolve before merging

Issue #282, opened 2017-03-12
Add device GPS locations 


craig-mulligan:
Allow same functionality as offered from the dashboard.

alexandrosm:
@lucianbuzzo -- this is the kind of thing i was talking about when
discussing the sdk :)

-- 

*Alexandros Marinos*

Founder & CEO, Resin.io


+1 206-637-5498

@alexandrosm

LucianBuzzo:
@alexandrosm Yep this is a perfect example! 
The TLDR -  ideally the SDK should be able to reflect changes in the resin SBVR model without requiring new code to be written.

pimterry:
This is a planned part of https://github.com/resin-io/resin-ui/issues/400#issuecomment-285693667, and I'm expecting to get this in early this week.

(SDK support for the gps locations I mean, not auto-generating the SDK, although I totally agree that that'd be a good thing :+1:)

pimterry:
Fixed in 6.0.0.

PR #291, opened 2017-03-22
Add device.setCustomLocation and device.unsetCustomLocation

pimterry:
This solves #282, and is yet another part of https://github.com/resin-io/resin-ui/issues/400.


lib/models/device.coffee
```diff
@@ -642,6 +642,71 @@ getDeviceModel = (deps, opts) ->
...
+	# resin.models.device.setCustomLocation('7cf02a6', { latitude: 123, longitude: 456 }, function(error) {
+	# 	if (error) throw error;
+	# });
+	###
+	exports.setCustomLocation = (uuidOrId, location, callback) ->
+		exports.get(uuidOrId).then (device) ->
+			return pine.patch
+				resource: 'device'
+				body:
+					custom_latitude: String(location.latitude),
```
emirotin: Why do we send (and store?) them as strings?

emirotin: Oh, is it for empty string to denote no value? Can nulls be used instead?

pimterry: I've mainly done it this way simply because that's what the existing UI code currently does and expects. Not sure if there's a reason we can't use nulls here, that sounds reasonable to me. @LucianBuzzo?

PR #294, opened 2017-04-03
Handle non-semver OS versions

pimterry:
This is the core of the fix for https://github.com/resin-io/resin-cli/issues/470. I'm doing integration with resin-image-manager and resin-cli as we speak, but this needs merging & releasing first.

I'm intentionally not trying to build a tidy fully featured semver-replacement here, and instead I'm just handling the likely short-term cases so we can get the CLI working again for now. We're going to have an architecture call later, and all of this will be replaced one way or the other.


lib/models/os.coffee
```diff
@@ -49,10 +52,13 @@ getOsModel = (deps, opts) ->
 			"/image/#{deviceType}/versions"
 		postProcess: ({ body }) ->
 			{ versions, latest } = body
-			versions.sort(semver.rcompare)
-			recommended = reject(versions, semver.prerelease)?[0] || null
+			[ validVersions, invalidVersions ] = partition(versions, semver.valid)
+
+			validVersions.sort(semver.rcompare)
+			recommended = reject(validVersions, semver.prerelease)?[0] || null
```
pimterry: @LucianBuzzo Looking at this makes me realise that whichever way we handle the version numbers from now on, we should be commonizing the OS version recommendation logic.

Issue #295, opened 2017-04-03
SDK doesn't encode + symbol when downloading images


pimterry:
The SDK currently requests URLs like `http://img.resinstaging.io/api/v1/image/raspberrypi3?version=2.0.0+rev1`, where the `+` is interpreted as a URL encoded space. This should encoded instead, as `%2B`.

PR #296, opened 2017-04-03
Encode plus symbols in version numbers when downloading OS images

pimterry:
Connects to #295.

Tests here won't actually pass, since they make queries to production, and the version they look for hasn't been published yet. I've run these same tests against staging though, where `2.0.0+rev1` does exist, and they pass happily there for me.


tests/integration.spec.coffee
```diff
@@ -1740,6 +1740,14 @@ describe 'SDK Integration Tests', ->
 						promise = resin.models.os.getLastModified('raspberrypi')
 						m.chai.expect(promise).to.eventually.be.an.instanceof(Date)
 
+					it 'should be able to query for a specific version', ->
+						promise = resin.models.os.getLastModified('raspberrypi', '1.24.0')
+						m.chai.expect(promise).to.eventually.be.an.instanceof(Date)
+
+					it 'should be able to query for a version containing a plus', ->
+						promise = resin.models.os.getLastModified('raspberrypi', '2.0.0+rev1')
```
petrosagg: These will be eventually removed, I'm not sure if we should reference them from the tests. Especially 1.24

pimterry: True, but we don't have any integration testing up here that isn't against the real service and releases, so I'm not sure how else to test this. Any ideas? Doesn't feel like a huge disaster if these start failing in future when we remove the images anyway - it'll be easy and clear how to fix them, and we'll have more time to work with.

petrosagg: ok. I'm approving to get this moving

lib/models/os.coffee
```diff
@@ -74,6 +74,9 @@ getOsModel = (deps, opts) ->
 			throw new Error("Invalid semver version: #{v}")
 		return vNormalized
 
+	deviceImageUrl = (deviceType, version) ->
```
emirotin: I think `size_estimate` requests need the same fix

https://github.com/resin-io/resin-sdk/blob/e79d43ed47aa423403ff5b3f55fe47bc19a8567e/lib/util/img-maker.coffee#L27
usage example: https://github.com/resin-io/resin-sdk/blob/e79d43ed47aa423403ff5b3f55fe47bc19a8567e/lib/models/os.coffee#L45

pimterry: Fixed here: https://github.com/resin-io/resin-sdk/pull/297

PR #297, opened 2017-04-05
Encode plus symbols in version numbers when estimating OS image size

pimterry:
As @emirotin pointed out [here](https://github.com/resin-io/resin-sdk/pull/296/files#r109640039), new ResinOS format also breaks image maker requests for `size_estimate`. This does break anything other than the SDK I think (doesn't seem like we use it elsewhere yet), but we should be using this code in the UI (which currently doesn't show a size for this image) instead of duplicating the logic, and really this needs fixing anyway.

This PR fixes that by ensuring we always safely encode version numbers (and device names, just in case). It was a bit fiddly to work out exactly how the underlying URLs get built here, so I've also simplified some of the memoization code along the way, and added some tests for this as well.


emirotin:
We do request size estimate in the UI and I think it's currently not using the SDK

lib/util/img-maker.coffee
```diff
@@ -60,12 +43,10 @@ getImgMakerHelper = (imageMakerUrl, request) ->
 		else if maxAge is null
 			maxAge = undefined
 
-		memoizedFn = promiseMemoize(callHelper, { resolve: normalizer, maxAge })
+		memoizedFunction = promiseMemoize(callHelper, { maxAge })
```
emirotin: So this will always stringify args for the key and will use `latest` as the version for the requests (like `/versions`) that don't have this param. Sounds OK, just verifying it.

tests/integration.spec.coffee
```diff
@@ -1722,6 +1722,16 @@ describe 'SDK Integration Tests', ->
 						promise = resin.models.os.getDownloadSize('raspberrypi')
 						m.chai.expect(promise).to.eventually.satisfy((n) -> typeof n is 'number')
 
+				describe 'given a specific OS version', ->
+
+					it 'should get a result for ResinOS v1', ->
+						promise = resin.models.os.getDownloadSize('raspberry-pi', '1.24.0')
```
emirotin: BTW any chance to test that memoization works?
One suggestion is to call the `getDownloadSize` for two different versions and ensure the results are different. That as the very least will ensure that the version does actually influence the cache key.

lib/util/img-maker.coffee
pimterry: Yes, exactly.

tests/integration.spec.coffee
pimterry: Good idea! I've now added a few tests to cover memoization specifically - looks all good.

tests/util.spec.coffee
```diff
@@ -117,3 +121,35 @@ describe 'Pine option merging', ->
...
 			-> mergePineOptions(unknownKey: 'value', {})
 		).not.to.throw()
+
+describe 'ImgMakerHelper', ->
+
+	ROOT_URL = 'https://img.resin.io'
+
+	beforeEach =>
+		@requestStub = send: m.sinon.stub().returns(new Promise(
+			-> # never return, don't worry about responses
```
pimterry: This is _very_ annoying, I've hit it before too, and it doesn't seem to come up in local linting... I'll going to find a proper way to fix it now.

pimterry: Success! I think... Seems hound's `no_empty_functions` default is different to our local one for some reason. No idea why, but explicitly turning it off seems to have worked. Also fixed incorrect config key in `.hound.yml`. It now matches the docs, although I'm not wholly sure that's a major problem. Sorted anyway, and I've tested that it does work for other errors. Sorry for the notification noise!

PR #300, opened 2017-04-06
Update the readme, since imageMakerUrl is now also used in browsers

pimterry:
It now always defaults to `img.resin.io`. I'm not sure why it was pulling from the settings before, given that the api url doesn't. Is there a good reason for that? I've optimistically removed it anyway, so everything's consistent.


emirotin:
I believe there's no good reason to pull it from the settings, but it was doing that because the settings client module was putting it there by default.

PR #301, opened 2017-04-11
Automatically publish tagged master commits to npm

pimterry:
This is the first step to adding versionbot to the SDK. With this, any time we tag a change, travis will publish it to npm. That should mean once we add versionbot, versionbot merges will automatically publish the new version to npm. Neat! This uses the resin.io credentials, but with the standard travis encryption setup to keep them hidden.

This does mean we don't have to sit and wait for the publish tests to run for us locally, but I suspect it also means we'll have to deal with instability on deploy occasionally from #252. That needs solving independently anyway (although I still don't have a useful answer).

Some reformatting here because I set this up with the travis cli tool, and it doesn't like our yaml formatting, I'm inclined to stick with its decisions there.

This might be of interest to @hedss, since this is our solution to https://github.com/resin-io-modules/resin-procbots/issues/30.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,14 +1,22 @@
 language: node_js
 node_js:
-  - "6"
-  - "4"
+- '6'
+- '4'
 before_install:
-  - npm -g install npm
+- npm -g install npm
 notifications:
   email: false
   webhooks:
     urls:
-      - https://webhooks.gitter.im/e/8b150eaf525c280ec2ac
+    - https://webhooks.gitter.im/e/8b150eaf525c280ec2ac
     on_success: change
     on_failure: always
     on_start: never
+deploy:
+  provider: npm
+  email: 'accounts@resin.io'
+  api_key:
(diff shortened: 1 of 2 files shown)


lurch:
> I suspect it also means we'll have to deal with instability on deploy occasionally

Yup, one of the Travis tests on this PR has failed already ;-)

Pinging @jviotti too, as he recently added automated publishing of Etcher snapshot builds to S3 from Travis and Appveyor. (but in that case he just used secure env-vars on the CI services, rather than adding encrypted variables to the YAML files)

pimterry:
@lurch Eugh, good point, rerun.

The actual deploy later on should be a fair bit more reliable than normal builds, because the deploy step will be a single test run, rather than four. No need to deploy separately for Node 4 or 6 or appveyor :-). And if it's the case that's the normal test runs are somehow interfering, it should be _totally_ reliable, which would be nice. Fingers crossed.

Issue #302, opened 2017-04-12
Take revX metadata into account when sorting OS versions


pimterry:
https://www.flowdock.com/app/rulemotion/resin-devops/threads/yCepfZVbu664FOBN4movTFi7iLl

pimterry:
Fixed in #303, closing.

PR #303, opened 2017-04-12
Sort OS versions taking their rev into account if present

pimterry:
The test at the very bottom is the best place to look if you're just checking you're happy with the resulting order we'll be seeing.

Recommended logic doesn't change, that's still just "first in the list that's not prerelease".


lib/util/index.coffee
```diff
@@ -39,6 +40,26 @@ exports.treatAsMissingDevice = (uuidOrId) ->
...
 
+exports.osVersionCompare = (versionA, versionB) ->
+	semverResult = semver.rcompare(versionA, versionB)
+	if semverResult != 0
+		return semverResult
+
+	revA = getRev(versionA)
+	revB = getRev(versionB)
+
+	return revB - revA
```
emirotin: NIT: this coerces `null` to `0` which works but is slightly ugly, I'd just return `0` from `getRev`

lib/util/index.coffee
```diff
@@ -39,6 +40,26 @@ exports.treatAsMissingDevice = (uuidOrId) ->
 		replacementErr.stack = err.stack
 		throw replacementErr
 
+exports.osVersionCompare = (versionA, versionB) ->
```
emirotin: Also NIT: this is R-compare so I would rename the method to make it obvious the comparison is inversed here.

lurch:
Hah, looks a bit like the code @jviotti is adding to Etcher in a recent PR ;-)
https://github.com/resin-io/etcher/pull/1183/files#diff-761eb090648cc858c72e655f2a6171bcR206

lurch:
P.S. @pimterry Should this be connecting to #302 ?

lib/util/index.coffee
pimterry: Sure, done.

pimterry:
@lurch yes, it should, thanks! Sort-of similar, although that's how ranges match to prerelease tags, and this ordering of build tags. It might be nice to share all this logic eventually though. Maybe in the medium term we should have a resin-semver module, which extends semver to take prerelease/build metadata into account in the way we'd like? Could be useful, especially if more of this emerges in future.

lib/util/index.coffee
pimterry: Sure, done.

Issue #286, opened 2017-03-21
Unmet peer dependency?


emirotin:
Looks like npm dedupe plays against us.

If I go to the SDK repo and run `npm install` all goes fine.
If I install `resin-sdk-preconfigured` as a dependency to some other project it works fine, too.

But when clone the resin-cli and did npm install there I've seen:
```
npm ERR! peer dep missing: resin-token@^3.0.0, required by resin-pine@5.0.1
npm ERR! peer dep missing: resin-token@^3.0.0, required by resin-request@8.3.0
```

And:
```
✔ npm ls resin-request
resin-cli@5.5.0 /Users/eugene/Work/resin-cli
├─┬ resin-sdk-preconfigured@0.1.0
│ └─┬ resin-sdk@6.0.0-beta5
│   ├── resin-request@8.3.0
│   └── UNMET PEER DEPENDENCY resin-token@3.0.0
└─┬ resin-sync@7.0.2
  └─┬ resin-sdk@5.4.0
    └── resin-request@4.1.2
```

I think here's the issue:
```
✔ ls node_modules | grep resin
resin-cli-auth
resin-cli-errors
resin-cli-form
resin-cli-visuals
resin-config-json
resin-device-config
resin-device-init
resin-device-logs
resin-device-operations
resin-device-status
resin-discoverable-services
resin-errors
resin-image-fs
resin-image-manager
resin-image-write
resin-pine
resin-register-device
resin-request
resin-sdk
resin-sdk-preconfigured
resin-settings-client
resin-settings-storage
resin-sync
resin-token

✔ npm ls resin-token
resin-cli@5.5.0 /Users/eugene/Work/resin-cli
└─┬ resin-cli-auth@1.1.2
  └── resin-token@2.4.3
```

So resin-cli-auth depends on the older version fo token which is lifted to the top-level.
At the same time `resin-request` was deduped and lifted to the top level where it now has the peer dependency... of the wrong version.

Should we remove the peer deps declarations in spite of this?
The docs (of the resin-request and resin-pine) already say the token should be injected.
---

Also `pinejs-client` is no longer a direct dependency and can be removed.

pimterry:
Looks like this is just a problem with how npm displays this, not with the packages that get installed or where they're put. On a clean install, I end up with a node_modules folder structure including:

```
- resin-sdk @ 6.0.0-beta5
  - node_modules
     - resin-token @ 3.0.0

- resin-sdk-preconfigured @ 0.1.0
[...]

PR #304, opened 2017-04-13
Drop unnecessary direct pinejs-client-dep dependency

pimterry:
Connects to https://github.com/resin-io/resin-ui/issues/643


pimterry:
I think you spotted it in #286, but then we never actually got around to removing it (according to `git log -p -S pinejs-client -- package.json`)

PR #305, opened 2017-04-15
Add a `device.generateDeviceKey` that can generate a device specific key for use by a device

Page-:
Depends on https://github.com/resin-io/resin-api/pull/183


lib/models/device.coffee
```diff
@@ -1268,6 +1268,36 @@ getDeviceModel = (deps, opts) ->
...
+	# @example
+	# resin.models.device.generateDeviceKey('7cf02a6');
+	#
+	# @example
+	# resin.models.device.generateDeviceKey(123);
+	#
+	# @example
+	# resin.models.device.generateDeviceKey('7cf02a6', function(error) {
+	# 	if (error) throw error;
+	# });
```
pimterry: Nit: these examples all generate a key and throw it away. It'd be nice to add a `.then` to make it clearer how you get the result.

lib/models/device.coffee
```diff
@@ -1268,6 +1268,36 @@ getDeviceModel = (deps, opts) ->
...
+	###
+	exports.generateDeviceKey = (uuidOrId, callback) ->
+		getId(uuidOrId).then (deviceId) ->
+			return request.send
+				method: 'POST'
+				url: "/api-key/device/#{deviceId}/device-key"
+				baseUrl: apiUrl
```
pimterry: What happens if you pass an id, and the device doesn't exist?

`getId` won't throw, it'll just pass it through. Assuming that endpoint comes back with a 404, you'll just get a `ResinRequestError`. It'd be better to do `.catch(notFoundResponse, treatAsMissingDevice(uuidOrId))`.

Page-: Done - although it's not a 404 in this case

lib/models/device.coffee
Page-: Done - changed to match examples that have a result (I think I copied from a set that didn't)

lib/models/device.coffee
```diff
@@ -1268,6 +1268,49 @@ getDeviceModel = (deps, opts) ->
...
+	# @public
+	# @function
+	# @memberof resin.models.device
+	#
+	# @param {String|Number} uuidOrId - device uuid (string) or id (number)
+	# @returns {Promise}
+	#
+	# @example
+	# resin.models.device.generateDeviceKey('7cf02a6').then(function(deviceApiKey) {
+	# 	console.log(registrationInfo);
```
pimterry: Variable name is wrong here

PR #308, opened 2017-04-26
update browserify

emirotin:
This is a trivial change that slightly reduces the bundle size
Browserify 14 has dropped support for IE <= 10 which we don't care about anyway, by adopting the faster typed array.

I'm going to merge it as soon as the tests pass.


emirotin:
Ugh, `ArrayBuffer.isView` is not supported in Phantom (but is supported in our target browsers https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/ArrayBuffer/isView)

Issue #311, opened 2017-04-27
Add a method for downloading an applications config.json file


LucianBuzzo:

pimterry:
@LucianBuzzo this can be closed, right?

LucianBuzzo:
@pimterry  It can indeed :smile:

PR #313, opened 2017-04-27
Add a method for downloading an applications config.json file

LucianBuzzo:
Connects to #311

change-type: minor


emirotin:
@pimterry Travis did two builds here

emirotin:
`m.chai.expect(...).to.eventually.become.a is not a function`

lib/models/os.coffee
```diff
@@ -304,6 +304,42 @@ getOsModel = (deps, opts) ->
...
 			.asCallback(callback)
 
+	###*
+	# @summary Get an applications config.json
+	# @name getConfig
+	# @public
+	# @function
+	# @memberof resin.models.os
+	#
+	# @param {Number} Id - application id (number)
```
pimterry: This should probably be `nameOrId`, and disambiguate the two, like most of the application methods. Partly for consistency, and also because that's what you've assumed it is in the examples here :smile:

lib/models/os.coffee
```diff
@@ -304,6 +304,42 @@ getOsModel = (deps, opts) ->
...
+	#
+	# resin.models.os.getConfig(123).then(function(config) {
+	# 	fs.writeFile('foo/bar/config.json', JSON.stringify(config));
+	# });
+	#
+	#
+	# resin.models.os.download('MyApp', function(error, config) {
+	# 	if (error) throw error;
+	# 	fs.writeFile('foo/bar/config.json', JSON.stringify(config));
+	# });
```
pimterry: This example is calling `os.download`

lib/models/os.coffee
```diff
@@ -304,6 +304,42 @@ getOsModel = (deps, opts) ->
...
+	###
+	exports.getConfig = (id, callback) ->
+		callback = findCallback(arguments)
+
+		request.send
+			method: 'GET'
+			url: "/download-config?appId=#{id}"
+			baseUrl: apiUrl
+		.get('body')
+		.asCallback(callback)
```
pimterry: It's nicer to handle 404s by throwing a `ResinApplicationNotFound` error, rather than a ResinRequest error that people will have to introspect, so you can consistency handle that elsewhere. The application model has a bunch of [examples](https://github.com/resin-io/resin-sdk/blob/master/lib/models/application.coffee#L324).

tests/integration.spec.coffee
```diff
@@ -1829,6 +1829,26 @@ describe 'SDK Integration Tests', ->
...
+					m.chai.expect(promise).to.eventually.have.property('userId')
+					m.chai.expect(promise).to.eventually.have.property('username')
+					m.chai.expect(promise).to.eventually.have.property('deviceType')
+					m.chai.expect(promise).to.eventually.have.property('files')
+					m.chai.expect(promise).to.eventually.have.property('apiEndpoint')
+					m.chai.expect(promise).to.eventually.have.property('registryEndpoint')
+					m.chai.expect(promise).to.eventually.have.property('vpnEndpoint')
+					m.chai.expect(promise).to.eventually.have.property('pubnubSubscribeKey')
+					m.chai.expect(promise).to.eventually.have.property('pubnubPublishKey')
+					m.chai.expect(promise).to.eventually.have.property('listenPort')
```
pimterry: We should add another test to check you get a sensible error if the application id doesn't exist.

lib/models/os.coffee
```diff
@@ -304,6 +304,42 @@ getOsModel = (deps, opts) ->
...
+	# });
+	#
+	#
+	# resin.models.os.download('MyApp', function(error, config) {
+	# 	if (error) throw error;
+	# 	fs.writeFile('foo/bar/config.json', JSON.stringify(config));
+	# });
+	###
+	exports.getConfig = (id, callback) ->
+		callback = findCallback(arguments)
```
pimterry: You don't need to do this unless you have optional arguments. With only required arguments, you know exactly which argument is the callback already, so you can just use it directly here.

tests/integration.spec.coffee
```diff
@@ -1829,6 +1829,26 @@ describe 'SDK Integration Tests', ->
 						promise = resin.models.os.download('foo-bar-baz')
 						m.chai.expect(promise).to.be.rejectedWith('No such device type')
 
+			describe 'resin.models.os.getConfig()', ->
+				before ->
```
pimterry: This should be `beforeEach`. We delete everything created between every test. Just a single `before` will work for one test, but as soon as you try to add another in here the application will mysteriously not exist 😮

pimterry:
Also, I'm assuming you're pushing failing builds because you can't easily test these locally. It is a bit painful to run the whole suite (takes about 10m+, depending on latency). Most of the time though you only want to run the set of tests closely related to what you're changing, and you should be able to do so very quickly and easily by adding `.only` to the individual tests you care about (i.e. changing to `describe.only` and `it.only`). That should take it down to about 20s for a small set.

You'll also need to set yourself up with a test user, and add their settings to a `.env` file in your local checkout, if you haven't already. The README has docs for this: https://github.com/resin-io/resin-sdk#tests

pimterry:
Oh, and @emirotin it ran two builds because I'd changed the Travis settings to enable branch builds, but #312 hadn't yet been merged. It now has. From now, any PRs that are based on top of current master should only run a single build for the PR, not the branch.

LucianBuzzo:
@pimterry Should be good to go now!

Issue #319, opened 2017-05-08
getConfig method doesn't take into account configuration options


LucianBuzzo:

pimterry:
@LucianBuzzo this can be closed, right?

PR #320, opened 2017-05-08
Allow config options when using the models.os.getConfig() method.

LucianBuzzo:
Connects to #319

change-type: minor


lib/models/os.coffee
```diff
@@ -330,13 +342,25 @@ getOsModel = (deps, opts) ->
 	# 	fs.writeFile('foo/bar/config.json', JSON.stringify(config));
 	# });
 	###
-	exports.getConfig = (nameOrId, callback) ->
+	exports.getConfig = (nameOrId, options = {}, callback) ->
```
pimterry: Now that you have an optional argument, you do need to use `findCallback` to work out which argument was the callback. Currently `getConfig(123, function (config) { ... ` will try to use that callback function as the options, and blow up.

lib/models/os.coffee
```diff
@@ -330,13 +342,25 @@ getOsModel = (deps, opts) ->
...
+		defaultOpts =
+			appUpdatePollInterval: 1
+			network: 'ethernet'
+			wifiKey: null
+			wifiSsid: null
+			ip: null
+			gateway: null
+			netmask: '255.255.255.0'
```
pimterry: This default is already configured in `resin-device-config`, so I don't think we should duplicate it here.

lib/models/os.coffee
```diff
@@ -330,13 +343,25 @@ getOsModel = (deps, opts) ->
 	# 	fs.writeFile('foo/bar/config.json', JSON.stringify(config));
 	# });
 	###
-	exports.getConfig = (nameOrId, callback) ->
+	exports.getConfig = (nameOrId, options = {}, callback) ->
+		defaultOpts =
+			appUpdatePollInterval: 1
```
pimterry: I know this defaults to 1 in the UI, but are we sure it's in minutes at the API level? `resin-device-config` [defaults](https://github.com/resin-io-modules/resin-device-config/blob/master/lib/config.coffee#L77) to 60000... Feels like somebody must be wrong there.

LucianBuzzo: good catch, its `ms` https://github.com/resin-io/resin-api/blob/master/src/routes/applications.coffee#L80

LucianBuzzo:
@pimterry This PR is blocked until the next API release that allows us to POST the the `/download-config` endpoint.

tests/integration.spec.coffee
```diff
@@ -1880,6 +1880,26 @@ describe 'SDK Integration Tests', ->
...
+						version: 'v1+foo'
+					promise = resin.models.os.getConfig(@application.id, configOptions)
+					# NOTE: the interval is converted to ms in the config object
+					m.chai.expect(promise).to.eventually.have.property('appUpdatePollInterval').that.equals(configOptions.appUpdatePollInterval * 60 * 1000)
+					m.chai.expect(promise).to.eventually.have.property('wifiKey').that.equals(configOptions.wifiKey)
+					m.chai.expect(promise).to.eventually.have.property('wifiSsid').that.equals(configOptions.wifiSsid)
+					m.chai.expect(promise).to.eventually.have.property('version').that.equals(configOptions.version)
+					m.chai.expect(promise).to.eventually.have.property('files')
+						.that.has.property('network/network.config')
+						.that.includes("#{configOptions.ip}/#{configOptions.netmask}/#{configOptions.gateway}")
```
pimterry: Each of these asserts returns a promise, but you only actually wait on the last promise here. If any except the last fails, the test will still pass (though you should get an 'unhandled rejection' error in the console).

These should be wrapped with a Promise.all.

lib/models/os.coffee
```diff
@@ -330,13 +344,22 @@ getOsModel = (deps, opts) ->
 	# 	fs.writeFile('foo/bar/config.json', JSON.stringify(config));
 	# });
 	###
-	exports.getConfig = (nameOrId, callback) ->
+	exports.getConfig = (nameOrId, options = {}, callback) ->
+		callback = findCallback(arguments)
+
+		defaultOpts =
+			appUpdatePollInterval: 60000
```
pimterry: Now that this is the same as resin-device-config's default, we should probably remove this too, and let that handle the defaults there itself.

DOCUMENTATION.md
```diff
@@ -2369,15 +2369,24 @@ resin.models.os.download('raspberry-pi', function(error, stream) {
...
 
-| Param | Type | Description |
-| --- | --- | --- |
-| nameOrId | <code>String</code> \| <code>Number</code> | application name (string) or id (number) |
+| Param | Type | Default | Description |
+| --- | --- | --- | --- |
+| nameOrId | <code>String</code> \| <code>Number</code> |  | application name (string) or id (number) |
+| [options] | <code>Object</code> | <code>{}</code> | OS configuration options to use |
+| [options.network] | <code>String</code> | <code>&#x27;ethernet&#x27;</code> | The network type that |
+| [options.appUpdatePollInterval] | <code>Number</code> |  | How often the OS checks for updates, in minutes. the device will use, one of 'ethernet' or 'wifi' |
```
zvin: @LucianBuzzo You mixed up the network and appUpdatePollInterval lines

lib/models/os.coffee
```diff
@@ -313,6 +314,19 @@ getOsModel = (deps, opts) ->
 	# @memberof resin.models.os
 	#
 	# @param {String|Number} nameOrId - application name (string) or id (number)
+	# @param {Object} [options={}] - OS configuration options to use
+	# @param {String} [options.network='ethernet'] - The network type that
+	# @param {Number} [options.appUpdatePollInterval] - How often the OS checks
+	# for updates, in minutes.
+	# the device will use, one of 'ethernet' or 'wifi'
```
zvin: lines mixed up

PR #324, opened 2017-05-11
Manually handhold travis, so it runs the right tests/deploys at the right times

pimterry:
The comments at the top explain the result I'm looking for I think.

This should make deploys quite a bit quicker too, since we don't run the tests before `npm publish` (which also runs the tests), and we don't have to bother with different node versions.

While this skips the tests for non-tagged commits that aren't on PRs, it doesn't skip the build entirely, since Travis doesn't allow that. This means every skipped commit will get a little green tick in the github UI, even though it hasn't actually run the tests. I could make that a failure instead, but I think this is fine, since any interesting commits (on PRs, new versions on master) will see correct markers.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,9 +1,25 @@
+# We want to avoid unnecessary test runs, because they load the API, so:
+# - we run tests for all changes in PRs
+# - we run deploys for all version-tagged commits
+# - we don't want to do anything for any other cases
+# Travis has a bunch of filtering that's supposed to cover this sort of this, but
+# it doesn't quite work.
+
 language: node_js
-node_js:
-- '6'
-- '4'
+matrix:
+  include:
+    - node_js:
+      - '6'
+      env:
+      - 'CAN_DEPLOY=true'
+    - node_js:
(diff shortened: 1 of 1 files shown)


pimterry:
Intriguingly half an hour ago Travis announced Build Stages, which relates closely to this: https://blog.travis-ci.com/2017-05-11-introducing-build-stages.

That would allow us to neatly have a multi-version 'test' stage, and a single-version 'deploy' stage, and neatly separate them in the UI too. Looks like it doesn't let you conditionally include stages though, and I don't think it's worth more hassle here right now, but it's an interesting idea for the future.

pimterry:
The appveyor issue is a separate one: an npm module we depend on just published a totally broken version, see https://github.com/crypto-browserify/createHmac/issues/17. Looks like Travis passes because Yarn locking saves the day 😄 .

(This passing isn't actually a good thing overall: we should really fail if the SDK is now uninstallable, which it temporarily is. I'm edging back towards Page's original thoughts against Yarn in modules like this...)

PR #325, opened 2017-05-16
Always choose the production image as the recommended version

Page-:
Change-type: patch


lib/models/os.coffee
```diff
@@ -54,8 +54,10 @@ getOsModel = (deps, opts) ->
 			{ versions, latest } = body
 			[ validVersions, invalidVersions ] = partition(versions, semver.valid)
 
-			validVersions.sort(osVersionRCompare)
-			recommended = reject(validVersions, semver.prerelease)?[0] || null
+			potentialRecommendedVersions = reject versions, (version) ->
+				semver.prerelease(version) or _.includes(version, '.dev')
+			potentialRecommendedVersions.sort(osVersionRCompare)
+			recommended = potentialRecommendedVersions?[0] || null
```
emirotin: Should we recommend the newest .dev version if all the versions are dev?

emirotin:
Windows tests failed with buildUrl(...).then is not a function. Any idea why?

Page-:
I've got no idea on that error, I can't see why this PR would affect it (and definitely not windows-only)

lib/models/os.coffee
Page-: Maybe, but that shouldn't ever be the case - how would you suggest doing it?

Page-:
It looks like the tests were broken by https://github.com/resin-io/resin-sdk/pull/324 @pimterry, as that PR failed with the same message as this when I retried it in https://ci.appveyor.com/project/resin-io/resin-sdk/build/1.0.1150/job/n961n0qw6yaopuy6

pimterry:
That PR is only changed Travis's config, so I can't believe it's the cause of Appveyor's failures. I'll take a closer look now though.

Page-:
You're right, when I initially looked I didn't see an appveyor.yml and thought it must be because it was defaulting to travis.yml, but it looks like something else has changed - actually I just checked and I'm pretty sure it's got to be due to coffee-script 1.12.5: http://coffeescript.org/#changelog

pimterry:
Eugh, ok, that makes sense. Explains why Travis passes too - it's using the yarn-locked versions.

(As an aside, I'd really like to make these consistent, and preferably find a way to ensure dependencies don't shift under the feet of downstream users either. Might be worth looking at shrinkwrap again. Issue: https://github.com/resin-io/resin-sdk/issues/326)

lib/models/os.coffee
```diff
@@ -54,8 +54,10 @@ getOsModel = (deps, opts) ->
 			{ versions, latest } = body
 			[ validVersions, invalidVersions ] = partition(versions, semver.valid)
 
-			validVersions.sort(osVersionRCompare)
```
pimterry: This no longer sorts the versions list, so the version list shown downstream won't be sorted at all. That definitely needs to stay I think.

pimterry: This also highlights that we don't have any tests for this code... If you'd like to add some that'd be wonderful, or you're welcome to just file an issue instead and I'll take a look at it soon.

Page-: You're right - I'll fix this, was just trying to avoid the sort thinking we didn't need it

lib/models/os.coffee
emirotin:
```
releaseVersions = reject versions, (version) ->
  semver.prerelease(version)
prodVersions = reject releaseVersions, (version) ->
  semver.prerelease(version) or _.includes(version, '.dev')
potentialRecommendedVersions = if prodVersions.length then prodVersions else releaseVersions
```

But I don't insist this is required

Page-:
Ok, I've added a test for the sorting @pimterry and as part of it I had to make the sorting be consistent - I think these changes need to be copied into the ui for the host update code though (although it might be worth taking the opportunity to deduplicate)

emirotin:
The HUP code is very similar but slightly different, thus the duplication

pimterry:
@Page- This actually breaks the tests - we've got some tests for the sorting logic itself here: https://github.com/resin-io/resin-sdk/blob/master/tests/util.spec.coffee#L156. Those need updating with this new change.

The travis build not failing from those is a bug that actually did come from that Travis PR, I've got a fix up at https://github.com/resin-io/resin-sdk/pull/328. Should be a quick review and merge there, and then we can rebase this on top to check this is all working correctly..

Page-:
Ok, I'll fix those @pimterry

Page-:
@pimterry I was able to fix it, and it was painfully hard to find :(  Apparently I put the `require` for `osVersionRCompare` in the nodejs only branch by accident, so it wasn't there in the browser env

pimterry:
Once you merge this, it'd be really useful if you could test the auto publish code too. Just push a tagged commit to master which adds the version number to package.json and the changelog, and it should get automatically picked up, built, and pushed to npm. Once I'm sure that's working versionbot can take over that step too, but I'd rather do it by hand until we've seen it all working happily at least once!

Page-:
I pushed a tagged commit @pimterry

PR #328, opened 2017-05-17
Fix small but breaking testing/CI bugs

pimterry:
This fixes issues we've hit in https://github.com/resin-io/resin-sdk/pull/325.

We should probably dig into the actual code that coffee script is now compiling incorrectly for the image maker, but I'd rather fix the tests themselves for now, and take a closer look at that later.


pimterry:
Thanks @emirotin! I'll merge this as soon as the tests pass.

Issue #337, opened 2017-06-28
Add app methods for enabling and disabling all device URLs


LucianBuzzo:
Currently there is no app wide method for doing this.

LucianBuzzo:
Connects to https://github.com/resin-io/resin-ui/issues/204

PR #338, opened 2017-06-28
Add application methods for enabling and disabling all device URLs

LucianBuzzo:
Connects to #337

change-type: minor


tests/integration/models/application.spec.coffee
```diff
@@ -191,36 +191,30 @@ describe 'Application Model', ->
 				promise = resin.models.application.getApiKey(999999)
 				m.chai.expect(promise).to.be.rejectedWith('Application not found: 999999')
 
+	describe 'when toggling device URLs', ->
+		beforeEach ->
+			resin.models.application.create('DeviceUrlsTestApp', 'raspberry-pi').then (application) =>
+				@application2 = application
```
pimterry: Looking closer, I _think_ you can drop the `application2` part of this. You can just use the existing `@application` - it doesn't matter if you mess with it, because we delete all apps and devices and recreate everything between every single test.

tests/integration/models/application.spec.coffee
```diff
@@ -191,36 +191,30 @@ describe 'Application Model', ->
...
+				resin.models.device.register(@application2.id, resin.models.device.generateUniqueKey())
+			.then (deviceInfo) =>
+				@deviceInfo2 = deviceInfo
```
pimterry: Is there a reason this needs the `2`? I can't see another `deviceInfo` in this file.

Issue #339, opened 2017-06-29
Test on master branch is failing


LucianBuzzo:
Running `npm test` on master fails with the following error:
```
1) Application Model given no applications resin.models.application.create() should be rejected if the name has less than three characters:
     AssertionError: expected promise to be rejected with an error including 'It is necessary that each app name that is of a user (Auth), has a Length (Type) that is greater than or equal to 4.' but got 'Request error: It is necessary that each app name that is of a user (Auth), has a Length (Type) that is greater than or equal to 4 and is less than or equal to 30.’
```

PR #340, opened 2017-06-29
Fix failing application name test

LucianBuzzo:
Connects to #339

change-type: patch


LucianBuzzo:
I'll add it to this PR.
+1 for using the phrase 'floats your boat' btw 😄

Issue #342, opened 2017-06-30
Add self-billing v0 related methods


thgreasi:
Refers to implementation for the [step 1 of the respective spec PR](https://github.com/resin-io/hq/pull/586#issuecomment-297467607).
HQ: https://github.com/resin-io/hq/issues/585

thgreasi:
Resolved by #346

PR #346, opened 2017-07-06
Add self-billing related methods

thgreasi:
Connects To: #342
HQ: resin-io/hq#585
Depends-On: resin-io/resin-api#340
Change-Type: minor

See: https://github.com/resin-io-modules/resin-request/issues/102
---- Autogenerated Waffleboard Connection: Connects to #342


emirotin:
I'll give you today the new version of resin-request so you can update this PR before merging

thgreasi:
OK thanks, that's great @emirotin :)

emirotin:
https://github.com/resin-io-modules/resin-request/pull/103/files

emirotin:
@thgreasi resin-request@8.5.0 at your disposal, sir

thgreasi:
OK thanks for the heads-up @emirotin
Thanks for the insta-PR + merge :+1:

lib/models/billing.coffee
```diff
@@ -0,0 +1,222 @@
...
+	# });
+	#
+	# @example
+	# resin.models.billing.getAccount(function(error, billingAccount) {
+	# 	if (error) throw error;
+	# 	console.log(billingAccount);
+	# });
+	###
+	exports.getAccount = (callback) ->
+		callback = findCallback(arguments)
```
pimterry: You only need this `findCallback` line (here and in all the functions below) when you have an optional argument before the callback, so it's not clear from the parameter positions which argument is the callback.

Here, with only one parameter, you can just drop it.

lib/models/billing.coffee
```diff
@@ -0,0 +1,222 @@
...
+
+	###*
+	# @summary Download a specific invoice
+	# @name downloadInvoice
+	# @public
+	# @function
+	# @memberof resin.models.billing
+	#
+	# @param {String} - an invoice number
+	# @fulfil {String|ReadableStream} - string blob on the browser, download stream on node
```
pimterry: This should be `Blob|ReadableStream`, and it's not a 'string blob', I think it's just a 'blob' (if it does have a mime type, I expect it's application/pdf - there's no string data here)

tests/integration/models/billing.spec.coffee
```diff
@@ -0,0 +1,257 @@
...
+						month: '8'
+						full_name: 'John Doe'
+					})
+				else
+					m.chai.expect(billingInfo).to.deep.equal({})
+
+	describe 'resin.models.billing.updateBillingInfo()', ->
+		# givenABillingAccountIt 'should return an billing account info object', ->
+		# 	promise = resin.models.billing.updateBillingInfo('')
+		# 	m.chai.expect(promise).to.become({})
```
pimterry: Commented out code - we should have a positive test in here (if possible?)

tests/integration/models/billing.spec.coffee
```diff
@@ -0,0 +1,257 @@
...
+
+	givenLoggedInUser()
+
+	hasActiveBillingAccount = false
+
+	withoutABillingAccountIt = (description, testFn) ->
+		it description, ->
+			if hasActiveBillingAccount
+				return this.skip()
+			testFn.apply(this, arguments)
```
pimterry: I think the `given` case is good, but this `without` behaviour isn't quite what there yet.

The issue is that here we end up here with tests that aren't run on the CI server. That means a user testing with a billing account locally won't notice if they fail, and neither will the CI server, and so somebody else later on will turn up and find everything broken (or worse: we'll have released the broken code and broken everything for real free users).

We should aim to be running _all_ the tests in CI, and only a subset locally for dev users that don't have a paid billing code set up.

I'm hoping there's a route through here where if you already have an account, we can use `updateInfo()` to remove it, run the free tests, and then put it back again afterwards. Does that sound plausible? This is a little tricky I know, ping me if you want to chat more about it.

README.md
```diff
@@ -94,6 +94,8 @@ You also have to provide the following environment variables from an account tha
 - `RESINTEST_REGISTER_PASSWORD`: The password of the account to register.
 - `RESINTEST_REGISTER_USERNAME`: The username of the account to register.
```
pimterry: This needs brief instructions about how to add billing to your test account, and what that means in practice (the paid tests will be skipped without it).

Ohh, interesting idea that fits with the 'how do we run free tests everywhere' point: should we actually just have devs add an optional `RESINTEST_RECURLY_TOKEN_ID` here? Could we use that to set up and easily turn off/on the billing info in the tests?

lib/models/billing.coffee
thgreasi: Oh, ok, thanks for the detailed heads up!

tests/integration/models/billing.spec.coffee
thgreasi: Removing it. I think that a positive test isn't possible since it would require `recurly.js` client library + the generated tokens are only valid for 20 minutes.

tests/integration/models/billing.spec.coffee
thgreasi: Did as [discussed](https://www.flowdock.com/app/rulemotion/resin-frontend/threads/K5YYmmRRSwWLWkFhJ5sGocbT6On) :bulb:.

README.md
thgreasi: Actually, since the recurly token is used by the API's env vars configuration, the best thing that I can thing is to use the new `RESINTEST_API_URL` var this this PR also adds and target the api server that you prefer. Does this sound good enough?

pimterry: @thgreasi, sorry, I have no idea what that means. Can you explain some more? What exactly would this look like, and what would it give us?

thgreasi:
@pimterry 
- [x] We need to take the changelog lines out, and add Change-Type to the commits instead
- [x] The commits needs squishing down a bit.
  - I always keep in mind to squash as much as it makes sense before merging. Kept the build on a separate commit for now to ease any future rebasing.
- [ ] The CI tests need to be configured with a paid user, so they're running the paid tests. You can set that up here: https://travis-ci.org/resin-io/resin-sdk/settings.
- [ ] The tests need to pass - presumably this is waiting on an API deploy?
  -  Yes, this `Depends-on: resin-io/resin-api#340`

thgreasi:
It seems that the API has been published.
Please re-review.
Added a separate commit with the generated changes to the build files, yarn.lock & the docs after running:
```
rm -r node_modules/ && yarn install && npm test && npm run docs
```
Should be fine, but I'm always cautious when I see +-diffs that I don't expect. Please take a look.
(Will squash is before merging if everything seems ok.)

The billing tests pass but interestingly enough 2 old tests fail now(!?):
* resin.models.device.hasDeviceUrl()
  * given device url is disabled
    1 "before each" hook for "should eventually be false given a device uuid"
* resin.models.key.get()
  2 should be rejected if the key id is invalid

lib/models/billing.coffee
```diff
@@ -0,0 +1,209 @@
...
+	#
+	# @example
+	# resin.models.billing.getInvoices().then(function(billingInfo) {
+	# 	console.log(billingInfo);
+	# });
+	#
+	# @example
+	# resin.models.billing.getInvoices(function(error, billingInfo) {
+	# 	if (error) throw error;
+	# 	console.log(billingInfo);
```
pimterry: Both these examples have `billingInfo` instead of `invoices`.

thgreasi: Nice catch :+1:

PR #352, opened 2017-07-21
Add function to allow looking up app based on name and owner username

CameronDiver:
This can be useful for looking up shared applications when the user in
question has an application of the same name as the one that is shared.

Connects-to: #349
Change-type: minor
Signed-off-by: Cameron Diver <cameron@resin.io>
---- Autogenerated Waffleboard Connection: Connects to #349


lib/models/application.coffee
```diff
@@ -149,6 +149,53 @@ getApplicationModel = (deps, opts) ->
...
+	#
+	# @example
+	# resin.models.application.getAppWithOwner('MyApp', 'MyUser').then(function(application) {
+	#		console.log(application);
+	# });
+	###
+	exports.getAppWithOwner = (appName, owner, options = {}, callback) ->
+		callback = findCallback(arguments)
+
+		a.pine.get
```
pimterry: What is `a`? I'm surprised this code works...

lib/models/application.coffee
```diff
@@ -149,6 +149,53 @@ getApplicationModel = (deps, opts) ->
...
+						$eq: [
+							$tolower: $: 'app_name'
+							appName
+						]
+					expand:
+						user:
+							$filter:
+								$eq: [
+									$tolower: $: 'username'
+									owner
```
pimterry: We should `toLowerCase` both the owner and appName here, if we're going to compare them to `$tolower`.

lib/models/application.coffee
CameronDiver: Oh shoot. I've copied the debugging code out of another file. It's essentially the same, but with different vars... Sorry for the noise.

lib/models/application.coffee
CameronDiver: Same as above, I've copied the wrong code...

CameronDiver:
@pimterry I was using resin-sdk-preconfigured to do the testing and accidentally committed the wrong code! Sorry about that. Anyway I've added the correct code and added some tests for the new method.

lib/models/application.coffee
```diff
@@ -149,6 +149,56 @@ getApplicationModel = (deps, opts) ->
...
+	#
+	# @param {String} appName - application name
+	# @param {String} owner - The owner's username
+	# @param {Object} [options={}] - extra pine options to use
+	# @fulfil {Object} - application
+	# @returns {Promise}
+	#
+	# @example
+	# resin.models.application.getAppWithOwner('MyApp', 'MyUser').then(function(application) {
+	#		console.log(application);
```
emirotin: Unneeded double tab?

tests/integration/models/application.spec.coffee
```diff
@@ -15,6 +15,13 @@ describe 'Application Model', ->
 				promise = resin.models.application.getAll()
 				m.chai.expect(promise).to.become([])
 
+		describe 'resin.models.application.getAppWithOwner()', ->
+
+			it 'should eventually reject', ->
+				promise = resin.models.application.getAppWithOwner('testapp', 'FooBar')
+				m.chai.expect(promise).to.be.rejected
+
+
```
emirotin: Extra empty line

Issue #356, opened 2017-07-27
New API version breaks a keys model test


thgreasi:
It seems that Pinejs v4 broke [this test](https://github.com/resin-io/resin-sdk/blob/94d1e5e8d8ef788d63fbbc4cabc6a3fc9e14874b/tests/integration/models/keys.spec.coffee#L77-L81) since it now rejects the request with `Request error: value "99999999999" is out of range for type integer`.
[Here is a travis build](https://travis-ci.org/resin-io/resin-sdk/jobs/256877834) that worked 3 days ago.

thgreasi:
Resolved by #357

PR #357, opened 2017-07-27
Fix keys test after API update

thgreasi:
Connects-To: #356
Change-Type: patch
---- Autogenerated Waffleboard Connection: Connects to #356


tests/integration/models/keys.spec.coffee
```diff
@@ -76,7 +76,9 @@ describe 'Key Model', ->
 
 			it 'should be rejected if the key id is invalid', ->
 				promise = resin.models.key.get(99999999999)
-				m.chai.expect(promise).to.be.rejectedWith('Request error: Internal Server Error')
+				m.chai.expect(promise).to.be.rejectedWith('Request error: value "99999999999" is out of range for type integer')
```
thgreasi: Should we just test for `rejectedWith('Request error')`? Since it seems that the error historically tends to change.

pimterry: Yeah, you're right, let's simplify it to just look for a request error generally.

I don't think we need the historical comments below either, since all this info is in the commit log. I'd just drop them.

PR #368, opened 2017-08-28
Detect Resin OS v1.0.0-pre devices and fill the os_version field

### LINKED ISSUE #367: Add a way to detect and report Resin OS v1.0.0-pre
The UI already has two different workarounds ([1](https://github.com/resin-io/resin-ui/blob/550fe6b6581efd86c612dcf706ecd2a6555ad585/src/scripts/pages/dashboard/controllers/device-controller.ts#L169) [2](https://github.com/resin-io/resin-ui/blob/550fe6b6581efd86c612dcf706ecd2a6555ad585/src/services/models/device-model-service.ts#L53-L59)) to detect and report Resin OS v1.0.0-pre but the CLI doesn't.
It would be great to move the detection in the SDK and use it in both places.

See: https://github.com/resin-io/resin-ui/issues/1193
See: https://www.flowdock.com/app/rulemotion/resin-frontend/threads/OC1Ui8wPz0KqEcKHUhiv1pyS9sL?filter=search&query=1193

thgreasi:
Resolves #367
Connects-To: #367
Change-Type: minor
---- Autogenerated Waffleboard Connection: Connects to #367


thgreasi:
Is there a way to mock devices sending supervisor statuses so I can add tests for this?

lib/models/application.coffee
```diff
@@ -146,6 +158,7 @@ getApplicationModel = (deps, opts) ->
 					if size(applications) > 1
 						throw new errors.ResinAmbiguousApplication(nameOrId)
 				.get(0)
+				.tap normalizeDeviceInfo
```
pimterry: Single line function calls should normally have brackets

pimterry: I'd probably refactor the `normalizeDeviceInfo` out, so we only do it once. I think if you put a single `tap` unindented just above `asCallback`, it should handle both cases in one go.

lib/util/device-os-version.coffee
```diff
@@ -0,0 +1,9 @@
+
```
pimterry: Extra empty line at the top here

lib/models/application.coffee
```diff
@@ -43,6 +46,12 @@ getApplicationModel = (deps, opts) ->
 
 	exports._getId = getId
 
+	normalizeDeviceInfo = (application) ->
```
pimterry: Can we call this `normalizeApplication`? Makes it clearer how we should use it in future, and it'd be a useful place for other similar normalizations, because I'm sure we'll want some more.

thgreasi:
> Is there a way we could do this with pine more reliably?

@pimterry It would be really interesting. I hoped that you had something in mind. Any suggestion on who I should ping for that?

pimterry:
Page might know. @Page-, in the SDK, we want to reliably transform all pine results of a certain type. E.g. in this PR, we want to fix the `os_version` on all devices returned by queries, even if they're in unexpected locations from a downstream-provided `$expand`.

Do you have any suggestions for good ways we could handle that? To just recursively search the body for every response, we'd need some clear marker of the type of each object all the way down - does that exist?

pimterry:
@thgreasi looks like the tests failed due to random instability, I've rerun them.

thgreasi:
Yea, this happens lately.

Issue #369, opened 2017-08-31
Support process-wide SDK configuration


pimterry:
This is a plan to allow us to use the resin SDK inside shared frontend components while inheriting the app-wide SDK configuration (e.g. which API url it's using, whether it's in debug mode, etc), and also use those components in an environment that doesn't already have an SDK.

In _addition_ to the current usage pattern:

* Create an SDK instance with one set of options
* Create an SDK instance with another set of options
* Use both independently

you should also be able to:

* At the top-level of your codebase, define the standard set of shared default options your whole process should be using.
* Ask for an SDK elsewhere, and automatically get one using these shared default options.

and:

* Ask for an SDK using the shared default options, and get an SDK using sensible options even though no defaults have been defined.

This could look something like:

```typescript
const ResinSdk = require('resin-sdk');

// Current/normal usage:
let sdk1 = ResinSdk(customOptions); // create a new SDK from a given set of options
let sdk2 = ResinSdk(otherOptions); // create a new SDK which uses totally different options

// Process-wide options usage:
ResinSdk.setSharedOptions(sharedDefaultOptions); // Set the shared default options
let sdk3 = ResinSdk.fromSharedOptions(); // Returns an SDK using the shared default options
```

As a safety guard, it's likely that calling `setSharedOptions` any time after somebody else has called `fromSharedOptions` should fail, since that means you've got a race condition and two parts of the code that explicitly asked to use the same options are not necessarily doing so.

It's not sufficient to store the shared options as a static variable within the SDK's state, because that isn't a true singleton (it's surprisingly easy to initialize the same module twice, and we may end up with multiple SDK installs & instances in our module tree, all of which should be using the same shared options). We could use symbols for this (see https://derickbailey.com/2016/03/09/creating-a-true-singleton-in-node-js-with-es6-symbols/), but a simple global with a unique & clear enough name should be fine.

We need to be happy this works when we have disagreements on the SDK version used. If I depend on SDK v7, and I call `sdk7.fromSharedOptions()`, and the options have been set by an SDK 6 instance, I should get back a working SDK v7 instance nonetheless. I think that should come out nicely in the wash, but it needs careful thought to make sure it won't go wrong in other weird cases.

[...]

PR #370, opened 2017-08-31
Add methods for generating an sdk instance using shared options

LucianBuzzo:
Connects to #369

change-type: minor


pimterry:
Somehow _all_ the builds failed due to occasional timeouts on the API, on both CI services? Very weird, presumably some API instability? I've retriggered them all anyway, lets see what happens.

lib/resin.coffee
```diff
@@ -219,3 +224,64 @@ module.exports = getSdk = (opts = {}) ->
...
+# @example
+# resin.setSharedOptions({
+# 	apiUrl: 'https://api.resin.io/',
+# 	imageMakerUrl: 'https://img.resin.io/',
+# 	apiVersion: 'v2',
+# 	isBrowser: true,
+# });
+###
+getSdk.setSharedOptions = (opts) ->
+	defaults(opts, defaultOpts)
```
pimterry: I don't think you need to do this (since it'll get done within the eventual `getSdk` call anyway), and I think doing so might actually break cross-version compatibility (since in that case we'll be setting the defaults from the version of the library used to set the options, not the version of the library that contains the actual SDK code that's going to be used).

lib/resin.coffee
```diff
@@ -219,3 +224,64 @@ module.exports = getSdk = (opts = {}) ->
...
+
+	root = window || GLOBAL
```
pimterry: This doesn't work in node - if `window` is not defined (as opposed to being defined with a value of undefined) then it'll throw an exception, not use `GLOBAL`.

lib/resin.coffee
```diff
@@ -219,3 +224,64 @@ module.exports = getSdk = (opts = {}) ->
...
+# If options have not been set using this method, then this method will use the
+# same defaults as the main SDK factory function.
+#
+# @params {Object} opts - The shared default options
+#
+# @example
+# const sdk = resin.fromSharedOptions();
+###
+getSdk.fromSharedOptions = ->
+	root = (window || GLOBAL)
```
pimterry: As above - this won't work

lib/resin.coffee
```diff
@@ -219,3 +224,64 @@ module.exports = getSdk = (opts = {}) ->
...
+# });
+###
+getSdk.setSharedOptions = (opts) ->
+	defaults(opts, defaultOpts)
+
+	root = window || GLOBAL
+
+	if root[RESIN_SDK_HAS_USED_SHARED_OPTIONS]
+		console.error('Shared SDK options have already been used. You may have a race condition in your code.')
+
```
pimterry: I think we should handle the case where you accidentally set the shared options twice too, since that's definitely a bug.

lib/resin.coffee
```diff
@@ -219,3 +224,64 @@ module.exports = getSdk = (opts = {}) ->
...
+###*
+# @summary Set shared default options
+# @name setSharedOptions
+# @public
+# @function
+# @memberof resin
+#
+# @description
+# Set options that are used by calls to `resin.fromSharedOptions()`.
+# The options accepted are the same as those used in the main SDK factory function.
```
pimterry: Should document here that this needs to be called as soon as possible during app startup, and definitely before any calls to `fromSharedOptions`

lib/resin.coffee
```diff
@@ -219,3 +224,64 @@ module.exports = getSdk = (opts = {}) ->
...
+#
+# @example
+# const sdk = resin.fromSharedOptions();
+###
+getSdk.fromSharedOptions = ->
+	root = (window || GLOBAL)
+
+	sharedOpts = root[RESIN_SDK_SHARED_OPTIONS]
+
+	defaults(sharedOpts, defaultOpts)
```
pimterry: You can do this, with no downside, but you don't need to, as getSdk does the same already.

LucianBuzzo: d'oh

lib/resin.coffee
```diff
@@ -58,13 +65,14 @@ sdkTemplate =
...
 
-module.exports = getSdk = (opts = {}) ->
+getSdk = (opts = {}) ->
 	defaults opts,
 		apiUrl: 'https://api.resin.io/'
 		imageMakerUrl: 'https://img.resin.io/'
 		apiVersion: 'v2'
 		isBrowser: window?
 
+
```
pimterry: Extra newline

tests/integration/resin.spec.coffee
```diff
@@ -101,3 +101,20 @@ describe 'Resin SDK', ->
 					m.chai.expect(called).to.equal true,
 						'responseError should be called when authentication fails'
 
+		describe 'getSdk.setSharedOptions()', ->
```
pimterry: Really these should be unindented a level - they're currently within `factory function`, but it's not calling the factory function (directly at least)

PR #372, opened 2017-09-01
DO NOT MERGE - run tests against staging

pimterry:
This is to help debug performance issues in the API, as these tests should reliably fail with timeouts, and we can enable profiling on staging to help us investigate that. See https://www.flowdock.com/app/rulemotion/resin-devops/threads/swpV-lO77T3GqUFRzDfowpeQGGY for details.
---- Autogenerated Waffleboard Connection: Connects to #374

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -6,14 +6,21 @@
 # it doesn't quite work.
 
 language: node_js
+env:
+  global:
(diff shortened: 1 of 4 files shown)


pimterry:
For some reason this isn't working properly with appveyor - I'm just going to ignore it, as the two travis builds should be enough for a repro

Issue #371, opened 2017-08-31
Add re-usable helper methods to the SDK


LucianBuzzo:
The resin dashboard currently uses a number of helper methods for presenting things like:
- The time a device was last seen
- The human readable device state
- OS download sizes

These should be moved to the SDK, so they can more easily be shared across components.

thgreasi:
All these are available at this point :tada:

PR #373, opened 2017-09-01
Added lastOnline() method to device model

LucianBuzzo:
Connects to #371

change-type: minor


tests/integration/models/device.spec.coffee
```diff
@@ -15,7 +15,7 @@ makeRequest = (url) ->
 				isError: !!err
 				response: res?.text
 
-describe 'Device Model', ->
+describe.only 'Device Model', ->
```
thgreasi: Leftover `.only`

Issue #378, opened 2017-09-05
Throw ResinSupervisorLockedError on reboot, shutdown and purge errors


MoranF:

MoranF:
Resolved by PR #379

PR #379, opened 2017-09-05
Add reboot, shutdown and purge methods to application model and throw ResinSupervisorLockedError

MoranF:
Connects-To: #378

Change-Type: minor
---- Autogenerated Waffleboard Connection: Connects to #378


lib/models/application.coffee
```diff
@@ -24,6 +24,8 @@ errors = require('resin-errors')
 
 { isId, findCallback, mergePineOptions, notFoundResponse, treatAsMissingApplication } = require('../util')
 
+LOCKED_STATUS_CODE = 423
```
LucianBuzzo: Is there anywhere we can put this constant so that it can be shared between files?

lib/models/application.coffee
```diff
@@ -446,6 +448,108 @@ getApplicationModel = (deps, opts) ->
...
+	# @name purge
+	# @public
+	# @function
+	# @memberof resin.models.application
+	#
+	# @param {Number} appId - application id
+	# @returns {Promise}
+	#
+	# @example
+	# resin.models.application.purge(123, 456);
```
LucianBuzzo: This example can be removed

lib/models/application.coffee
```diff
@@ -446,6 +448,108 @@ getApplicationModel = (deps, opts) ->
...
+
+	###*
+	# @summary Shutdown
+	# @name shutdown
+	# @public
+	# @function
+	# @memberof resin.models.application
+	#
+	# @param {Number} appId - application id
+	# @param {Boolean} force (optional)
```
LucianBuzzo: Can you change this to an options object so that it matches `resin.models.device.shutdown()`

lib/models/application.coffee
```diff
@@ -446,6 +448,108 @@ getApplicationModel = (deps, opts) ->
...
+
+	###*
+	# @summary Reboot
+	# @name reboot
+	# @public
+	# @function
+	# @memberof resin.models.application
+	#
+	# @param {Number} appId - application id
+	# @param {Boolean} force (optional)
```
LucianBuzzo: Can you change this to an options object so that it matches `resin.models.device.shutdown()`

lib/models/application.coffee
```diff
@@ -446,6 +448,108 @@ getApplicationModel = (deps, opts) ->
...
+	# @name shutdown
+	# @public
+	# @function
+	# @memberof resin.models.application
+	#
+	# @param {Number} appId - application id
+	# @param {Boolean} force (optional)
+	# @returns {Promise}
+	#
+	# @example
```
LucianBuzzo: These examples need to be updated

lib/models/application.coffee
```diff
@@ -446,6 +448,108 @@ getApplicationModel = (deps, opts) ->
...
+	# resin.models.application.reboot(123);
+	#
+	# @example
```
LucianBuzzo: These examples need to be updated

lib/models/application.coffee
MoranF: I guess it creates a global constant, so I removed it from the application file and kept it only in the device file

LucianBuzzo: @MoranF coffeescript will simple hoist the variable to the top of the file, not create a global, so this change won't work :(

lib/models/application.coffee
```diff
@@ -529,20 +521,17 @@ getApplicationModel = (deps, opts) ->
 	# resin.models.application.reboot(123);
 	#
 	# @example
-	# resin.models.application.reboot(123, 456);
-	#
-	# @example
-	# resin.models.application.reboot(123, 456, true);
+	# resin.models.application.reboot(123, true);
```
LucianBuzzo: This example doesn't match the method

lib/models/application.coffee
```diff
@@ -494,20 +489,17 @@ getApplicationModel = (deps, opts) ->
 	# resin.models.application.shutdown(123);
 	#
 	# @example
-	# resin.models.application.shutdown(123, 456);
-	#
-	# @example
-	# resin.models.application.shutdown(123, 456, true);
+	# resin.models.application.shutdown(123, true);
```
LucianBuzzo: This example doesn't match the method

lib/models/application.coffee
```diff
@@ -446,6 +446,99 @@ getApplicationModel = (deps, opts) ->
...
+	#
+	# @param {Number} appId - application id
+	# @param {Boolean} force (optional)
+	# @returns {Promise}
+	#
+	# @example
+	# resin.models.application.shutdown(123);
+	#
+	# @example
+	# resin.models.application.shutdown(123, true);
```
LucianBuzzo: This example needs updating

lib/models/application.coffee
```diff
@@ -446,6 +446,99 @@ getApplicationModel = (deps, opts) ->
...
+	#
+	# @param {Number} appId - application id
+	# @param {Boolean} force (optional)
+	# @returns {Promise}
+	#
+	# @example
+	# resin.models.application.reboot(123);
+	#
+	# @example
+	# resin.models.application.reboot(123, true);
```
LucianBuzzo: This example needs updating

lib/models/application.coffee
```diff
@@ -446,6 +446,95 @@ getApplicationModel = (deps, opts) ->
...
+		return request.send
+			method: 'POST'
+			url: '/supervisor/v1/purge'
+			baseUrl: apiUrl
+			body:
+				appId: appId
+				data:
+					appId: "#{appId}"
+		.catch (err) ->
+			if err.statusCode == LOCKED_STATUS_CODE
```
LucianBuzzo: This is undefined at the moment. Can you add the `LOCKED_STATUS_CODE` in `util` so it can be shared between `device.coffee` and `application.coffee`?

lib/models/application.coffee
```diff
@@ -446,6 +446,95 @@ getApplicationModel = (deps, opts) ->
...
+			body:
+				appId: appId
+				data:
+					force: Boolean(options.force)
+		.catch (err) ->
+			if err.statusCode == LOCKED_STATUS_CODE
+				throw new errors.ResinSupervisorLockedError()
+
+			throw err
+
```
LucianBuzzo: Can you have these new application methods accept and optional callback, the same way that the device methods do?

lib/models/application.coffee
MoranF: where exactly? The util.coffee looks almost empty

LucianBuzzo: Do you mean this file --> https://github.com/resin-io/resin-sdk/blob/master/tests/util.coffee
Thats a file for test utilities.
You should put it in here I think --> https://github.com/resin-io/resin-sdk/blob/master/lib/util/index.coffee

thgreasi:
My checklist:
- [ ] I would expect this PR to be named something like `Add reboot(), shutdown() and purge() methods to application model`
- since it adds new methods
  * [ ] It should be a `minor` release
  * [ ] it should update the docs since
- [ ] ideally it should contain 2 commits:
  * `Add reboot(), shutdown() and purge() methods to application model`
  * `Add supervisor locked error throwing to device methods`

Issue #383, opened 2017-09-06
Command "Ping" is not working properly


horia-delicoti:

pimterry:
Fixed in 6.14.3.

PR #384, opened 2017-09-06
Fix command PING to work properly

horia-delicoti:
Connects-To: #383
Change-Type: patch

---- Autogenerated Waffleboard Connection: Connects to #383


pimterry:
@horia-delicoti the travis build here is failing because it needs the env vars from our travis settings, which contain credentials used by the tests, and those aren't available to PRs from forked repositories.

Can you make these changes on a branch in this repo, and then open a PR from that? I think you should have permissions for that, but let me know if not. I'm going to close this in the meantime.

PR #387, opened 2017-09-11
Fix command PING to work properly

horia-delicoti:
Connects-To: #383
Change-Type: patch
---- Autogenerated Waffleboard Connection: Connects to #383


pimterry:
Looks good now. You'll have to rebase though, and for some reason it looks like versionbot isn't happy with your commit tags - I think it's because there's a new line after the change-type, and it only considers tags used in the footer (after the last empty line) to count.

Issue #389, opened 2017-09-12
Should use balena-semver module instead of semver


LucianBuzzo:
This means we can homogenize our normalization methods across all frontend components.

PR #390, opened 2017-09-12
Use resin-semver instead of semver module

LucianBuzzo:
Connects to #389

change-type: patch

Depends on https://github.com/resin-io-modules/resin-semver/pull/55


lib/models/os.coffee
```diff
@@ -85,21 +85,12 @@ getOsModel = (deps, opts) ->
...
 	# utility method exported for testability
 	exports._getMaxSatisfyingVersion = (versionOrRange, osVersions) ->
 		if versionOrRange in [ 'default', 'latest', 'recommended' ]
 			return osVersions[versionOrRange]
 
-		semverVersions = osVersions.versions.map(fixNonSemver)
-		maxVersion = semver.maxSatisfying(semverVersions, fixNonSemver(versionOrRange))
-
-		return unfixNonSemver(maxVersion)
+		return semver.maxSatisfying(semverVersions, fixNonSemver(versionOrRange))
```
pimterry: Can we move `fixNonSemver` handling of the range into `resin-semver` too? It'd be nice to be able to remove that from here entirely.

lib/util/index.coffee
```diff
@@ -43,45 +43,15 @@ exports.treatAsMissingDevice = (uuidOrId) ->
...
-	if revA isnt revB
-		return revB - revA
-
-	devA = exports.isDevelopmentVersion(versionA)
-	devB = exports.isDevelopmentVersion(versionB)
-	if devA isnt devB
-		return devA - devB
-
-	return versionA.localeCompare(versionB)
+	return semver.rcompare(versionA, versionB)
```
pimterry: This function is now completely equivalent to `rSemver.rcompare`, so we might as well just remove it and use that directly everywhere.

lib/models/os.coffee
LucianBuzzo: I think this is functionally equivalent to https://github.com/resin-io-modules/resin-semver/blob/master/src/index.js#L11 what do you think?

pimterry: There are potential differences, but I do agree it's close enough that it should work for everything we care about - i.e. if you just safeSemver the range inside resin-semver, I think everything'll be fine.

Note that @thgreasi is in the process of drastically updating that function (https://github.com/resin-io-modules/resin-semver/pull/35/files#diff-f41e9d04a45c83f3b6f6e630f10117fe). I think this is still fine though :+1:.

pimterry:
@LucianBuzzo looks like you're rebasing this - there's still some big conflicts here (for example, since this PR all the build files have been removed from the SDK, and right now this PR readds them all), and we're about to have more because api-v3 and the enormous v7 branch is hopefully going to land imminetly (as soon as Page is up and reviews it).

Can I ask nicely that we just leave this for a day or two? Should be easier to sort out once all that is out of the way.

LucianBuzzo:
@pimterry Can do 👍

PR #396, opened 2017-09-20
Improve performance by selecting only the fields required where possible

pimterry:
This reduces load on the API, and traffic on the wire, by transferring only the data on objects that we need, rather than the full object.

This isn't done for every case yet: this covers all the obvious and automatically tested cases, and there's definitely more we could do later. This also doesn't do this for any cases that would be sensible but breaking - that's a whole separate problem.

There's one notable fundamental change in here: we now enforce that any query for an object _always_ returns the id field, even if you try to select it away. Otherwise lots of the existing code depends on results including the id, and including it automatically makes this fix a lot simpler.


lib/util/index.coffee
```diff
@@ -123,7 +123,16 @@ exports.mergePineOptions = (defaults, extras) ->
 
 	for own option, value of extras
 		switch option
-			when 'select', 'orderby', 'top', 'skip'
+			when 'select'
+				if value?
+					if not isArray(value)
+						value = [value]
+					if value.indexOf('id') == -1
```
thgreasi: How about `_.includes`?

lib/util/device-os-version.coffee
```diff
@@ -2,7 +2,7 @@ isEmpty = require('lodash/isEmpty')
 { isProvisioned } = require('.')
 
 exports.normalizeDeviceOsVersion = (device) ->
-	if isEmpty(device.os_version) and isProvisioned(device)
+	if device.os_version? and isEmpty(device.os_version) and isProvisioned(device)
```
thgreasi: Would it make more sense to use `os_version in device`?

pimterry: I think I'd rather keep it this way, since it's a bit more explicit and the TS equivalent (`device.os_version != null`) is enough to let TS infer that `device.os_version` is `undefined | null` (and so catch errors better), which the alternative doesn't. Small thing, but I'd rather lean in that direction.

lib/util/index.coffee
pimterry: Done!

lib/util/device-os-version.coffee
thgreasi: So, my concern was that,
if you didn't `select` the `os_version`, but you `select` the `supervisor_version` & `last_connectivity_event` and the `isProvisioned` test passes, then you **will** get a `'Resin OS 1.0.0-pre'` `os_version` which might be incorrect.
On the other hand, if you didn't select it, then we don't expect that you will be using it.

pimterry: I don't think that's true. Because this is `and` all the way through, if you don't select `os_version` then `device.os_version?` is false, and so this does nothing.

This code only gives you the `-pre` version if _all_ the below are true:

* Your query did include an os_version result
* That result is an empty string
* The device is provisioned

I don't think there's any case where you do that, but it's not genuinely a `1.0.0-pre` device. Am I missing something?

(More generally: if you think there's a bug or incorrect behaviour like this, don't approve the PR! I'm very happy to discuss things like this before merge, it's much better than accidentally releasing a bug somewhere)

Issue #358, opened 2017-07-31
Creating an application with a discontinued device type should fail


pimterry:
HQ: resin-io/hq#785
See: resin-io/resin-ui#998

thgreasi:
As far as I could find in the already deprecated repos, [the condition seems to be `state == 'discontinued'`](https://github.com/resin-os/resin-parallella/blob/73583416a7289cc709a9e74f20dd8caef3e6c716/parallella-hdmi-resin.coffee#L11).

pimterry:
@thgreasi good point, updated.

thgreasi:
@pimterry I also think that we currently [do not filter at all the supported device types](https://github.com/resin-io/resin-sdk/blob/c7b6bd60f74660f3df05ec436269b75913afc13c/lib/models/device.coffee#L1159-L1162).
The interested question is whether we should also filter out both the `DEPRECATED` (the UI already does this manually) and the `discontinued` states. (Loved the inconsistent casing!)

pimterry:
I think we still don't want to do any filtering there, even after this change. The UI and other places still need to know about these discontinued device types, so they can handle existing applications using it, it's only in application creation behaviour that anything major will change (plus potential UI changes in how we display these device types).

thgreasi:
OK!

pimterry:
Fixed in 7.0.0

PR #398, opened 2017-09-22
Breaking: Don't allow creating devices with discontinued device types

pimterry:
This fixes this issue from a month or so ago: https://github.com/resin-io/resin-sdk/issues/358.

This is a breaking change (because users currently might be automatically creating `edge` devices, and that will now fail). To handle that, we're not going to merge this into master or use versionbot: we're going to merge quietly into a v7 branch for now, along with other upcoming breaking changes (API v3 & API keys), and then merge these all together with versionbot to master, to do a single breaking v7 release with all 3 included.

Connects-To: #358
Change-Type: major (!!!)
HQ: resin-io/hq#785
See: resin-io/resin-ui#998
---- Autogenerated Waffleboard Connection: Connects to #358


lib/models/application.coffee
```diff
@@ -341,13 +341,16 @@ getApplicationModel = (deps, opts) ->
...
+		deviceManifestPromise = deviceModel().getManifestBySlug(deviceType)
 		.tap (deviceSlug) ->
 			if not deviceSlug?
 				throw new errors.ResinInvalidDeviceType(deviceType)
 
-		return Promise.all([ deviceSlugPromise, parentAppPromise ])
-		.then ([ deviceSlug, parentApplication ]) ->
+		return Promise.all([ deviceManifestPromise, parentAppPromise ])
+		.then ([ deviceManifest, parentApplication ]) ->
+			if deviceManifest.state == 'DISCONTINUED'
```
thgreasi: How about adding
```coffee
throw new errors.ResinInvalidDeviceType(deviceType) if !deviceManifest
```
?

pimterry: I think we discussed this at summit, and it is already covered by the `.tap` check above.

thgreasi: Opps, forgot to approve right after!

Issue #400, opened 2017-09-26
Remove generated browser build files


pimterry:
We'll still keep generated docs in this repo - removing generated output is purely to let us avoid the many conflicts they produce (minified versions especially).

Fortunately, since we publish to npm, the canonical published files live there for each version anyway, so we can directly switch to those if  we use [unpkg.com](https://unpkg.com) to let users directly reach them.

PR #401, opened 2017-09-26
Uncommit (though still publish) generated browser files

pimterry:
Generated files are now available via npm, via unpkg in the browser for users who want a direct non-npm URL. This simplifies the repo, stops devs having to remember to commit changes, and drastically reduces the number of git conflicts we'll hit.

I only just spotted this, and I'd only heard of unpkg recently. Seems like a big easy win. Anything I'm missing? Any reason this might not work somewhere?

Connects-To: #400 
Change-Type: patch
---- Autogenerated Waffleboard Connection: Connects to #400


pimterry:
As long as those people are using commit hashes, they'll be fine, and they can upgrade ok in their own time (and they won't be doing so automatically, so no big breaking change surprises).

If they're literally just referencing the master branch at `https://rawgit.com/resin-io/resin-sdk/master/build/resin-browser.min.js` then they've got zero guarantees anyway, and while having to move to unpkg might be annoying, if they don't then they're going to be _really_ unhappy when the entire API changes under their feet for v7 and v8 :smiley:.

PR #403, opened 2017-09-27
Explicitly include build output in packaged releases

pimterry:
I have on idea why the build folder isn't already included, but I'm reasonably confident that explicitly adding it will work.

Note that according to https://docs.npmjs.com/misc/developers README, CHANGELOG, LICENSE and package.json are all included automatically.

You can see the resulting included files at https://unpkg.com/resin-sdk@6.14.4-publish.test/ (based on be doing a local publish with the tests disabled for speed, but I believe that should be representative).

Change-Type: patch


package.json
```diff
@@ -8,6 +8,10 @@
     "type": "git",
     "url": "git://github.com/resin-io/resin-sdk.git"
   },
+  "files": [
+    "build/",
+    "DOCUMENTATION.md"
```
thgreasi: Should we also include the `yarn.lock`?
I'm not sure whether this makes any sense but I've seen that some of the dependencies of the UI include their's:
* awesome-typescript-loader
* css-to-react-native
* husky
* hyphenate-style-name
* pubnub
* react-aria-tooltip/node_modules/css-to-react-native
* styled-grid/node_modules/css-to-react-native

pimterry: I'm not sure either way if that's how yarn is supposed to work, no idea.

Regardless though, I'm planning to kill yarn in the SDK anyway (it's useful for a components like the UI where we control the install process, but since we don't, it really just means our tests aren't necessarily representative of the real result - https://github.com/resin-io/resin-sdk/issues/326), so publishing without it is fine for now I think.

thgreasi: :+1:

PR #406, opened 2017-09-28
Fix a deprecation warning, plus other small tidyup

pimterry:
A few small fixes and tweaks spotted while doing the big v3 migration, pulled into a separate PR for nice easy reviewing.


lib/resin.coffee
```diff
@@ -251,7 +251,7 @@ getSdk = (opts = {}) ->
 # });
 ###
 getSdk.setSharedOptions = (opts) ->
-	root = if window? then window else GLOBAL
+	root = if window? then window else global
```
thgreasi: How about creating a single global object provider?
Also, in the past I've used:
```
typeof window !== 'undefined' ? window :
    typeof self !== 'undefined' ? self :
    global
```
so that it also works inside workers. It might be an overkill here though.

pimterry: Good idea! It would definitely be nice to try and get this working in web workers too. I've updated it to use almost exactly that logic.

lib/resin.coffee
```diff
@@ -30,6 +30,15 @@ RESIN_SDK_SHARED_OPTIONS = 'RESIN_SDK_SHARED_OPTIONS'
...
 RESIN_SDK_HAS_SET_SHARED_OPTIONS = 'RESIN_SDK_HAS_SET_SHARED_OPTIONS'
 
+# Use window (web)/self (web worker)/global (node) as appropriate
+globalEnv = if typeof window != 'undefined'
+	window
+else if typeof self != 'undefined'
+	self
+else if typeof global != 'undefined'
+	global
+else null
```
thgreasi: How about returning an empty object instead of `null`? This way the code will continue to work even on that case. It could be harder to debug though.

pimterry: Yeah, I thought about that, but I think it's better if it explodes, which is why this explicitly avoids that. I have added a comment though, because you're right, this isn't clear.

This is only for the `sharedOptions` functions, and if you try to call those in a state where we can't guarantee that the options will be reliably shared (e.g. if you share them in an object here, they'll be shared for one install of this the SDK, but not _all_ installs of it, if you have npm conflicts etc) then imo it's better for it to make that clear, so you can handle it yourself properly.

thgreasi: Oh, fair enough, I didn't have that much context before and didn't think of that that way.
PS: Liked the `explodes` part :boom: !

PR #407, opened 2017-09-28
Break everything we can pre-v3

pimterry:
API v3 is an excellent excuse to bring it lots of important but breaking changes I've been putting off for a while. This PR works on the v2 API but changes behaviour by:

* Remove lots of our model add-ons 
  * These aren't currently reliable, as they mean you get different objects depending on whether you expand an object or query it directly. Really this logic should live in the API, but right now it's not strictly necessary at all (it's only used in the CLI afaik), so let's just kill it
  * This is also necessitated by the next step, which is hard to do while keeping these properties
* Stop expanding relationships by default
  * This is going to significantly reduce unnecessary load on the API, by letting the UI do much cheaper queries.
  * `getApplicationName` is now the only `expand:` left in the codebase, and I think it's reasonable and unavoidable.

This is another PR in the queue for SDK v7, so no versionbot, and merging to the `v7` branch instead of master until we're ready to ship.

### DIFF
--- a/lib/models/application.coffee
+++ b/lib/models/application.coffee
@@ -83,7 +83,6 @@ getApplicationModel = (deps, opts) ->
 				options:
 					mergePineOptions
 						orderby: 'app_name asc'
-						expand: 'device'
 						filter:
 							user: userId
 					, options
@@ -187,15 +186,14 @@ getApplicationModel = (deps, opts) ->
 						$eq: [
 							$tolower: $: 'app_name'
 							appName
-						]
-					expand:
+						],
 						user:
-							$filter:
-								$eq: [
+							$any:
+								$alias: 'u',
+								$expr: $eq: [
 									$tolower: $: 'username'
 									owner
 								]
-							$select: 'id'
 				, options
 		.tap (applications) ->
 			if isEmpty(applications)
--- a/lib/models/application.coffee
+++ b/lib/models/application.coffee
@@ -156,7 +156,7 @@ getApplicationModel = (deps, opts) ->
 
 	###*
 	# @summary Get a single application using the appname and owner's username
-	# @name getAppWithOwner
+	# @name getAppByOwner
 	# @public
 	# @function
 	# @memberof resin.models.application
@@ -168,11 +168,11 @@ getApplicationModel = (deps, opts) ->
 	# @returns {Promise}
 	#
 	# @example
-	# resin.models.application.getAppWithOwner('MyApp', 'MyUser').then(function(application) {
+	# resin.models.application.getAppByOwner('MyApp', 'MyUser').then(function(application) {
 	# 	console.log(application);
 	# });
 	###
-	exports.getAppWithOwner = (appName, owner, options = {}, callback) ->
+	exports.getAppByOwner = (appName, owner, options = {}, callback) ->
 		callback = findCallback(arguments)
 
 		appName = appName.toLowerCase()
--- a/lib/models/application.coffee
+++ b/lib/models/application.coffee
@@ -88,11 +88,7 @@ getApplicationModel = (deps, opts) ->
 							user: userId
 					, options
 
-		# TODO: It might be worth to do all these handy
-		# manipulations server side directly.
 		.map (application) ->
-			application.online_devices = filter(application.device, is_online: true).length
-			application.devices_length = application.device?.length or 0
 			normalizeApplication(application)
 			return application
 
--- a/lib/models/device.coffee
+++ b/lib/models/device.coffee
@@ -767,7 +767,7 @@ getDeviceModel = (deps, opts) ->
 		.then ({ application, device }) ->
 
 			if device.device_type isnt application.device_type
-				throw new Error("Incompatible application: #{applicationNameOrId}")
+				throw new errors.ResinInvalidDeviceType("Incompatible application: #{applicationNameOrId}")
 
 			return pine.patch
 				resource: 'device'
--- a/lib/models/device.coffee
+++ b/lib/models/device.coffee
@@ -149,7 +149,6 @@ getDeviceModel = (deps, opts) ->
 			resource: 'device'
 			options:
 				mergePineOptions
-					expand: 'application'
 					orderby: 'name asc'
 				, options
 
@@ -269,10 +268,7 @@ getDeviceModel = (deps, opts) ->
 				pine.get
 					resource: 'device'
 					id: uuidOrId
-					options:
-						mergePineOptions
-							expand: 'application'
-						, options
+					options: options
 				.tap (device) ->
 					if not device?
 						throw new errors.ResinDeviceNotFound(uuidOrId)
@@ -281,7 +277,6 @@ getDeviceModel = (deps, opts) ->
 					resource: 'device'
 					options:
 						mergePineOptions
-							expand: 'application'
 							filter:
 								uuid: $startswith: uuidOrId
 						, options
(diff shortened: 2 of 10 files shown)


pimterry:
@Page- I've mostly added you to this PR just because it'll make you happy: since we're breaking the SDK's API to support API v3, I'm removing all the automatic expands everywhere too, which should make a world of difference.

thgreasi:
Unbelievable title :rofl:

lib/models/device.coffee
```diff
@@ -117,9 +117,6 @@ getDeviceModel = (deps, opts) ->
 		return url.resolve(dashboardUrl, "/apps/#{options.appId}/devices/#{options.deviceId}/summary")
 
 	addExtraInfo = (device) ->
-		# TODO: Move this to the server
-		device.application_name = device.application[0].app_name
-		device.dashboard_url = getDashboardUrl({ appId: device.application[0].id, deviceId: device.id })
```
Page-: Strictly speaking the ui doesn't give a damn about the app id in dashboard, and using the wrong app id will still get you the right device (something that I abuse fairly often when I know the device id but not app id) - with that in mind it might be possible to give a url direct to a device that the dashboard can understand, cc @LucianBuzzo

lib/models/device.coffee
```diff
@@ -117,9 +117,6 @@ getDeviceModel = (deps, opts) ->
 		return url.resolve(dashboardUrl, "/apps/#{options.appId}/devices/#{options.deviceId}/summary")
 
 	addExtraInfo = (device) ->
-		# TODO: Move this to the server
-		device.application_name = device.application[0].app_name
```
Page-: This is also nice for a multi-app future 👍

lib/models/device.coffee
```diff
@@ -389,7 +381,11 @@ getDeviceModel = (deps, opts) ->
 	# });
 	###
 	exports.getApplicationName = (uuidOrId, callback) ->
-		exports.get(uuidOrId, select: 'application_name').get('application_name').asCallback(callback)
+		exports.get uuidOrId,
+			expand: 'application': $select: 'app_name'
```
Page-: You can drop the quotes from `'application'`

Page-: Can this be:
```
select: id
expand: 'application': $select: 'app_name'
```
to avoid fetching the device fields we don't care about for this case (or invert the query to be on application and filter by device/(id|uuid) - I expect that's a lot harder though)

lib/models/device.coffee
```diff
@@ -770,7 +766,7 @@ getDeviceModel = (deps, opts) ->
 		.then ({ application, device }) ->
 
 			if device.device_type isnt application.device_type
-				throw new Error("Incompatible application: #{applicationNameOrId}")
+				throw new errors.ResinInvalidDeviceType("Incompatible application: #{applicationNameOrId}")
```
Page-: Should this be backported?

lib/models/application.coffee
```diff
@@ -191,15 +186,14 @@ getApplicationModel = (deps, opts) ->
 						$eq: [
 							$tolower: $: 'app_name'
 							appName
-						]
-					expand:
+						],
 						user:
```
Page-: This can be backported right?

lib/models/device.coffee
pimterry: True. I don't think that matters much for this though - for now it's only used in the CLI, which is going to want to expand the app anyway, so I'll just extract the same logic unchanged to there and keep things simple. In a multi-app future, we'll definitely want device only URLs, but for now I'm not too worried.

lib/models/application.coffee
pimterry: I don't think it can? Before we were expanding user, so you'd get it in your result (but only the id), now we're not expanding it at all.

lib/models/device.coffee
pimterry: It's a bit debatable. I think there are a few weird cases where this _could_ break your code (if you're catching ResinInvalidDeviceType specifically for some other reason, expecting `move` to fail differently). That's a bit tenuous, but there's no desperate need to backport things like this, so I'd rather bundle it into the big clear break just to be safe.

lib/models/device.coffee
pimterry: Done (the simpler option - the 2nd is indeed much harder)

pimterry: Done

lib/util/index.coffee
```diff
@@ -128,8 +128,6 @@ exports.mergePineOptions = (defaults, extras) ->
 				if value?
 					if not isArray(value)
 						value = [value]
-					if !includes(value, 'id')
```
thgreasi: Oh so now we will have to explicitly include the ID :+1: More deterministic result object structure is on the way :)

lib/models/device.coffee
thgreasi: The UI will also need a getDashboardUrl method for resin-io/resin-ui#1351 .
Would it make sense to change that method to the suggested url and make this PR depend on the UI ticket?

lib/util/index.coffee
pimterry: Yep, it was just there before because lots of the object normalisation stuff needed it - now that's gone, this can go too :-)

lib/models/device.coffee
pimterry: I imagine the UI wouldn't want a URL method anyway, since it can use internal routing mechanisms instead (tbc, no idea what our routing will look like in React land). And to be clear, the CLI definitely will keep this functionality, it's just where the logic lives that'll be changing!

lib/models/environment-variables.coffee
```diff
@@ -229,9 +229,7 @@ getEnvironmentVariablesModel = (deps, opts) ->
 			return pine.get
```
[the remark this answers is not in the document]
pimterry: There's a few cases like this throughout the SDK, where we _always_ make the upfront request for the id, even though we might not need to. They exist because we want to consistently fail with a missing device error if the device doesn't exist. If we only make the single request below (to look up env vars for the device id) then we can't tell the difference between "the device doesn't exist" and "the device has no env vars". I'd like to avoid that in future, but unless there's an easy answer I'm planning to ignore it for now.

lib/models/application.coffee
```diff
@@ -191,15 +186,14 @@ getApplicationModel = (deps, opts) ->
 						$eq: [
 							$tolower: $: 'app_name'
 							appName
-						]
-					expand:
```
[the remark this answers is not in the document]
pimterry: Good point! To solve this, let's just rename the method :smiley:. It's now `getAppByOwner`. Clearer?

lib/models/application.coffee
```diff
@@ -191,15 +186,14 @@ getApplicationModel = (deps, opts) ->
...
+						],
```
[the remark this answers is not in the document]
pimterry: Our linter doesn't care (it gets run automatically during builds, so if it did care, it'd fail the PR). We're going to rewrite this into TS and add prettier etc soon-ish anyway, I'm not too worried about minor style points either way.

lib/models/application.coffee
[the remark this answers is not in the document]
thgreasi: :+1:

lib/models/environment-variables.coffee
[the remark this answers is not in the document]
thgreasi: How about expanding the `device_environment_variable`s through the `deviceModel().get` call?
That should throw the "the device doesn't exist" error properly? It might be an overkill though and I'm happy with the current implementation.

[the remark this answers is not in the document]
pimterry: Hmmm, yes, you're probably right that there's a route through there (though it may be messy). I don't want to worry about it right now, but I've filed an issue: https://github.com/resin-io/resin-sdk/issues/409

PR #410, opened 2017-10-02
Replace generateApiKey with generateProvisioningKey

pimterry:
This is already deprecated in the API, and we really shouldn't be using or exposing it here. This replaces it with provisioning keys, which are the modern alternative.

This is another fun breaking change, so another inclusion for big lovely v7.

### DIFF
--- a/tests/integration/models/application.spec.coffee
+++ b/tests/integration/models/application.spec.coffee
@@ -186,24 +186,24 @@ describe 'Application Model', ->
 				promise = resin.models.application.remove(999999)
 				m.chai.expect(promise).to.be.rejectedWith('Application not found: 999999')
 
-		describe 'resin.models.application.getApiKey()', ->
+		describe 'resin.models.application.generateProvisioningKey()', ->
 
-			it 'should be able to generate an API key by name', ->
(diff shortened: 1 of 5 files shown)


tests/integration/models/application.spec.coffee
```diff
@@ -186,24 +186,24 @@ describe 'Application Model', ->
...
 				m.chai.expect(promise).to.be.rejectedWith('Application not found: 999999')
 
-		describe 'resin.models.application.getApiKey()', ->
+		describe 'resin.models.application.generateProvisioningKey()', ->
 
-			it 'should be able to generate an API key by name', ->
-				resin.models.application.getApiKey(@application.app_name).then (apiKey) ->
-					m.chai.expect(_.isString(apiKey)).to.be.true
-					m.chai.expect(apiKey).to.have.length(32)
+			it 'should be able to generate an provisioning key by name', ->
```
thgreasi: Is "a provisioning key" better?

pimterry: Damn, good spot. Fixed.

pimterry:
I'm going to wait for @Page- to confirm before I merge this, just because I'm not confident my understanding of the provisioning key work is totally correct.

PR #414, opened 2017-10-05
Stop requests that don't need auth from failing if you have an invalid token

pimterry:
This depends on https://github.com/resin-io-modules/resin-request/pull/105, and it's going to fail until that's been approved and merged and released. Tested locally together though and this works nicely.

This is part of the solution for https://github.com/resin-io/resin-cli/issues/443, to stop the whole CLI breaking if you have an invalid token on disk.


lib/auth.coffee
```diff
@@ -342,6 +342,7 @@ getAuth = (deps, opts) ->
 			url: '/user/register'
 			baseUrl: apiUrl
 			body: credentials
+			sendToken: false
```
thgreasi: :+1:

Issue #415, opened 2017-10-09
Node 8 support


Page-:
Should we add node 8 to the ci builds in order to guarantee we support node 8?

pimterry:
Good spot! Yes, we should test against node 8. We should also deprecate Node 4 really, since we're breaking everything else, and the CLI already dropped support months ago.

For now I've opened a PR adding node 8 to the test targets on travis against master, and a PR to drop node 4 against the v7 branch.

PR #416, opened 2017-10-10
Make sure we're running the test suite against Node 8

pimterry:
Connects-To: #415

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -9,6 +9,8 @@ language: node_js
 matrix:
   include:
     - node_js:
+      - '8'
+    - node_js:
       - '6'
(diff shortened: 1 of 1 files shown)


thgreasi:
Should we also configure new user credentials for this on travis?

pimterry:
@thgreasi nope. For better or worse, travis uses the same env vars for all environments. That's why it's slower, has it has to run them sequentially, rather than in parallel. That means this will make the builds take longer, but only until #417 is in, so I'm not too worried.

It's a good point though anyway, we should also refactor that to tie separate users to each environment, so we can have separate builds in parallel.

PR #417, opened 2017-10-10
*BREAKING*: Stop actively supporting node 4.

pimterry:
Another breaking change for the SDK v7 branch.

Connects to #415, and comes on top of #416, which adds basic node 8 tests to travis.

### DIFF
--- a/package.json
+++ b/package.json
@@ -35,6 +35,9 @@
   },
   "author": "Juan Cruz Viotti <juan@resin.io>",
   "license": "Apache-2.0",
+  "engines": {
+    "node": ">=6.0"
+  },
   "devDependencies": {
     "browserify": "^14.3.0",
     "catch-uncommitted": "^1.0.0",
--- a/.travis.yml
+++ b/.travis.yml
@@ -12,8 +12,6 @@ matrix:
       - '6'
       env:
       - 'CAN_DEPLOY=true'
-    - node_js:
-      - '4'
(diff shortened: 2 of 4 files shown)


package.json
```diff
@@ -35,6 +35,9 @@
   },
   "author": "Juan Cruz Viotti <juan@resin.io>",
   "license": "Apache-2.0",
+  "engines": {
+    "node": ">=6.0"
```
thgreasi: So it looks like that this enforces node v6+ which is a bit different than `Stop actively supporting node 4`.

pimterry: This isn't enforced by anybody (as far as I know), it's [advisory only](https://docs.npmjs.com/files/package.json#engines). It makes npm print a warning if you install it using an unsupported the wrong node version, but no more.

pimterry:
@thgreasi I'm going to merge because I think it's ok, and you've marked this as approved.

I feel like there's been quite a few times recently where you've put comments, but marked the PR as approved too. As long as you really are happy with things being merged as is, that's totally fine, but if you're not sure whether something's good to merge you should definitely feel free to comment/request changes on PRs instead!

thgreasi:
I was fine with it, just wanted to clarify that point. Regardless of the answer it seems fine and that's why I :+1:  upfront.

pimterry:
Cool cool cool, just checking! :+1:

PR #418, opened 2017-10-11
Move to API v3

pimterry:
Woo hoo - this moves the SDK to the new pine 5/api v3, updating all the relationships along the way, and hopefully finishing up 7.0.0. This has been made easier by the previous PRs removing lots of relationship logic entirely, but there's still quite a few left. 

Note that while this doesn't break an enormous amount of code internally, it will break a huge amount of user code - anything that tries to follow relationships between models is going to have to be updated.

I'd definitely like your review on this to confirm before I merge @Page-, just in case I've missed something obvious.

Once this is done, I'm going to merge this into the v7 branch, do a last little bit of manual testing to double check things, and then I'm going to open a PR to master to look at, and get versionbot to do a 7.0.0 there release with everything in.

### DIFF
--- a/README.md
+++ b/README.md
@@ -64,7 +64,6 @@ var resin = require('resin-sdk')({
 
 Where the factory method accepts the following options:
 * `apiUrl`, string, *optional*, is the resin.io API url. Defaults to `https://api.resin.io/`,
-* `apiVersion`, string, *optional*, is the version of the API to talk to, like `v2`. Defaults to the current stable version: `v2`,
 * `apiKey`, string, *optional*, is the API key to make the requests with,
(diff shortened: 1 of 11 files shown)


thgreasi:
Woohoo!

README.md
```diff
@@ -64,7 +64,6 @@ var resin = require('resin-sdk')({
 
 Where the factory method accepts the following options:
 * `apiUrl`, string, *optional*, is the resin.io API url. Defaults to `https://api.resin.io/`,
-* `apiVersion`, string, *optional*, is the version of the API to talk to, like `v2`. Defaults to the current stable version: `v2`,
```
Page-: Yeah, removing this makes a lot of sense

PR #420, opened 2017-10-12
Release Resin-SDK v7.0.0

pimterry:
This includes lots of breaking changes, notably including the upgrade to the new API v3.

All these changes have previously been reviewed, this PR is just merging them all together and using versionbot to actually ship them to npm.

For posterity, this PR includes:
* #398 
* #407 
* #410 (plus some changes after rebasing over #419)
* #417 
* #418 
* a couple of last extra changes and bugfixes on the end 
---- Autogenerated Waffleboard Connection: Connects to #358


thgreasi:
Woohoo! I'm on it!

pimterry:
@thgreasi I've been doing my own review of the codebase - I think I've actually found a couple of bugs in here in the untested device methods (things that do stuff with real devices, like shutdown/reboot) from the combination of changing the relationships and removing the expands. I'm going to do some quick testing to confirm, fix them, and then update this PR with the results soon.

lib/models/device.coffee
```diff
@@ -421,7 +428,7 @@ getDeviceModel = (deps, opts) ->
 	exports.getApplicationInfo = (uuidOrId, callback) ->
 		exports.get(uuidOrId).then (device) ->
 			ensureSupervisorCompatibility(device.supervisor_version, MIN_SUPERVISOR_APPS_API).then ->
-				appId = device.application[0].id
+				appId = device.belongs_to__application[0].id
```
thgreasi: I think that this needs to change since we are not expanding the `belongs_to__application`.
I guess that we should either expand it or use `device.belongs_to__application.__id`.

lib/models/device.coffee
```diff
@@ -812,7 +819,7 @@ getDeviceModel = (deps, opts) ->
 	exports.startApplication = (uuidOrId, callback) ->
 		exports.get(uuidOrId).then (device) ->
 			ensureSupervisorCompatibility(device.supervisor_version, MIN_SUPERVISOR_APPS_API).then ->
-				appId = device.application[0].id
+				appId = device.belongs_to__application[0].id
```
thgreasi: Like above
> I think that this needs to change since we are not expanding the `belongs_to__application`.
I guess that we should either expand it or use `device.belongs_to__application.__id`.

lib/models/device.coffee
```diff
@@ -855,7 +862,7 @@ getDeviceModel = (deps, opts) ->
 	exports.stopApplication = (uuidOrId, callback) ->
 		exports.get(uuidOrId).then (device) ->
 			ensureSupervisorCompatibility(device.supervisor_version, MIN_SUPERVISOR_APPS_API).then ->
-				appId = device.application[0].id
+				appId = device.belongs_to__application[0].id
```
thgreasi: Like above
> I think that this needs to change since we are not expanding the `belongs_to__application`.
I guess that we should either expand it or use `device.belongs_to__application.__id`.

lib/models/device.coffee
```diff
@@ -982,7 +989,7 @@ getDeviceModel = (deps, opts) ->
 				baseUrl: apiUrl
 				body:
 					deviceId: device.id
-					appId: device.application[0].id
+					appId: device.belongs_to__application[0].id
```
thgreasi: Like above

lib/models/device.coffee
```diff
@@ -1024,9 +1031,9 @@ getDeviceModel = (deps, opts) ->
...
 					data:
-						appId: device.application[0].id
+						appId: device.belongs_to__application[0].id
```
thgreasi: Like above twice

lib/models/device.coffee
```diff
@@ -1071,7 +1078,7 @@ getDeviceModel = (deps, opts) ->
 				baseUrl: apiUrl
 				body:
 					deviceId: device.id
-					appId: device.application[0].id
+					appId: device.belongs_to__application[0].id
```
thgreasi: Like above

pimterry:
@thgreasi See the last 4 commits - I've been through and manually tested every untested method in the devices file with a real pi. I've fixed the few that were broken by missing relationships, plus fixing a couple of preexisting bugs and mess in some of the others.

I've also stopped exposing the weird `ensureSupervisorCompatibility` method (though we still use it internally), which just did a semver comparison and and then threw an exception, and didn't really know anything about supervisor versions at all.

Rereview please!

lib/models/device.coffee
```diff
@@ -1560,12 +1583,16 @@ getDeviceModel = (deps, opts) ->
...
 	###
 	exports.disableTcpPing = (uuidOrId, callback) ->
-		exports.get(uuidOrId).then (device) ->
+		exports.get uuidOrId,
+			select: 'id'
+			expand: belongs_to__application: $select: 'id'
+		.then (device) ->
 			return request.send
-				method: 'DELETE'
+				method: 'POST'
```
thgreasi: I guess that this changed at some point

lib/models/device.coffee
```diff
@@ -1070,8 +1085,13 @@ getDeviceModel = (deps, opts) ->
 	# 	if (error) throw error;
 	# });
 	###
-	exports.update = (uuidOrId, options, callback) ->
-		exports.get(uuidOrId).then (device) ->
+	exports.update = (uuidOrId, options = {}, callback) ->
```
thgreasi: What happens when updating with an empty object? Does it serve like a ping?

lib/models/device.coffee
```diff
@@ -101,7 +101,7 @@ getDeviceModel = (deps, opts) ->
 	# 	console.log('Is compatible');
 	# });
 	###
-	exports.ensureSupervisorCompatibility = ensureSupervisorCompatibility = Promise.method (version, minVersion) ->
+	ensureSupervisorCompatibility = Promise.method (version, minVersion) ->
```
thgreasi: It seems that we don't use it in the dashboard, so :+1:

lib/models/device.coffee
pimterry: As far as I can tell, it's needed this [since basically forever](https://github.com/resin-io/resin-api/commit/2e7c2896c82fbc530172c27b3774e9c501a70fa2), so it's never worked... The API supervisor proxy requires a POST, plus the real method to use in the body.

Horia fixed the same issue in `ping` recently. The problem is we just don't have automated tests for this stuff right now, so any useful but rarely used methods like this that require real devices can be silently broken :cry:

lib/models/device.coffee
pimterry: Update options are actually only allowed to be `force: true/false` (or nothing). All this method does is ping a device to get it to check for an update. I'm not 100% sure why that's useful, but it's a feature the supervisor supports, so I'm happy to leave it here for now.

pimterry:
Here we go!

Issue #422, opened 2017-10-16
models.os.getMaxSatisfyingVersion cannot return a dev image


Page-:
This is due to `semver.maxSatisfying(['2.0.8+rev1.prod','2.0.8+rev1.dev'], '2.0.8+rev1.dev')` returning `'2.0.8+rev1.prod'`, so even with maximum specificity you still get given the prod image (presumably because `p` > `d` and build metadata is ignored for comparsions)

pimterry:
Fixed and released in 7.0.1

PR #423, opened 2017-10-17
Ensure os.getMaxSatisfyingVersion returns exact results for exact matches

pimterry:
Fix for #422. This is a quick patch so we can get the production issue fixed in resin-cli - in time we should do this more thoroughly and carefully in resin-semver as part of fully integrating that (#390). The tests can stay in here regardless, which should provide an easy framework for those changes later.

Change-Type: patch


pimterry:
(Merging & releasing issues are caused by github status problems: https://status.github.com/messages. I'll have another go in a bit)

Issue #411, opened 2017-10-03
Add support for multicontainer (API v4)


pimterry:
Still an open question on how labels will work here, not sure what the API model looks like yet.

## Key ideas

* Conceptual model should fit with the way the dashboard wants to present things (and the CLI, though that should be very easy), but keep flexibility to explore the raw underlying model
  * Dashboard is likely to be fairly representative of other use cases
  * Consistent terminology around the Resin data model between the SDK & UI, so it's easy to get started with the SDK and know what's going on
* Broadly stick with the same structure and design, keep as many existing methods as possible
	* Largely the same models
	* Every method should still take UUIDs or ids transparently (according to string/number input)
	* Every pine-based method should still take an optional set of pine options to include
* Build an API focused on how it'll be used, not just (or not _only_) exposing the data model directly
* Only use expands in non-default methods. Any `.get` or `.getAll` for example should return only one model.
  * Special methods (`device.getWithServiceDetails()`) can combine multiple for specific use cases - all of these have `WithX` in the name, so hopefully users don't use them unless they need X.
  * Don't necessarily need to implement these for all models, if there's no use case, but don't use the names for something else.
  * Still doing wildcard selects by default for usability, but with a few specific exceptions for expensive fields (e.g. build logs)

## Planned changes (WIP)

Every 'id' here will also allow UUIDs, and every `.models.*` method has an implicit optional pine options argument.

Items marked as done are done on the multicontainer branch, not yet merged (all of this will get merged together, since changing to v4 will break everything that isn't updated).

### Replace `sdk.models.build` with `sdk.models.release`

Existing build methods are replaced by:

* [x] `.getAllByApplication(appId)` - get the release model only, for all releases for an app
* [x] `.get(releaseId)` - return the release model only, for a single release
* [x] `.getWithImageDetails(releaseId)` - returns release plus associated images
	* Includes more or less everything except release logs (can `select` those in if you like though)

### `sdk.models.image`:

* [x] `.get(imageId)` - return the details of an image (without logs by default, but you could `select` them in)
* [x] `.getLogs(imageId)` - returns the full build logs for an image

### `sdk.models.application`:

Most existing methods remain, returning the same/similar data to v3.

* [ ] `.configVar.getAll(appId)`
* [ ] `.configVar.getAllWithOverrides(appId)` - get app config vars and any device overrides together
* [ ] `.configVar.get(appId, key)`
* [ ] `.configVar.set(appId, key, value)`
[...]

Issue #353, opened 2017-07-21
Convert SDK and all the major submodules to TypeScript


pimterry:

pimterry:
Progress: 3 modules (out of about 7) converted to TS. We've now established the structure and standard config for this, so the others are fairly simple, just copy paste most of the changes from the existing ones (e.g. resin-token).

PR #427, opened 2017-11-02
WIP: First steps towards Multicontainer

pimterry:
It's m-m-m-multicontainer time!

Connects to #411 

Depends on [do we have an API issue for multicontainer?] and notably needs https://github.com/resin-io/resin-api/pull/636 too. 

This PR currently makes the first step: it creates a test that creates a representative (I think - thoughts @CameronDiver?) set of data for a live multicontainer application, and adds a method to query a summary of a device's multicontainer state.

This is clearly going to fail the tests for a while, since I haven't fixed any other methods. Despite that, this does give us a base we can start working with to get essential data into the UI.

I'm going to build out a minimum SDK that passes the tests built around this, get it onto the v8 branch, and open new PRs for further methods and changes as we go. The tests also need tidying up, since the huge amount of code required to build the multicontainer state is pretty ridiculous. Thoughts _very_ welcome in the meantime though!


lib/models/device.coffee
```diff
@@ -301,6 +304,106 @@ getDeviceModel = (deps, opts) ->
...
+				release = install.image[0].is_part_of__release[0]
+				service = install.image[0].is_a_build_of__service[0]
+
+				return Object.assign {}, omit(install, 'image'),
+					service_name: service.service_name
+					commit: release.commit
+
+			current_release = rawData.should_be_running__release[0]
+
+			# Strip expanded fields (we reformat and readd them below)
```
thgreasi: > read**d**

tests/integration/models/device.spec.coffee
```diff
@@ -753,6 +753,201 @@ describe 'Device Model', ->
 				.then ({ is_accessible_by_support_until__date }) ->
 					m.chai.expect(is_accessible_by_support_until__date).to.be.null
 
+	describe 'given a multicontainer application with a single offline device', ->
+
+		beforeEach ->
+			resin.models.application.create('FooBar', 'raspberry-pi').then (application) =>
```
thgreasi: Would it make sense to make it a facade (with fewer or no arguments) that is part of the `setup.coffee` ?

tests/integration/models/device.spec.coffee
```diff
@@ -753,6 +753,208 @@ describe 'Device Model', ->
...
+
+			it 'should retrieve the current service details', ->
+				resin.models.device.getWithServiceDetails(@device.id).then (deviceDetails) =>
+					m.chai.expect(deviceDetails).to.deep.match
+						device_name: @device.device_name
+						uuid: @device.uuid
+						current_release: @currentRelease.commit
+						current_services:
+							web: [
+								commit: 'new-release-commit'
```
LucianBuzzo: It would be really handy to have the service install id retrieved here, so that we can key the container elements correctly in the UI.

pimterry: Do you mean the service install or the image install? Service installs should, I think, be unique by device + name, so this should be sufficient already. Is there an annoying special case somewhere where that's not true?

LucianBuzzo: ah you're right, I need the image install id, because you can have multiple image installs on a device with the same device+name combination.

tests/integration/models/device.spec.coffee
pimterry: I might do this later, but I'm not going to for now, since it depends a little on how the rest of the multicontainer code works out.

lib/models/device.coffee
```diff
@@ -301,6 +304,106 @@ getDeviceModel = (deps, opts) ->
...
+			device = omit(rawData, [
+				'image_install'
+				'should_be_running__release'
+			])
+
+			device.current_release = current_release.commit
```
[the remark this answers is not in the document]
pimterry: Excellent spot - fixed!

lib/models/device.coffee
pimterry: `re-add`, actually, but you're right, fixed

lib/models/device.coffee
```diff
@@ -301,6 +304,106 @@ getDeviceModel = (deps, opts) ->
...
+								$select: ['id']
+								$expand:
+									is_part_of__release:
+										$select: ['id', 'commit']
+									is_a_build_of__service:
+										$select: ['id', 'service_name']
+			, options
+		.then (rawData) ->
+			containers = rawData.image_install.map (install) ->
+				release = install.image[0].is_part_of__release[0]
```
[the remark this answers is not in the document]
pimterry: Good spot - the model's now been updated to support `install.is_provided_by__release` directly, so we're using that instead

lib/models/device.coffee
```diff
@@ -301,6 +304,106 @@ getDeviceModel = (deps, opts) ->
...
+
+					image_install:
+						$select: [
+							'id'
+							'download_progress'
+							'is_running'
+							'install_date'
+						]
+						$expand:
+							'image':
```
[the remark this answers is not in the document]
pimterry: Done

lib/models/device.coffee
```diff
@@ -301,6 +304,106 @@ getDeviceModel = (deps, opts) ->
 		.asCallback(callback)
 
 	###*
+	# @summary Get a single device together with its associated service's essential details
```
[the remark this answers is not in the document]
pimterry: Agreed, done

tests/integration/models/device.spec.coffee
pimterry: Bonus: this is actually already there conveniently (as `id`, in each of these objects), the tests just didn't check for it. The tests now do check for it, since you're right that's totally useful, and you should be able to start using it immediately regardless.

LucianBuzzo:
@pimterry Why did VB just close this PR?

pimterry:
It'll be because the v8 branch just got merged (huzzah!)

The multicontainer branch still exists, so that should keep working fine in the UI. I'll need to reopen this PR separately, probably against master.

PR #431, opened 2017-11-14
Change the getDashboardUrl to use the device UUID

### LINKED ISSUE #425: Device.getDashboardUrl should support UUIDs

thgreasi:
Resolves #425
Change-Type: major

### DIFF
--- a/lib/models/device.coffee
+++ b/lib/models/device.coffee
@@ -17,7 +17,6 @@ limitations under the License.
 url = require('url')
 Promise = require('bluebird')
 isEmpty = require('lodash/isEmpty')
-isFinite = require('lodash/isFinite')
 once = require('lodash/once')
 without = require('lodash/without')
 find = require('lodash/find')
@@ -110,21 +109,19 @@ getDeviceModel = (deps, opts) ->
 	# @function getDashboardUrl
 	# @memberof resin.models.device
 	#
-	# @param {Object} options - options
-	# @param {Number} options.appId - Application id
-	# @param {Number} options.deviceId - Device id
+	# @param {String} uuid - Device uuid
 	#
 	# @returns {String} - Dashboard URL for the specific device
-	# @throws Exception if either appId or deviceId are empty
+	# @throws Exception if the uuid is empty
 	#
 	# @example
-	# dashboardDeviceUrl = resin.models.device.getDashboardUrl({ appId: 123, deviceId: 456 })
+	# dashboardDeviceUrl = resin.models.device.getDashboardUrl('a44b544b8cc24d11b036c659dfeaccd8')
 	###
-	exports.getDashboardUrl = getDashboardUrl = (options = {}) ->
-		for key in [ 'appId', 'deviceId' ]
-			throw new Error("#{key} option is not a finite number") if not isFinite(options[key])
+	exports.getDashboardUrl = getDashboardUrl = (uuid) ->
+		if typeof uuid != 'string' || isEmpty(uuid)
+			throw new Error('The uuid option should be a non empty string')
 
-		return url.resolve(dashboardUrl, "/apps/#{options.appId}/devices/#{options.deviceId}/summary")
+		return url.resolve(dashboardUrl, "/devices/#{uuid}/summary")
 
 	addExtraInfo = (device) ->
 		normalizeDeviceOsVersion(device)
--- a/tests/integration/models/device.spec.coffee
+++ b/tests/integration/models/device.spec.coffee
@@ -3,7 +3,7 @@ m = require('mochainon')
 superagent = require('superagent')
 Promise = require('bluebird')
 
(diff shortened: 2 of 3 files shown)


thgreasi:
Marked as major but I'm not really sure how the changelog and merging works on branches that are intended to create a single major release.
If we change our minds and want to support the old parameters as well (and generate the old schema url) just give me a hint.

lib/models/device.coffee
```diff
@@ -111,20 +110,20 @@ getDeviceModel = (deps, opts) ->
...
-	# @param {Number} options.deviceId - Device id
+	# @param {String} options.deviceUuid - Device uuid
 	#
 	# @returns {String} - Dashboard URL for the specific device
-	# @throws Exception if either appId or deviceId are empty
+	# @throws Exception if the deviceUuid is empty
 	#
 	# @example
-	# dashboardDeviceUrl = resin.models.device.getDashboardUrl({ appId: 123, deviceId: 456 })
+	# dashboardDeviceUrl = resin.models.device.getDashboardUrl({ deviceUuid: 'a44b544b8cc24d11b036c659dfeaccd8' })
```
pimterry: Since it's just one argument, we could make this match all the other methods, and take a uuidOrId parameter.

thgreasi: Do you mean to make it just accept uuid as a parameter?
Adding the Id as well would require to change the method to return a promise, since it will have to make a request to find the uuid in that case.
Is changing the return type to a promise something that we are fine with?

pimterry: That is what I mean, but I hadn't thought about the promise return type part.

I'm on the fence, but I think it's probably still the right choice? Makes it consistent with everything else here, and that's definitely quite valuable from a dev UX point of view. What do you think?

thgreasi: Tbh I don't really like the idea of turning a previously sync function to an async one.
I'm thinking that a user with no under the hood insights could write the following:
```js
sdk.model.device.getAllByApplication(12345)
.map(device => {
    sdk.model.device.getDashboardUrl(device.id)
})
```
and that would do N additional requests for data that are already downloaded.
What do you think?

thgreasi: Moreover, shouldn't we start hiding IDs from our public APIs as well?

pimterry: > Moreover, shouldn't we start hiding IDs from our public APIs as well?

This is a good point. I don't think we want to strip them immediately, but we should start gently moving in that direction. I'm sold: let's go for a single UUID argument, and keep it sync :+1:

tests/integration/models/device.spec.coffee
```diff
@@ -865,6 +865,22 @@ describe 'Device Model', ->
...
+				m.chai.expect(
+					resin.models.device.getDashboardUrl('af1150f1b1734c428fb1606a4cddec6c')
+				).to.equal("#{dashboardUrl}/devices/af1150f1b1734c428fb1606a4cddec6c/summary")
+
+			it 'should throw when a device uuid is not a string', ->
+				m.chai.expect( -> resin.models.device.getDashboardUrl(1234567))
+				.to.throw()
+
+			it 'should throw when a device uuid is not provided', ->
+				m.chai.expect( -> resin.models.device.getDashboardUrl({}))
```
pimterry: I think this should be `getDashboardUrl()`

lib/models/device.coffee
```diff
@@ -110,21 +109,19 @@ getDeviceModel = (deps, opts) ->
...
-	# @param {Number} options.deviceId - Device id
+	# @param {String} uuid - Device uuid
 	#
 	# @returns {String} - Dashboard URL for the specific device
-	# @throws Exception if either appId or deviceId are empty
+	# @throws Exception if the deviceUuid is empty
 	#
 	# @example
-	# dashboardDeviceUrl = resin.models.device.getDashboardUrl({ appId: 123, deviceId: 456 })
+	# dashboardDeviceUrl = resin.models.device.getDashboardUrl({ deviceUuid: 'a44b544b8cc24d11b036c659dfeaccd8' })
```
pimterry: Example needs updating too

thgreasi:
@pimterry How should we move this forward?
Should I just use the merge button?
Is everything fine (change type, labels, etc)?
I think that we also need #435 (that temporarily disabled the getDashboardUrl tests for offline devices) to make the tests pass. Should we cherry-pick it on top of the v8 branch or merge this anyway?

pimterry:
Yes, we should merge, but it would be nicer to merge with the tests passing if we can.

You can just fast-forward `v8` to `master`, rebase this on top, and that should get the tests passing. Don't cherry-pick: it's nicer if we have literally the same commit on master & v8, rather than the same content in two separate commits, and since v8 doesn't have anything else on it yet there's no downside.

Once that's done, you can just merge manually.

thgreasi:
OK great, will do that.

PR #433, opened 2017-11-16
Add Application & Device Tag methods

### LINKED ISSUE #430: Add tag related methods
Just adding this so we can track the respective PR.

HQ: https://github.com/resin-io/hq/pull/281

thgreasi:
Resolves #430
Connects-To: #430
Change-Type: minor
See: https://www.flowdock.com/app/rulemotion/resin-frontend/threads/5LLJwWzFnU0rI4gY_ftT6PpLLKb
See: https://www.flowdock.com/app/rulemotion/r-beginners/threads/MOdDV50nu4a_pLWytJFwqYS-mqX
---- Autogenerated Waffleboard Connection: Connects to #430


thgreasi:
After some research on upserts, it seems that they are not supported when using unique pairs.
Page suggested that the best we can do atm is to POST and then PATCH if a unique constrain violation occurs.
Kept that as a separate commit in any case.

tests/integration/models/device.spec.coffee
```diff
@@ -890,3 +890,220 @@ describe 'Device Model', ->
...
+						.then (tags) ->
+							m.chai.expect(tags).to.have.length(1)
+							m.chai.expect(tags[0].tag_key).to.equal('EDITOR')
+							m.chai.expect(tags[0].value).to.equal('vim')
+
+					afterEach ->
+						Promise.all([
+							resin.models.device.tags.remove(@device.uuid, 'EDITOR', 'vim')
+							resin.models.device.tags.remove(@device.uuid, 'LANGUAGE', 'js')
+						])
```
pimterry: Gracias, pero no.

tests/integration/models/device.spec.coffee
```diff
@@ -890,3 +890,220 @@ describe 'Device Model', ->
...
+
+			describe 'resin.models.device.tags.set()', ->
+
+				afterEach ->
+					resin.models.device.tags.remove(@device.id, 'EDITOR')
```
pimterry: Nei takk

tests/integration/models/device.spec.coffee
```diff
@@ -890,3 +890,220 @@ describe 'Device Model', ->
...
+						.then (tags) ->
+							tags = _.sortBy(tags, 'tag_key')
+							m.chai.expect(tags).to.have.length(2)
+							m.chai.expect(tags[0].tag_key).to.equal('EDITOR')
+							m.chai.expect(tags[0].value).to.equal('emacs')
+							m.chai.expect(tags[1].tag_key).to.equal('LANGUAGE')
+							m.chai.expect(tags[1].value).to.equal('js')
+
+					afterEach ->
+						resin.models.device.tags.remove(@device.id, 'LANGUAGE')
```
pimterry: Dim Diolch

tests/integration/models/device.spec.coffee
```diff
@@ -890,3 +890,220 @@ describe 'Device Model', ->
...
+							m.chai.expect(tags).to.have.length(0)
+
+					it 'should be able to remove a tag by device uuid', ->
+						resin.models.device.tags.remove(@device.uuid, @tag.tag_key).then =>
+							resin.models.device.tags.getAllByDevice(@device.id)
+						.then (tags) ->
+							m.chai.expect(tags).to.have.length(0)
+
+				afterEach ->
+					resin.models.device.tags.remove(@device.id, 'EDITOR')
```
pimterry: nem, köszönöm

tests/integration/models/device.spec.coffee
```diff
@@ -890,3 +890,220 @@ describe 'Device Model', ->
...
+					afterEach ->
+						resin.models.device.tags.remove(@device.id, 'LANGUAGE')
+
+			describe 'given an existing tag', ->
+
+				beforeEach ->
+					resin.models.device.tags.set(@device.id, 'EDITOR', 'vim').then (tag) =>
+						@tag = tag
+
+				describe 'resin.models.device.tags.remove()', ->
```
pimterry: In all these tests, there's a lot of duplication. Can we reduce that, to simplify & shorten these a bit?

Doesn't need to be one function for the whole lot, especially when there's clearly awkward special cases - we can have a few methods like `itShouldAllowRemovable(resin.models.device.tags)` or something, to run the `.remove` tests for the various cases, and then end up just composing a few reusable functions like that together, plus the extra cases, to get this same test suite?

It's pretty noisy already, and it's only going to get worse once we add in release tags too. I'm not totally sure what this would look like, but it'd be useful if you could play around a little here and see if there are any simple routes to improve it.

lib/models/tags.coffee
```diff
@@ -0,0 +1,106 @@
...
+					name: 'ResinRequestError'
+					body: 'Unique key constraint violated'
+				}
+				->
+					pine.patch
+						resource: "#{associatedResource}_tag"
+						options:
+							filter:
+								"#{associatedResource}": resourceId
+								tag_key: tagKey
```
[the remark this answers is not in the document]
pimterry: Actually, I guess the unique constraint violation probably guarantees we don't have a problem here?

lib/models/tags.coffee
```diff
@@ -0,0 +1,106 @@
+Promise = require('bluebird')
+isEmpty = require('lodash/isEmpty')
+errors = require('resin-errors')
+
+{
+	isId
+	findCallback
+	mergePineOptions
+} = require('../util')
+
```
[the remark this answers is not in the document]
thgreasi: I actually originally placed this in utils but it then after some time it felt to "core" to be in utils since it felt as a base class factory for all tags as you said.
Yea I will try to put a decent comment on the top. I'm glad we get the same confusions & end up with similar solutions.

lib/models/tags.coffee
```diff
@@ -0,0 +1,106 @@
...
+exports.isValidTagKey = isValidTagKey = (tagKey) ->
+	/^[a-zA-Z_\.]+[a-zA-Z0-9_\.]*$/.exec(tagKey)?
+
+exports.isSystemTagKey = (tagKey) ->
+	/^io\.resin\./.test(tagKey)
+
+exports.tagsModel = (deps, opts, factoryOptions) ->
```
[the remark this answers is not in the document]
thgreasi: I only kept that there for consistency with the rest of the models.
Removing.

lib/models/tags.coffee
```diff
@@ -0,0 +1,106 @@
...
+	{ pine } = deps
+	{
+		associatedResource
+		getResourceId
+		ResourceNotFoundError
+	} = factoryOptions
+
+	exports = {}
+
+	exports.getAllByAssociatedResource = (uniqueParam, options = {}, callback) ->
```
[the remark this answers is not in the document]
thgreasi: Changing that.
It just felt that `device.tags.getAllByDevice` & `application.tags.getAllByApplication` have really similar implementations to not abstract them.

lib/models/tags.coffee
```diff
@@ -0,0 +1,106 @@
...
+		callback = findCallback(arguments)
+
+		getResourceId(uniqueParam).then (resourceId) ->
+			exports.getAll(
+				mergePineOptions
+					filter: "#{associatedResource}": resourceId
+					orderby: 'tag_key asc'
```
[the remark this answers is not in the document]
thgreasi: I would expect getAllByX to be ready to use to populate a grid with a sensible ordering but kept getAll without ordering to be just used as an entry point to the resource and avoiding the cost of sorting when the user does not request a specific ordering.
Will add ordering to getAll, I now don't see any good reason not to.

lib/models/tags.coffee
```diff
@@ -0,0 +1,106 @@
...
+				# If we never checked that the resource actually exists
+				# then we should reject an appropriate error
+				# when the post adds nothing to the DB
+				if isId(uniqueParam) && isEmpty(tag)
+					throw new ResourceNotFoundError(uniqueParam)
+			.catch(
+				{
+					name: 'ResinRequestError'
+					body: 'Unique key constraint violated'
+				}
```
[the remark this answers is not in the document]
thgreasi: That's awesome! Didn't have that part of the utils in mind, thanks for the heads-up!

lib/models/tags.coffee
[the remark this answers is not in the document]
thgreasi: Yea, it fails server side, I will add a test case for that! Nice catch/remark :+1:

tests/integration/models/application.spec.coffee
```diff
@@ -293,3 +293,179 @@ describe 'Application Model', ->
...
+
+				describe 'given two existing tags', ->
+
+					beforeEach ->
+						Promise.all([
+							resin.models.application.tags.set(@application.id, 'EDITOR', 'vim')
+							resin.models.application.tags.set(@application.id, 'LANGUAGE', 'js')
+						])
+						.then (args) =>
+							[@tag, @tag2] = args
```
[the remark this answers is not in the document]
thgreasi: I was thinking that I would use the tag ID or something later in the tests.
Removing.

tests/integration/models/device.spec.coffee
```diff
@@ -890,3 +890,220 @@ describe 'Device Model', ->
...
+
+					it 'should retrieve the tag by device uuid', ->
+						resin.models.device.tags.getAllByDevice(@device.uuid)
+						.then (tags) ->
+							m.chai.expect(tags).to.have.length(1)
+							m.chai.expect(tags[0].tag_key).to.equal('EDITOR')
+							m.chai.expect(tags[0].value).to.equal('vim')
+
+					afterEach ->
+						resin.models.device.tags.remove(@device.uuid, 'EDITOR')
```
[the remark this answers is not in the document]
thgreasi: Trying Greek for the Thessaloniki presentation? :smile:

thgreasi:
@pimterry I love how the new tests look like! Great suggestions :+1:

tests/integration/models/application.spec.coffee
```diff
@@ -226,6 +232,28 @@ describe 'Application Model', ->
...
 
+		describe 'resin.models.application.tags', ->
+
+			tagTestOptions =
+				model: resin.models.application.tags
+				resourceName: 'application'
+				uniquePropertyName: 'app_name'
+
+			beforeEach ->
+				tagTestOptions.resource = @application
```
pimterry: It's a bit clearer if you combine this with the above, rather than building the two halves separately. You lose the scoped variable, but mocha gives you a `this` context for exactly that:

```coffeescript
beforeEach ->
     @tagTestOptions = { ... }

describe '...' ->
    itShouldGetAllTags(@tagTestOptions)
```

tests/integration/models/device.spec.coffee
```diff
@@ -753,6 +759,40 @@ describe 'Device Model', ->
...
+
+			beforeEach ->
+				appTagTestOptions.resource = @application
+				deviceTagTestOptions.resource = @device
+				# used for tag creation during the
+				# device.tags.getAllByApplication() test
+				appTagTestOptions.setTagResource = @device
+
+			describe 'resin.models.device.tags.getAllByApplication()', ->
+				itShouldGetAllTagsByResource(appTagTestOptions)
```
pimterry: Since `appTagTestOptions` only gets used once, here, I'd just inline it

tests/integration/models/tags.spec.coffee
```diff
@@ -0,0 +1,188 @@
...
+_ = require('lodash')
+m = require('mochainon')
+{ titleCase } = require('../../util')
+
+getAllByResourceFactory = (model, resourceName) ->
+	(idOrUniqueParam) ->
+		model["getAllBy#{titleCase(resourceName)}"](idOrUniqueParam)
+
+exports.itShouldGetAllTagsByResource = (opts) ->
+	{ model, resourceName, uniquePropertyName } = opts
```
pimterry: Here (and below), you can just inline this destructuring into the function definition, there's no need for the extra `opts` step:

```coffeescript
exports.abc = ({ model, resourceName, ... }) ->
```

tests/util.coffee
```diff
@@ -1,2 +1,5 @@
 exports.loadEnv = ->
 	require('dotenv').config(silent: true)
+
+exports.titleCase = (str) ->
+	str.replace(/\b(\w)/g, (w) -> w.toUpperCase())
```
pimterry: Does lodash not have a function for this?

tests/integration/models/application.spec.coffee
thgreasi: @pimterry This would error, since I'm using some of the properties to construct sensible titles during the `it()` calls, so calls like `it "should retrieve the tag by #{opts.resourceName} id"` will fail.
I initially tried to define a getter property/method on `tagTestOptions` that would reach `@application`, but obviously didn't manage to get the `this` scope of the `beforeEach` inside the `describe` call.

tests/integration/models/device.spec.coffee
thgreasi: Couldn't make it work since `describe` is called before the `beforeEach` where the `@device` becomes available.
Also tried using getter methods, but since the `this` inside `describe` and `beforeEach` are different, I still needed a higher level object to reference in both of them, so I couldn't inline it.
Would you prefer to make `itShouldGetAllTagsByResource` (and the rest methods) to expect the existence of `@resource` on their scope instead of piggy bugging in the options object?

tests/util.coffee
thgreasi: I think not and that's why I got this from the UI.
There is [startCase](http://devdocs.io/lodash~4/index#startCase) which isn't 100% the same but could work here.

tests/integration/models/application.spec.coffee
pimterry: That completely makes sense :+1:

tests/integration/models/device.spec.coffee
pimterry: :+1: yep, this is also me forgetting that we need to wait before `@device` is available. I definitely think passing options is clearer than having an implicit `this` property, let's just leave this as is.

tests/integration/models/tags.spec.coffee
pimterry: Ah, I see. Passing in options and then mutating them later (and depending on that process) like this is kind of confusing though. It feels like we should be passing a getter that grabs this, but you explained why that's difficult above, and that does make sense. Still feels awkward and a bit annoying, but I can't see an easy improvement.

For now let's just stick with it, but if you have any ideas to make this clearer in future I'd love to see them, as this all uses a few more clever tricks than I'd like right now.

tests/util.coffee
pimterry: Whatever's simplest, it's not a big deal either way.

tests/integration/models/tags.spec.coffee
thgreasi: Cool. Totally agree that it's a bit annoying but seemed to be the less complicated (tried the getter & several other approaches). I will keep that in mind for improvement.

thgreasi: How about changing `resource` & `setTagResource` to `resourceProvider()` & `setTagResourceProvider()`?
It's almost the same, but I think that making it a method makes it a bit clearer that this is not a fixed property but its something that changes.

PR #435, opened 2017-11-23
Temporarily disable the getDeviceUrl test for offline devices

thgreasi:
That's in a temporary measurement to make the CI pass again.
See: https://github.com/resin-io/resin-proxy/issues/134

### DIFF
--- a/tests/integration/models/device.spec.coffee
+++ b/tests/integration/models/device.spec.coffee
@@ -622,7 +622,7 @@ describe 'Device Model', ->
 					promise = resin.models.device.getDeviceUrl(@device.id)
 					m.chai.expect(promise).to.eventually.match(/[a-z0-9]{62}/)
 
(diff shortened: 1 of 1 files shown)


thgreasi:
@pimterry should this be marked in a way that will not add an entry to the changelog?

pimterry:
Yes, sounds sensible. That means use `no-checks` and merge manually.

Also, btw, you've linked the wrong issue I think. The cause of this is actually https://github.com/resin-io/resin-proxy/issues/134

pimterry:
I've filed a bug to fix this later, so we remember: https://github.com/resin-io/resin-sdk/issues/436

pimterry:
Oh, also you need to _not_ add a change type footer. Otherwise even if it doesn't release and update the changelog with this PR, it will get added to the changelog for the next release.

PR #437, opened 2017-11-23
Various small sdk test improvements

pimterry:
This PR:

* Prints browser console logs during tests, so you can tell what's going on
* Removes the USER_ID from the test config

The user id is that main change: with that gone, we can use the same config on multiple environments (if the login is the same). This is going to let us start running some of the tests against the multicontainer environment shortly.


pimterry:
Looks like the failing test is just the one that #435 will fix, should be good once that's merged and this is rebased.

tests/integration/auth.spec.coffee
```diff
@@ -136,6 +137,7 @@ describe 'SDK authentication', ->
 
 		describe 'resin.auth.getUserId()', ->
 
-			it 'should eventually be the user id', ->
-				promise = resin.auth.getUserId()
-				m.chai.expect(promise).to.eventually.equal(credentials.userId)
+			it 'should eventually be a user id', ->
+				resin.auth.getUserId()
+				.then (userId) ->
+					m.chai.expect(userId).to.be.a('number')
```
thgreasi: So, it's a funny realization that this is the only place that `credentials.userId` was used.
I'm fine as it is, but since this is the so famous `userId`, would it make any sense to also cross-check it with a server returned value eg requesting the user with that exact username from the api?
Or at least, how about verifying that it's a positive integer?

pimterry: I think doing a separate pine request to confirm feels a bit like overkill imo. I've added a bonus check to make sure it's a positive integer though.

thgreasi: Woohoo!

PR #438, opened 2017-11-24
Start using Typescript

thgreasi:
Connects-To: #353
Change-Type: patch


gulpfile.coffee
```diff
@@ -54,11 +57,20 @@ gulp.task 'lint', ->
...
+	gulp.src(OPTIONS.files.tsAppFiles)
+		.pipe(tsProject()) .on('error', gutil.log)
+		.pipe(gulp.dest(OPTIONS.directories.build))
+
+gulp.task 'build-node-coffee', ->
 	gulp.src(OPTIONS.files.app)
 		.pipe(coffee(header: true, bare: true)).on('error', gutil.log)
 		.pipe(gulp.dest(OPTIONS.directories.build))
 
+gulp.task 'build-node', ['build-node-ts', 'build-node-coffee']
```
pimterry: In general, I'd prefer to start moving away from gulp entirely - it's not making this any easier, and it's an annoy extra layer (or multiple layers) between you and what's going on, which is fairly annoying to work with imo. You can do all of this change except the mocha bit for example by just swapping `gulp build` in package.json for `tsc && gulp build`, which is way easier, and has a lot fewer moving parts.

thgreasi: Didn't have that in mind and thought that I should
* extend instead of write something that would fragment the build logic in two places
* make the gulp tasks parellelable

Moving some of the gulp tasks to package.json as you suggested.

thgreasi:
@pimterry should we merge this manually with no checks (like in #435) or is a patch version bot release fine?

pimterry:
Yep, merging as normal is fine here.

pimterry:
Sorry, by normal I mean with versionbot, to be clear :smiley:

pimterry:
(Although I've also just started merging https://github.com/resin-io/resin-sdk/pull/437, so you might want to wait a few minutes, or versionbot will get angry)

thgreasi:
I can wait till tomorrow.
I prefer merging SDK PRs in the morning while the API has less load :)
Takes less time retying builds :)

Issue #364, opened 2017-08-14
Add support for user-level API keys to be used instead of JWTs


flesler:
HQ: https://github.com/resin-io/hq/pull/861

Discussions:
1. https://www.flowdock.com/app/rulemotion/resin-tech/threads/Tp6Up7Zu2U9qAnOLfqsqY3bYqv0
2. https://www.flowdock.com/app/rulemotion/resin-frontend/threads/_g5ZBIGqmsKYzk5YG_cDx5B07-h
3. https://www.flowdock.com/app/rulemotion/platform/threads/pUlR_TwVuF5RywMqsgYCxYe-QVL

pimterry:
Just to summarize the current state and plan: the goal here is that as well as the currently available login methods, you should also be able to log in with an api key.

The API work for this is done (ask @flesler if you have any specific questions on the backend part), and there's a replacement for [resin-token](https://github.com/resin-io-modules/resin-token) called [resin-auth](https://github.com/resin-io-modules/resin-auth) that should give you a single interface that will work for both JWTs and api keys.

To do this, we're going to need to:

* Update [resin-request](https://github.com/resin-io-modules/resin-request) to use resin-auth, so that when it sends requests, they work with either the JWT or the API key
* Replace resin-token in the SDK with resin-auth:
    * There are some places that depend on the internal data of the token (https://jwt.io/ can be helpful to see this data for yourself), those will need to change (because API keys contain no data), and should generally all be making requests to our API instead.
    * The rest of this should be a fairly simple translation between the differences in resin-token and resin-auth's API
* Add a `resin.auth.loginWithApiKey` method

flesler:
One thing I'd like to add is. Historically, JWT have been sent on the `Authorization` header while API Keys where included as `apikey` in the querystring (most of the time). With [this PR](https://github.com/resin-io/resin-api/issues/509) that is close to production, due to a security concern, we're moving towards sending both JWTs and API Keys in the same place (`Authorization` header). This actually simplifies a lot the integration of `resin-auth` into `resin-request`. Because you just need to put whatever is in `auth.getKey()` in the header, regardless of what it is

MoranF:
@flesler , @pimterry , the SDK is using ```token.getUserId```, ```token.getUserName```, ```token.getEmail```. How should I use these methods with ```resin-auth``` ?

MoranF:
[...]

PR #439, opened 2017-11-28
Use resin-auth instead of resin-token

MoranF:
Connects-To: #364

Change-Type: major
---- Autogenerated Waffleboard Connection: Connects to #364


lib/2fa.coffee
```diff
@@ -47,7 +47,7 @@ get2fa = (deps, opts) ->
 	# });
 	###
 	exports.isEnabled = (callback) ->
-		token.getProperty('twoFactorRequired').then (twoFactorRequired) ->
+		auth.getProperty('twoFactorRequired').then (twoFactorRequired) ->
```
pimterry: Resin auth doesn't look like it has a `getProperty` method, so it seems unlikely that this works.

I think (@flesler can confirm?) API keys never need 2FA, so I think these methods should have `enabled` return false and `passed` return true, if you're using an api key.

lib/auth.coffee
```diff
@@ -132,31 +141,31 @@ getAuth = (deps, opts) ->
...
-	# @summary Login to Resin.io with a token
-	# @name loginWithToken
+	# @summary Login to Resin.io with a token or api key
+	# @name loginWithAuth
 	# @public
 	# @function
 	# @memberof resin.auth
 	#
-	# @description Login to resin with a session token instead of with credentials.
+	# @description Login to resin with a session token ar api key instead of with credentials.
```
pimterry: `ar` -> `or`

lib/auth.coffee
```diff
@@ -222,7 +231,7 @@ getAuth = (deps, opts) ->
 	# });
 	###
 	exports.getToken = (callback) ->
```
pimterry: I think this should probably become `getKey`, and the docs should change `token` to something like `raw API key or session token`.

lib/auth.coffee
```diff
@@ -280,9 +295,15 @@ getAuth = (deps, opts) ->
...
-			return email
+		request.send
+			method: 'GET'
+			url: '/user/v1/whoami'
+			baseUrl: apiUrl
+		.get('body')
+		.then (body) ->
+			return body.email
+		.catch ->
+			throw new errors.ResinNotLoggedIn()
```
pimterry: We should refactor out an internal function for this code, since we're repeating it a few times. It'd be nice if each of these methods did the same thing, but only needed to contain something like:

```coffeescript
getUserDetails()
.get('email')
.asCallback(callback)
```

pimterry: That would actually be especially nice because then we could add caching inside `getUserDetails`, to make all these methods very cheap after the 1st lookup.

Looking at the code below I think that caching is going to be necessary for this PR, since code like `getAll` gets called very frequently (I think the dashboard polls it every 5 seconds for example), and it now makes twice as many HTTP requests.

lib/models/application.coffee
```diff
@@ -85,14 +85,19 @@ getApplicationModel = (deps, opts) ->
 	exports.getAll = (options = {}, callback) ->
 		callback = findCallback(arguments)
 
-		token.getUserId().then (userId) ->
+		request.send
+			method: 'GET'
+			url: '/user/v1/whoami'
+			baseUrl: apiUrl
+		.get('body')
+		.then (body) ->
```
pimterry: This should reuse the existing code in auth to get the user id, rather than duplicating it here.

As above - we should also avoid making an extra request here, since this is very frequently used code, so we'll need to look at caching for this.

package.json
```diff
@@ -71,15 +71,15 @@
     "lodash": "^4.4.0",
     "memoizee": "^0.4.9",
     "moment": "^2.18.1",
-    "promise-memoize": "^1.2.0",
+		"promise-memoize": "^1.2.0",
+		"resin-auth": "^2.0.0",
```
pimterry: Indentation

tests/integration/auth.spec.coffee
```diff
@@ -30,8 +30,8 @@ describe 'SDK authentication', ->
 		describe 'resin.auth.authenticate()', ->
 
 			it 'should eventually be a valid token given valid credentials', ->
-				resin.auth.authenticate(credentials).then(resin.token.isValid).then (isValid) ->
-					m.chai.expect(isValid).to.be.true
+				resin.auth.authenticate(credentials).then(resin.auth.loginWithToken).then(resin.auth.getToken).then (token) ->
```
pimterry: Nit: This will be a little clearer on multiple lines:

```
resin.auth.authenticate(credentials)
.then(resin.auth.loginWithToken)
.then(resin.auth.getToken)
.then (token) ->
  ...
```

pimterry:
Ah, I see you mentioned the test failures happen on master too, that's odd.

Master should definitely pass. Are you sure you've set valid user details for the tests, as in [the docs](https://github.com/resin-io/resin-sdk#tests)? It was definitely passing yesterday: https://travis-ci.org/resin-io/resin-sdk/builds/307800409

MoranF:
@pimterry I think some of the tests fail also because the changes in made in the API. I will wait for the changes to be deployed and than run the tests again.

pimterry:
Do they pass against staging, or your local environment?

MoranF:
@pimterry , most of them pass against my local environment.

flesler:
@pimterry IMHO `loginWithToken` can remain as it is, but now supporting both. Following the naming I used, both JWTs and API Keys are `Tokens`.

lib/2fa.coffee
flesler: Indeed, it doesn't have it (we need a test then!) and API Keys don't have a test, so it's always `true`

lib/auth.coffee
MoranF: I'm changing it to ```getKey``` but needs to change it also in the UI: https://github.com/resin-io/resin-ui/issues/1582

flesler:
@MoranF why have you closed this?

MoranF:
@flesler , yes. Needs to open it again against the v8 branch.

PR #443, opened 2017-12-14
364 use resin auth

MoranF:
---- Autogenerated Waffleboard Connection: Connects to #364

### DIFF
--- a/DOCUMENTATION.md
+++ b/DOCUMENTATION.md
@@ -13,7 +13,6 @@ If you feel something is missing, not clear or could be improved, please don't h
     * [.interceptors](#resin.interceptors) : <code>Array.&lt;Interceptor&gt;</code>
         * [.Interceptor](#resin.interceptors.Interceptor) : <code>object</code>
     * [.request](#resin.request) : <code>Object</code>
-    * [.token](#resin.token) : <code>Object</code>
     * [.pine](#resin.pine) : <code>Object</code>
     * [.errors](#resin.errors) : <code>Object</code>
     * [.models](#resin.models) : <code>object</code>
@@ -128,7 +127,7 @@ If you feel something is missing, not clear or could be improved, please don't h
         * [.whoami()](#resin.auth.whoami) ⇒ <code>Promise</code>
         * [.authenticate(credentials)](#resin.auth.authenticate) ⇒ <code>Promise</code>
         * [.login(credentials)](#resin.auth.login) ⇒ <code>Promise</code>
-        * [.loginWithToken(token)](#resin.auth.loginWithToken) ⇒ <code>Promise</code>
+        * [.loginWithToken(authToken)](#resin.auth.loginWithToken) ⇒ <code>Promise</code>
         * [.isLoggedIn()](#resin.auth.isLoggedIn) ⇒ <code>Promise</code>
         * [.getToken()](#resin.auth.getToken) ⇒ <code>Promise</code>
         * [.getUserId()](#resin.auth.getUserId) ⇒ <code>Promise</code>
@@ -203,20 +202,6 @@ using the same token and hooks as the SDK.
 **Example**  
 ```js
 resin.request.send({ url: 'http://api.resin.io/ping' });
-```
-<a name="resin.token"></a>
-
-### resin.token : <code>Object</code>
-The resin-token instance used internally. This should not be necessary
-in normal usage, but can be useful if you want to directly get or set
-the auth token that the SDK will use.
-
-**Kind**: static property of <code>[resin](#resin)</code>  
-**Summary**: Resin token instance  
-**Access**: public  
-**Example**  
-```js
-resin.token.set('abcdef...');
 ```
 <a name="resin.pine"></a>
 
@@ -3053,7 +3038,7 @@ resin.models.billing.downloadInvoice('0000').then(function(stream) {
     * [.whoami()](#resin.auth.whoami) ⇒ <code>Promise</code>
     * [.authenticate(credentials)](#resin.auth.authenticate) ⇒ <code>Promise</code>
     * [.login(credentials)](#resin.auth.login) ⇒ <code>Promise</code>
-    * [.loginWithToken(token)](#resin.auth.loginWithToken) ⇒ <code>Promise</code>
+    * [.loginWithToken(authToken)](#resin.auth.loginWithToken) ⇒ <code>Promise</code>
     * [.isLoggedIn()](#resin.auth.isLoggedIn) ⇒ <code>Promise</code>
     * [.getToken()](#resin.auth.getToken) ⇒ <code>Promise</code>
     * [.getUserId()](#resin.auth.getUserId) ⇒ <code>Promise</code>
@@ -3233,24 +3218,24 @@ resin.auth.login(credentials, function(error) {
 ```
 <a name="resin.auth.loginWithToken"></a>
 
-#### auth.loginWithToken(token) ⇒ <code>Promise</code>
-Login to resin with a session token instead of with credentials.
+#### auth.loginWithToken(authToken) ⇒ <code>Promise</code>
+Login to resin with a session token or api key instead of with credentials.
 
 **Kind**: static method of <code>[auth](#resin.auth)</code>  
-**Summary**: Login to Resin.io with a token  
+**Summary**: Login to Resin.io with a token or api key  
 **Access**: public  
 
 | Param | Type | Description |
 | --- | --- | --- |
-| token | <code>String</code> | the auth token |
+| authToken | <code>String</code> | the auth token |
 
 **Example**  
 ```js
-resin.auth.loginWithToken(token);
+resin.auth.loginWithToken(authToken);
 ```
 **Example**  
 ```js
-resin.auth.loginWithToken(token, function(error) {
+resin.auth.loginWithToken(authToken, function(error) {
 	if (error) throw error;
 });
 ```
@@ -3289,9 +3274,9 @@ resin.auth.isLoggedIn(function(error, isLoggedIn) {
 This will only work if you used [module:resin.auth.login](module:resin.auth.login) to log in.
 
 **Kind**: static method of <code>[auth](#resin.auth)</code>  
-**Summary**: Get current logged in user's token  
+**Summary**: Get current logged in user's raw API key or session token  
 **Access**: public  
-**Fulfil**: <code>String</code> - session token  
+**Fulfil**: <code>String</code> - raw API key or session token  
 **Example**  
 ```js
 resin.auth.getToken().then(function(token) {
--- a/lib/2fa.coffee
+++ b/lib/2fa.coffee
@@ -15,7 +15,7 @@ limitations under the License.
 ###
 
 get2fa = (deps, opts) ->
-	{ token, request } = deps
+	{ auth, request } = deps
 	{ apiUrl } = opts
 
 	exports = {}
@@ -47,7 +47,7 @@ get2fa = (deps, opts) ->
 	# });
 	###
 	exports.isEnabled = (callback) ->
-		token.getProperty('twoFactorRequired').then (twoFactorRequired) ->
+		auth.needs2FA().then (twoFactorRequired) ->
 			return twoFactorRequired?
 		.asCallback(callback)
 
@@ -78,7 +78,7 @@ get2fa = (deps, opts) ->
 	# });
 	###
 	exports.isPassed = (callback) ->
-		token.getProperty('twoFactorRequired').then (twoFactorRequired) ->
+		auth.needs2FA().then (twoFactorRequired) ->
 			return not twoFactorRequired
 		.asCallback(callback)
 
@@ -107,7 +107,7 @@ get2fa = (deps, opts) ->
 			baseUrl: apiUrl
 			body: { code }
 		.get('body')
-		.then(token.set)
+		.then(auth.setKey)
 		.asCallback(callback)
 
 	return exports
--- a/lib/auth.coffee
+++ b/lib/auth.coffee
@@ -15,9 +15,10 @@ limitations under the License.
 ###
 
 errors = require('resin-errors')
+Promise = require('bluebird')
 
 getAuth = (deps, opts) ->
-	{ token, request } = deps
+	{ auth, request} = deps
 	{ apiUrl } = opts
 	twoFactor = require('./2fa')(deps, opts)
 
@@ -28,6 +29,21 @@ getAuth = (deps, opts) ->
 	# @memberof resin.auth
 	###
 	exports.twoFactor = twoFactor
+
+	userDetailsCache = null
+	getUserDetails = ->
+		Promise.try ->
+			if userDetailsCache
+				return userDetailsCache
+			return request.send
+				method: 'GET'
+				url: '/user/v1/whoami'
+				baseUrl: apiUrl
+			.get('body')
+			.tap (body) ->
+				userDetailsCache = body
+			.catch ->
+				throw new errors.ResinNotLoggedIn()
 
 	###*
 	# @summary Return current logged in username
@@ -62,7 +78,10 @@ getAuth = (deps, opts) ->
 	# });
 	###
 	exports.whoami = (callback) ->
-		token.getUsername().asCallback(callback)
+		getUserDetails()
+		.then (userDetails) -> userDetails.username || undefined
+		.catchReturn(undefined)
+		.asCallback(callback)
 
 	###*
 	# @summary Authenticate with the server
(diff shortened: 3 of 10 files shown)


MoranF:
@pimterry , it seems that the tests fail because the change I made for ```/whoami``` in the API is not in production yet.

lib/2fa.coffee
```diff
@@ -47,8 +47,14 @@ get2fa = (deps, opts) ->
 	# });
 	###
 	exports.isEnabled = (callback) ->
-		token.getProperty('twoFactorRequired').then (twoFactorRequired) ->
-			return twoFactorRequired?
+		auth.getType().then (tokenType) ->
+			if tokenType != 0
```
pimterry: Typically this should really use the enum type in resin-auth, rather than depending on the specific magic number (which willl change, if somebody reorders these two lines: https://github.com/resin-io-modules/resin-auth/blob/master/lib/token.ts#L31-L32). That'd also let you avoid the comment below, and make this more readable.

In fact though, I don't think you need this check at all. `needs2FA` is implemented for API keys too, but just returns false, so without this check I think this code will still do the same thing: https://github.com/resin-io-modules/resin-auth/blob/7a692b306174af5f89cd6979ecc85463f578f576/lib/api-key.ts#L102

lib/2fa.coffee
```diff
@@ -78,7 +84,13 @@ get2fa = (deps, opts) ->
...
 	exports.isPassed = (callback) ->
-		token.getProperty('twoFactorRequired').then (twoFactorRequired) ->
+		auth.getType().then (tokenType) ->
+			if tokenType != 0
+				# API key
+				return true
+			else
+				# JWT
+				return auth.needs2FA().then (twoFactorRequired) ->
 			return not twoFactorRequired
```
pimterry: The indentation here is confusing. This line should be more indented then the one above, right? What's going on here?

In addition, I think you can probably skip this one too, because `needs2FA` already handles this, probably (we should check though, for both these cases)

lib/auth.coffee
```diff
@@ -29,6 +30,23 @@ getAuth = (deps, opts) ->
 	###
 	exports.twoFactor = twoFactor
 
+	userDetails = null
+	exports.getUserDetails = ->
+		return new Promise (resolve, reject) ->
+			if(userDetails)
+				resolve userDetails
```
pimterry: Nit: `if` should have a space, and `resolve` should have brackets (we only skip brackets for multi-line function calls)

lib/auth.coffee
```diff
@@ -29,6 +30,23 @@ getAuth = (deps, opts) ->
...
+				request.send
+					method: 'GET'
+					url: '/user/v1/whoami'
+					baseUrl: apiUrl
+				.get('body')
+				.then (body) ->
+					userDetails = body
+					resolve userDetails
+				.catch ->
+					reject undefined
```
pimterry: Resolve and reject should have brackets here too.

Reject should not be called with undefined though! The catch function will be called with an error - you should pass the error to reject like:

```coffeescript
.catch (e) ->
  reject(e)
```

Conveniently, that can be shortened to just `.catch(reject)`.

Alternatively, maybe you meant to run `resolve(undefined)`? I don't think that's right though.

lib/auth.coffee
```diff
@@ -62,7 +80,12 @@ getAuth = (deps, opts) ->
 	# });
 	###
 	exports.whoami = (callback) ->
-		token.getUsername().asCallback(callback)
+		exports.getUserDetails()
+			.then (body) ->
+				return body.username || undefined
+			.catch ->
+				return undefined
```
pimterry: Nit: Bluebird has lots of lovely things. It takes a little while to get used to all of them, but there's a list here: http://bluebirdjs.com/docs/api-reference.html

Notably, you can replace this with just `.catchReturn(undefined)`

lib/auth.coffee
```diff
@@ -29,6 +30,23 @@ getAuth = (deps, opts) ->
 	###
 	exports.twoFactor = twoFactor
 
+	userDetails = null
+	exports.getUserDetails = ->
```
pimterry: If this is supposed to be a user-accessible function, it should have docs like the others. If it's not, it shouldn't be exported.

lib/auth.coffee
```diff
@@ -29,6 +30,23 @@ getAuth = (deps, opts) ->
...
+		return new Promise (resolve, reject) ->
```
pimterry: Most of the time, you don't really need to run `new Promise` yourself (with Bluebird especially), and there's a nicer answer, especially if you've already got a promise, like you do below.

In this case, you can simplify this to:

```coffeescript
Promise.try ->
  if userDetails
    return userDetails
  else
    return request.send
      ...
    .get('body')
    .then (body) ->
      userDetails = body
      return userDetails
```

`Promise.try` creates a promise, but acts like a normal `.then`. If you `return` then the promise resolves with that value. If you throw an exception, it rejects. If you return a promise, it uses the result of the promise, so you can skip the manual calls to `resolve`/`reject` entirely.

lib/auth.coffee
```diff
@@ -189,40 +212,38 @@ getAuth = (deps, opts) ->
...
-			baseUrl: apiUrl
-		.return(true)
-		.catch ->
-			return false
-		.asCallback(callback)
+		exports.getUserDetails()
+			.then ->
+				return true
+			.catch ->
+				return false
```
pimterry:
```coffeescript
.return(true)
.catchReturn(false)
```

_Magic_

lib/auth.coffee
```diff
@@ -303,7 +328,8 @@ getAuth = (deps, opts) ->
 	# });
 	###
 	exports.logout = (callback) ->
-		token.remove().asCallback(callback)
+		userDetails = null
```
pimterry: We also need to reset this anywhere you log in, or otherwise change the token. If you're logged in, you check the details, you log in as somebody else, and check the details again, you should get the 2nd login's details, not the 1st one's.

lib/models/application.coffee
```diff
@@ -85,21 +87,22 @@ getApplicationModel = (deps, opts) ->
...
-			return pine.get
-				resource: 'application'
-				options:
-					mergePineOptions
-						orderby: 'app_name asc'
-						filter:
-							user: userId
-					, options
+		auth.getUserDetails()
+			.then (body) ->
```
pimterry: Nit: I'd probably call this 'userDetails'

tests/integration/resin.spec.coffee
```diff
@@ -29,28 +29,30 @@ describe 'Resin SDK', ->
...
 
 		afterEach ->
 			resin.interceptors = []
 
-		it "should update if the array is set directly (not only if it's mutated)", ->
-			interceptor = request: m.sinon.mock().returnsArg(0)
-			resin.interceptors = [ interceptor ]
+		givenLoggedInUser()
+
+		getRequestInfo = (fn) ->
```
pimterry: I don't understand this method name. How it this getting request info? Maybe `ignoreWhoamiCalls` or something instead?

tests/integration/auth.spec.coffee
```diff
@@ -30,8 +30,10 @@ describe 'SDK authentication', ->
 		describe 'resin.auth.authenticate()', ->
 
 			it 'should eventually be a valid token given valid credentials', ->
-				resin.auth.authenticate(credentials).then(resin.token.isValid).then (isValid) ->
-					m.chai.expect(isValid).to.be.true
+				resin.auth.authenticate(credentials)
+				.then(resin.auth.loginWithToken)
+				.then(resin.auth.getKey).then (key) ->
```
pimterry: Ultra-nit: I think this `then` should be on the next line, so it matches the others

tests/integration/auth.spec.coffee
```diff
@@ -30,8 +30,10 @@ describe 'SDK authentication', ->
...
+					m.chai.expect(key).to.be.a('string')
```
pimterry: We definitely need some new tests here too. At the very least we need tests that log in with an API key, and checks the basic methods do sensible things.

lib/auth.coffee
thgreasi: I guess that this needs to be exported either way, since it's used in other files.
If we document this, then we should probably reject with a `ResinNotLoggedIn` error, like the rest get<UserDetailProperty> methods do and in that case we should be able to remove the respective `throw`s from the rest of the methods.

lib/auth.coffee
thgreasi: I guess that we are going with the `Promise.try` approach, so we might end up w/o a `.catch` if the error needs to be propagated (which from the rest of the code seems to be the case) or with a `.catchReturn` if it needs to be silenced.

lib/auth.coffee
```diff
@@ -251,10 +272,12 @@ getAuth = (deps, opts) ->
 	# });
 	###
 	exports.getUserId = (callback) ->
-		token.getUserId().then (id) ->
-			throw new errors.ResinNotLoggedIn() if not id?
-			return id
-		.asCallback(callback)
+		exports.getUserDetails()
+			.then (body) ->
```
thgreasi: I would prefer a var name like `userDetails` instead of body, but an even better way to do this could be with bluebird's [`.get('id')`](http://devdocs.io/bluebird/api/get).

lib/auth.coffee
```diff
@@ -251,10 +272,12 @@ getAuth = (deps, opts) ->
...
+				return body.id
+			.catch ->
+				throw new errors.ResinNotLoggedIn()
```
thgreasi: Reminder: If the `getUserDetails()` method starts returning a `ResinNotLoggedIn` error, then this can be removed and you could end up with a method as simple as `exports.getUserDetails().get('id')`.

lib/auth.coffee
thgreasi: @pimterry Would it also make sense to have a timeframe out of which the cache should be invalidated anyway? Eg: 10 minutes?

lib/auth.coffee
```diff
@@ -29,6 +30,23 @@ getAuth = (deps, opts) ->
 	###
 	exports.twoFactor = twoFactor
 
+	userDetails = null
+	exports.getUserDetails = ->
+		return new Promise (resolve, reject) ->
+			if(userDetails)
```
thgreasi: Needs a space after the `if` and also could drop the parentheses.

lib/auth.coffee
```diff
@@ -62,7 +80,12 @@ getAuth = (deps, opts) ->
 	# });
 	###
 	exports.whoami = (callback) ->
-		token.getUsername().asCallback(callback)
+		exports.getUserDetails()
+			.then (body) ->
```
thgreasi: NIT: Most chained method calls in the SDK have the same indentation as their previous statement. Please reduce the indentation by 1 level.
So ideally this should be
```
exports.getUserDetails()
.then (body) ->
```
In other words, regardless of whether the `.then` is on top or in a new line, the indentation of the code of the inner function ends up to be the same.

lib/auth.coffee
thgreasi: > NIT: Most chained method calls in the SDK have the same indentation as their previous statement. Please reduce the indentation by 1 level.

lib/auth.coffee
thgreasi: NIT: Most chained method calls in the SDK have the same indentation as their previous statement. Please reduce the indentation by 1 level.

lib/auth.coffee
```diff
@@ -280,10 +303,12 @@ getAuth = (deps, opts) ->
 	# });
 	###
 	exports.getEmail = (callback) ->
-		token.getEmail().then (email) ->
-			throw new errors.ResinNotLoggedIn() if not email?
-			return email
-		.asCallback(callback)
+		exports.getUserDetails()
+			.then (body) ->
```
thgreasi: > NIT: Most chained method calls in the SDK have the same indentation as their previous statement. Please reduce the indentation by 1 level.

lib/models/application.coffee
thgreasi: > NIT: Most chained method calls in the SDK have the same indentation as their previous statement. Please reduce the indentation by 1 level.

and also as already said, `body` isn't that descriptive for this context.

lib/auth.coffee
pimterry: Ahh, I didn't notice it was used in other files.

That's actually easier though: it looks like it's only used in `application`, to get the user's id. If that is the reason, then imo really we should not export this at all; we already export `getUserId`, so we should just use that instead.

lib/auth.coffee
```diff
@@ -280,9 +295,10 @@ getAuth = (deps, opts) ->
 	# });
 	###
 	exports.getEmail = (callback) ->
-		token.getEmail().then (email) ->
-			throw new errors.ResinNotLoggedIn() if not email?
-			return email
+		getUserDetails()
+		.get ('email')
+		.catch ->
+			throw new errors.ResinNotLoggedIn()
```
pimterry: Why is this catch needed here? It doesn't match getUserId or whoami, and `getUserDetails` already throws this error.

lib/models/application.coffee
```diff
@@ -85,21 +87,22 @@ getApplicationModel = (deps, opts) ->
...
 
-		.map (application) ->
-			normalizeApplication(application)
-			return application
+			.map (application) ->
+				normalizeApplication(application)
+				return application
 
-		.asCallback(callback)
+			.asCallback(callback)
```
pimterry: Nit: I normally wouldn't indent the `then`/`map`/`asCallback` here - I'd leave them lining up with the start of `auth.getUserId()`

tests/integration/auth.spec.coffee
```diff
@@ -29,9 +29,29 @@ describe 'SDK authentication', ->
 
 		describe 'resin.auth.authenticate()', ->
 
+			it 'should eventually be a valid api key given valid credentials', ->
+				resin.auth.authenticate(credentials)
+        .then(resin.auth.loginWithToken)
+        .then ->
```
pimterry: Indentation's gone odd here - I think you've got spaces instead of tabs

tests/integration/auth.spec.coffee
```diff
@@ -139,3 +159,38 @@ describe 'SDK authentication', ->
 			it 'should eventually be the user id', ->
 				promise = resin.auth.getUserId()
 				m.chai.expect(promise).to.eventually.equal(credentials.userId)
+
+	describe 'when logged in with API key', ->
+
+		givenLoggedInUserWithApiKey()
+
+		describe 'resin.auth.isLogged()', ->
```
pimterry: Typo, this should be `isLoggedIn`

DOCUMENTATION.md
```diff
@@ -128,9 +127,9 @@ If you feel something is missing, not clear or could be improved, please don't h
         * [.whoami()](#resin.auth.whoami) ⇒ <code>Promise</code>
         * [.authenticate(credentials)](#resin.auth.authenticate) ⇒ <code>Promise</code>
         * [.login(credentials)](#resin.auth.login) ⇒ <code>Promise</code>
-        * [.loginWithToken(token)](#resin.auth.loginWithToken) ⇒ <code>Promise</code>
+        * [.loginWithToken(authToken)](#resin.auth.loginWithToken) ⇒ <code>Promise</code>
         * [.isLoggedIn()](#resin.auth.isLoggedIn) ⇒ <code>Promise</code>
-        * [.getToken()](#resin.auth.getToken) ⇒ <code>Promise</code>
+        * [.getKey()](#resin.auth.getKey) ⇒ <code>Promise</code>
```
pimterry: Sorry, I know I've flip-flopped on this a bit, but I think this should stay as `.getToken`.

Before I wanted to change this and `loginWithToken` to `getKey`/`loginWithKey`, but Ariel pointed out that's the wrong terminology on the previous PR (https://github.com/resin-io/resin-sdk/pull/439#issuecomment-347612074). We should keep the two consistent I think.

pimterry:
One last point, if you do merge while I'm out, remember to merge into `v8` manually, without versionbot, using the `no-checks` label. If you don't, you'll do an npm release when merging into the v8 branch. We want to use versionbot & do the npm release only on the final merge into master.

tests/integration/auth.spec.coffee
```diff
@@ -109,7 +129,42 @@ describe 'SDK authentication', ->
 
 		givenLoggedInUser()
 
-		describe 'resin.auth.isLogged()', ->
+		describe 'resin.auth.isLoggedIn()', ->
```
thgreasi: NIT: Since these are essentialy the exact same test, could we abstract them like this:
```coffee
describe 'when logged in', ->

	itShouldHaveFulfillingAuthMethods = ->
		/* all the describe calls here */

	describe 'using credentials'

		givenLoggedInUser()

		itShouldHaveFulfillingAuthMethods()

	describe 'using API key', ->

		givenLoggedInUserWithApiKey()

		itShouldHaveFulfillingAuthMethods()
```

Otherwise just change the pre-existing `describe 'when logged in', ->` to something like `describe 'when logged in with credentials', ->` so that we can distinguish the two easily.

tests/integration/auth.spec.coffee
```diff
@@ -109,7 +129,42 @@ describe 'SDK authentication', ->
...
+			it 'should eventually be true', ->
+				promise = resin.auth.isLoggedIn()
+				m.chai.expect(promise).to.eventually.be.true
+
+		describe 'resin.auth.logout()', ->
+
+			it 'should logout the user', ->
+				resin.auth.logout().then ->
+					promise = resin.auth.isLoggedIn()
+					m.chai.expect(promise).to.eventually.be.false
```
thgreasi: NIT: I would love to also check the result of `getToken` in here as well, since this isn't tested anywhere atm.

thgreasi:
Just remember to squash before merging!

tests/integration/auth.spec.coffee
MoranF: @thgreasi it's tested on line 32

MoranF: @thgreasi , or did you mean that I should check it after logging out?

thgreasi: What I meant was to check that `getToken` behaves as expected right after a `logout`.

In [line 32](https://github.com/resin-io/resin-sdk/blob/21674ac53527c7f7dacac6510f4b6895af11a11c/tests/integration/auth.spec.coffee#L32-L47)  `getToken` is called after a `loginWithToken()`. The case I was talking about is the supplementary one, to ensure that no token is returned after a user logs out.

thgreasi:
@MoranF can we also change the title of the PR to `Use resin-auth instead of resin-token`?

thgreasi:
Woohoo!

PR #445, opened 2017-12-18
Add types

pimterry:
This pulls the typescript types out of the UI, and into the SDK, as the next step towards #353, and moving the CLI & SDK to TypeScript.

This is mostly just a copy/paste, but with a few tweaks, and a couple of extra closely related things pulled in (e.g. API config, and everything within it, including the device type types), since they're all coming from the SDK, and effectively in the SDK's control.

Next step from here is the CLI I think - I'd like to convert the entire SDK right away, but I'd rather wait until multicontainer is done, and avoid the huge conflict with the existing changes there.

I believe (doing some quick testing linking them together locally) the UI will keep happily working even with this as its custom typings will take priority, but I've also got a PR ready to go to update that to use this, which I can PR as soon as this gets released.

Change-Type: minor


typings/resin-sdk.d.ts
```diff
@@ -0,0 +1,590 @@
...
+
+		belongs_to__application: NavigationResource<Application>;
+		belongs_to__user: NavigationResource<User>;
+		is_managed_by__device: NavigationResource<Device>;
+
+		device_environment_variable: ReverseNavigationResource<
+			DeviceEnvironmentVariable
+		>;
+		manages__device: ReverseNavigationResource<Device>;
+		local_id?: string;
```
thgreasi: Can we move this below `provisioning_progress` so that these props are grouped by primitives/navigation/reverseNavigation? (have that change in my tags PR)

pimterry: Done

pimterry:
Updated to a `minor` change, I'll merge & ship this (and start on the UI side) as soon as all the tests pass.

pimterry:
> Should I push an extra commit on this branch w/ the typings for tags?

@thgreasi up to you. I think this PR is good to go either way, so let's race: if you push tags types onto this branch before the tests pass and I add a merge label, we can do it that way. If you don't, you can just open a new PR with the types separately whenever you like :smiley:

thgreasi:
@pimterry after doing a fresh yarn install and committing the tag changes, the commit hooks also made some extra formatting changes which I don't see in the current branch!?
Do you want me to push them as well?

pimterry:
@thgreasi yes, definitely, that's sensible.

We do have catch-uncommitted in here already though, so this must mean we've got some build steps that the hooks run but `npm run test` doesn't, which isn't great...

thgreasi:
@pimterry that's probably because prettier isn't part of the build step. It only runs manually or when committing a file.

PR #446, opened 2017-12-19
Run prettier on build

pimterry:
From discussion on https://github.com/resin-io/resin-sdk/pull/445#issuecomment-352741660.

With this PR, we'll run prettier in CI, so failing to run it will fail the build, and it'll automatically get run if you run builds or tests locally, which should make it pretty hard to forget to do in the first place. Seems super quick right now. We can reconsider if it becomes a serious slowdown, but for the SDK use case I don't think it matters too much.


pimterry:
We're hitting some new rate limiting problems in #445 - I'm going to resolve them there, merge that, then rebase this on top and these tests should start passing again :crossed_fingers:

PR #447, opened 2017-12-19
Further typing improvements

pimterry:
Previous typing PR #445 doesn't actually work, as we filter the published files in `package.json`, so they're not published.

This PR updates that, and adds a bunch more extra types that previously didn't exist at all, which I've run into as part of starting to convert the CLI (!!!). I've also pinned the typescript version to match the UI, just to make things simpler (I'm not _sure_ but I think I hit some issues there, and making these match seems sensible).

Change-Type: patch


typings/resin-sdk.d.ts
```diff
@@ -547,6 +567,22 @@ declare namespace ResinSdk {
...
+interface SdkOptions {
+	apiUrl?: string;
+	apiKey?: string;
+	imageMakerUrl?: string;
+	dataDirectory?: string;
+	isBrowser?: boolean;
+	debug?: boolean;
+}
+
+interface SdkConstructor {
```
thgreasi: :+1:

typings/resin-sdk.d.ts
```diff
@@ -418,10 +419,23 @@ declare namespace ResinSdk {
 			loginWithToken: (authToken: string) => Promise<void>;
 			logout: () => Promise<void>;
 			getToken: () => Promise<string>;
+			whoami: () => Promise<string>;
```
thgreasi: In the docs this is `Promise<string | undefined>` (not that it would make huge difference).

thgreasi: Should we also add
```
isLoggedIn: () => Promise<boolean>;
getUserId: () => Promise<number>;
getEmail: () => Promise<string>;
```
?

pimterry: Yes to both - done.

Issue #456, opened 2018-01-16
Add an option for user to generate an API key


MoranF:

thgreasi:
Closing this since it was fixed by #457
cc: @MoranF 
PS: since there is no waffle board in this repo you might want to (also) use the `Resolves #issueNumber` GH notation.

PR #457, opened 2018-01-16
Export a method for generating a new API key

MoranF:
Connects-To: #456

Change-Type: minor
---- Autogenerated Waffleboard Connection: Connects to #456


lib/auth.coffee
```diff
@@ -215,6 +215,41 @@ getAuth = (deps, opts) ->
...
+	# 	console.log(apiKey);
+	# });
+	###
+	exports.generateApiKey = (callback) ->
+		request.send
+			method: 'POST'
+			url: '/api-key/user/full'
+			baseUrl: apiUrl
+			body:
+				name: 'apiKey'
```
pimterry: Here you're naming the api key, and this name is what will appear in the UI when the user browses their registered api keys.

We shouldn't use a default `apiKey` name for that - instead this method should take a required name parameter, and we should pass that to the API.

lib/auth.coffee
```diff
@@ -215,6 +215,41 @@ getAuth = (deps, opts) ->
...
 		.asCallback(callback)
 
+	###*
+	# @summary Generate a new API key for user
+	# @name generateApiKey
+	# @public
+	# @function
+	# @memberof resin.auth
+	#
+	# @description This will only work if you user is logged in.
```
pimterry: I'm not sure what exactly this means. Do you mean 'only work if you are logged in with a JWT'? I don't think we need to say that you need to be logged in at all for this method to work - that's true for 90% of methods in the SDK.

Also, the description should also mainly explain what this method does: it registers a new api key for the current user with the name given.

lib/auth.coffee
```diff
@@ -215,6 +215,41 @@ getAuth = (deps, opts) ->
 		.catchReturn(false)
 		.asCallback(callback)
 
+	###*
+	# @summary Generate a new API key for user
+	# @name generateApiKey
```
pimterry: `generate` slightly suggests that this method doesn't have side effects, it just locally generates a key somehow, which isn't quite right.

Can we go for `createApiKey` instead? That's a little clearer, and also matches the `(device/application/etc).create` methods in the rest of the SDK.

tests/integration/auth.spec.coffee
```diff
@@ -33,14 +33,7 @@ describe 'SDK authentication', ->
...
 				.then(resin.auth.loginWithToken)
-				.then ->
-					return resin.request.send
-						method: 'POST'
-						url: '/api-key/user/full'
-						baseUrl: sdkOpts.apiUrl
-						body:
-							name: 'apiKey'
- 				.get('body')
+				.then(resin.auth.generateApiKey)
```
pimterry: Neat! Nice to see existing tests getting more useful and simpler at the same time :+1:

lib/auth.coffee
```diff
@@ -215,6 +215,43 @@ getAuth = (deps, opts) ->
...
+	# @example
+	# resin.auth.createApiKey().then(function(apiKey) {
+	# 	console.log(apiKey);
+	# });
+	#
+	# @example
+	# resin.auth.createApiKey(function(error, apiKey) {
+	# 	if (error) throw error;
+	# 	console.log(apiKey);
+	# });
```
pimterry: These examples need to take an api key name

pimterry:
Ah wait, no, we shouldn't merge this - my comment above about the example still needs fixing. I've removed the label.

Sorry, I put approved as 'tiny fix needed, but then this can get merged without rereview', which was a bit ambiguous. We need to fix that before we can merge (which also means you'll need to force push to remove versionbot's commit here.)

lib/auth.coffee
```diff
@@ -215,6 +216,43 @@ getAuth = (deps, opts) ->
 		.catchReturn(false)
 		.asCallback(callback)
 
+	###*
+	# @summary Creates a new API key for user
```
thgreasi: I think that we need a `the` before the `user`.
How about `Creates a new user API key`?
cc: @pimterry

lib/auth.coffee
```diff
@@ -215,6 +216,43 @@ getAuth = (deps, opts) ->
...
+	###
+	exports.createApiKey = (name, callback) ->
+		request.send
+			method: 'POST'
+			url: '/api-key/user/full'
+			baseUrl: apiUrl
+			body:
+				name: name
+		.get('body')
+		.catch ->
```
thgreasi: How about if the server returns a 500?
Oh, I see that we are doing the same in the rest of the auth methods! We might want to consider narrowing all these `.catch()`es in a separate PR.

pimterry: Issue filed: https://github.com/resin-io/resin-sdk/issues/462

thgreasi:
Woohoo!!!! :raised_hands:

PR #459, opened 2018-01-16
Fix bug when merging expand options

### LINKED ISSUE #458: Can't merge options where expand is an array
Using `mergePineOptions` in the following way

```
var result = mergePineOptions({
  expand: [
    {
      device: {
        $select: ['id']
      }
    }
  ]
}, {
  expand: [
    {
      device: {
        $expand: ['build']
      }
    }
  ]
});
```

gives:

```
{
  expand: {
    '[object Object]': {}
  }
}
```

LucianBuzzo:
Fixes #458

change-type: patch


lib/util/index.coffee
```diff
@@ -187,7 +187,13 @@ convertExpandToObject = (expandOption) ->
 	else if isString(expandOption)
 		return "#{expandOption}": {}
 	else if isArray(expandOption)
-		return fromPairs(expandOption.map((key) -> [key, {}]))
+		return fromPairs(expandOption.map((value) ->
+			if isString(value)
+				return [value, {}]
+			else
+				key = Object.keys(value).shift()
```
pimterry: This gets the first key (with somewhat arbitrary ordering) - I think it should probably be _all_ keys

lib/util/index.coffee
```diff
@@ -187,7 +187,13 @@ convertExpandToObject = (expandOption) ->
 	else if isString(expandOption)
 		return "#{expandOption}": {}
 	else if isArray(expandOption)
-		return fromPairs(expandOption.map((key) -> [key, {}]))
+		return fromPairs(expandOption.map((value) ->
```
thgreasi: NIT: since `value[key]` seems weird, would it make sense to rename `value` to (let's say) `option`?

tests/util.spec.coffee
```diff
@@ -110,6 +110,48 @@ describe 'Pine option merging', ->
...
+	it 'combines $expand params for expand options that are arrays of objects with multiple keys', ->
+		result = mergePineOptions
+			expand: [
+				device: $select: [ 'id' ]
+				application: $expand: [ 'user' ]
+			]
+		,
+			expand: [
+				device: $expand: [ 'build' ]
+				application: $expand: [ 'release' ]
```
thgreasi: Would it make sense to add a test doing `application: $expand: [ 'user' ]` & `application: $expand: 'release'`?

LucianBuzzo: Yeah we could do

PR #460, opened 2018-01-16
Deprecate passing an api key to the SDK constructor

pimterry:
Tiny change to stop publicising this option, now that `loginWithToken` works with api keys, so we can remove it entirely in future. Passing this option instead of logging in properly sends the api key in the query parameter, instead of the Authorization header.

Related discussion: https://www.flowdock.com/app/rulemotion/resin-frontend/threads/NvDyvCz_lL8Hgc4MK5oFsV71bXy


thgreasi:
I hate requesting changes just for comments... :disappointed:

pimterry:
I was trying to think of a way to handle this nicely in the SDK typings, I just gave up in the end, JSDoc is a great suggestion! Done.

I mainly think removing it is better than having it there and marked deprecated because very few (no?) people outside resin should be using it, as until recently it was really hard for them to get an api key in the first place. The only people who are likely to see it at the new people looking at the SDK once they've got an SDK for the dashboard, and I'd like to just hide it from those people completely. Anybody who's currently using it will see the breaking change in the changelog when it gets removed in future anyway.

PR #461, opened 2018-01-16
Never internally select:[], since that not valid

pimterry:
Discussion: https://www.flowdock.com/app/rulemotion/resin-frontend/threads/6vQXtbB9zk3QmcQGqXvTSit9lBG

Change-Type: patch


lib/models/application.coffee
```diff
@@ -249,7 +249,7 @@ getApplicationModel = (deps, opts) ->
 	# });
 	###
 	exports.has = (nameOrId, callback) ->
-		exports.get(nameOrId, select: []).return(true)
+		exports.get(nameOrId, select: ['id']).return(true)
```
Page-: This can also be a string as `select: 'id'` if you want

PR #471, opened 2018-02-08
WIP: Try to improve test stability with a small delay on reset

pimterry:
We've been getting lots and lots of test failures in the last day or two, let's see if this helps

### DIFF
--- a/tests/integration/models/application.spec.coffee
+++ b/tests/integration/models/application.spec.coffee
@@ -261,6 +261,7 @@ describe 'Application Model', ->
 				resin.models.device.register(@application.id, resin.models.device.generateUniqueKey())
 			.then (deviceInfo) =>
 				@deviceInfo = deviceInfo
+			.delay(100)
 
 		describe 'resin.models.application.enableDeviceUrls()', ->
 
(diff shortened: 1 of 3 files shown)


lurch:
Interesting - I guess if this does help then something similar should probably be done in the Python SDK tests too?

pimterry:
Potentially, depends on which fix actually works, I'm still investigating. I suspect there's some race conditions in the API somewhere (this has interesting got much worse since we added logic to the application delete hook: https://github.com/resin-io/resin-api/pull/685), because this shouldn't _really_ be required and the behaviour is very odd (we're creating devices that then don't exist when we try to use them).

If that's true and we chase those down successfully, we shouldn't really need changes in either SDK :crossed_fingers:

pimterry:
Despite adding lots of delays, and seeing some small reliability improvements I think, this still just isn't quite stable. I have however managed to find a repro where a simple loop registering and query devices sees invalid state though: https://www.flowdock.com/app/rulemotion/resin-tech/threads/L1h5i3yoD9wLEKYKqI9k6pmoWxH. That strongly suggests to me that this is an API issue.

pimterry:
Tests seem drastically more stable (completely stable?) now that https://github.com/resin-io/resin-api/issues/743 has hit production. I'm going to close this, we can investigate in future if we have further issues.

PR #474, opened 2018-03-01
Multicontainer apps v2

thgreasi:


lib/models/application.coffee
```diff
@@ -376,34 +376,51 @@ getApplicationModel = (deps, opts) ->
...
 	# });
 	###
-	exports.create = (name, deviceType, parentNameOrId, callback) ->
+	exports.create = ({ name, applicationType, deviceType, parent }, callback) ->
 		callback = findCallback(arguments)
 
-		parentAppPromise = if parentNameOrId
-			exports.get(parentNameOrId, $select: [ 'id' ])
+		applicationTypePromise = if isId(applicationType)
+			Promise.resolve(applicationType)
```
pimterry: Let's simplify a tiny touch: even in our own code we use application types instead of ids, so let's just not support application type ids at all, and drop this case. If you want an app type, use the slug.

typings/resin-sdk.d.ts
```diff
@@ -244,6 +245,19 @@ declare namespace ResinSdk {
 		is_depended_on_by__application: ReverseNavigationResource<Application>;
 	}
 
+	interface ApplicationType {
+		id: number;
+		name: string;
+		slug: string;
+		supports_gateway_mode: true;
+		supports_multicontainer: true;
+		supports_web_url: true;
```
pimterry: These should surely be `boolean`, not `true`.

lib/models/application.coffee
```diff
@@ -376,34 +376,51 @@ getApplicationModel = (deps, opts) ->
 	# @function
 	# @memberof resin.models.application
 	#
-	# @param {String} name - application name
-	# @param {String} deviceType - device type slug
-	# @param {(Number|String)} [parentNameOrId] - parent application name or id
+	# @param {Object} options - application creation parameters
+	# @param {String} options.name - application name
+	# @param {(Number|String)} options.applicationType - application type slug or id
```
pimterry: It'd be nice to include an example slug in here, since people might not know what this is (although I guess in future they will... Useful anyway). `e.g. microservice-starter`

pimterry: Just noticed - this is actually optional in the current API implementation, it'd be nice to have it optional here too.

thgreasi: Didn't know how the API would end up, so I decided to be stricter for start and strictly fit our UI use case. It seems that the API isn't going to change. Will change this :+1:

typings/resin-sdk.d.ts
thgreasi: That's what you get when you write your typings based on retried JSONs :)

lib/models/application.coffee
thgreasi: Isn't the examples section good enough to demonstrate the available slugs?
Regardless, I'm adding an example in the `@param` as well.

thgreasi:
@pimterry if you agree with the changes so far, how about cutting a new beta release?
Should we merge by fast-forwarding the main multicontainer branch?
What's the process on that? Is that something I can do as well (if we agree).

tests/integration/models/application.spec.coffee
```diff
@@ -35,43 +35,87 @@ describe 'Application Model', ->
...
+				.then ->
+					promise = resin.models.application.getAll()
+					m.chai.expect(promise).to.eventually.have.length(1)
+
+			it 'should be rejected if the application type is invalid', ->
+				promise = resin.models.application.create
+					name: 'FooBar'
+					applicationType: 'non-existing'
+					deviceType: 'raspberry-pi'
+				m.chai.expect(promise).to.be.rejectedWith('Invalid application type: foobarbaz')
```
pimterry: Surely this test fails? That's not the right app type in the error here.

thgreasi: You caught me...
I guess we will have to wait for the respective API branch to be deployed on staging in order to verify that all tests indeed pass.

pimterry: You should be able to run the API branch locally yourself in the devenv and test against that pretty easily. Ping me in FD if you want a hand.

pimterry:
Happy for this to get merged (sin VB), as long as the new tests are passing locally for you.

Issue #481, opened 2018-03-28
Upgrade Typescript to 3.0


thgreasi:

thgreasi:
Resolved by #630

Issue #480, opened 2018-03-27
Fully type pine options


thgreasi:
Now with TS 2.8  conditional types and type inference I believe that something like this is at last achievable  w/o the previously hack (`IS`, `DIFF`, `never`, which we never adopted). Already have some code on this and can't wait to open a PR.

PR #482, opened 2018-03-28
Upgrade Typescript to v2.8.1

thgreasi:
Fully types the body of post & patch requests.
Also moves anything non-model related to the pine typings file (& renaming to their original Pine client names)  and re-exporting.

Connects-To: #480
Resolves: #481
---- Autogenerated Waffleboard Connection: Connects to #480
---- Autogenerated Waffleboard Connection: Connects to #481


typings/pinejs-client-core.d.ts
```diff
@@ -1,3 +1,29 @@
...
+
+/**
+ * When not selected-out holds a deferred.
+ * When expanded hold an array with a single element.
+ */
+export type NavigationResource<T = WithId> = T[] | Deferred;
+
+/**
+ * When expanded holds an array, otherwise the property is not present.
+ * Selecting is not suggested,
```
LucianBuzzo: NIT What does 'Selecting is not suggested' mean? Can you be more verbose here?

typings/pinejs-client-core.d.ts
```diff
@@ -1,3 +1,29 @@
...
+export interface WithId {
+	id: number;
+}
+
+export interface Deferred {
+	__id: number;
+}
+
+/**
+ * When not selected-out holds a deferred.
```
LucianBuzzo: NIT What does 'not selected-out' mean? Can you be more verbose here?

thgreasi: Done. Removed the comment from here and updated the one in the re-exported interface found in the SDK, so that we get those helpful description tooltips on hover.

pimterry:
I'm not super super worried about the external users, to be clear. I'm happy ignoring types a little from semver-land (since everybody else does), but we might want to leave a little time anyway, if the upgrading work doesn't slow us down enough anyway. DT doesn't allow types with a new compiler for the first month after release for example, so there's time for people to catch up, and to spot any major issues in the compiler that might block upgrades.

thgreasi:
Oh, perhaps I was a bit too enthusiastic typing even more thing :(
I also opened a UI PR before seeing you comment.

Issue #486, opened 2018-04-05
Extend the sdk's API on named user api keys


thgreasi:

thgreasi:
Resolved by #487

PR #487, opened 2018-04-05
Add methods for listing & revoking named user api keys

thgreasi:
Also changes the createApiKey method to accept an optional description for the key.

Resolves: #486
Change-Type: minor


DOCUMENTATION.md
```diff
@@ -136,7 +136,9 @@ If you feel something is missing, not clear or could be improved, please don't h
         * [.login(credentials)](#resin.auth.login) ⇒ <code>Promise</code>
         * [.loginWithToken(authToken)](#resin.auth.loginWithToken) ⇒ <code>Promise</code>
         * [.isLoggedIn()](#resin.auth.isLoggedIn) ⇒ <code>Promise</code>
-        * [.createApiKey(name)](#resin.auth.createApiKey) ⇒ <code>Promise</code>
+        * [.createApiKey(name, [description])](#resin.auth.createApiKey) ⇒ <code>Promise</code>
+        * [.getAllApiKeys([options])](#resin.auth.getAllApiKeys) ⇒ <code>Promise</code>
+        * [.revokeApiKey(id)](#resin.auth.revokeApiKey) ⇒ <code>Promise</code>
```
thgreasi: @pimterry give me a note if you want to namespace all these methods (but that would be breaking for `createApiKey`).

Also, in the UI we are going to update both the name and the description with the same action.
Would it make sense to provide `renameApiKey` & `noteApiKey` as  methods in here, regardless of how will we implement that in the UI?
PS: would love to hear a naming suggestion for an edit method that updates both fields and doesn't feel awkward.

lib/auth.coffee
```diff
@@ -234,23 +238,102 @@ getAuth = (deps, opts) ->
...
+	#
+	# @example
+	# resin.auth.revokeApiKey(123);
+	#
+	# @example
+	# resin.auth.revokeApiKey(123, function(error) {
+	# 	if (error) throw error;
+	# });
+	###
+	exports.revokeApiKey = (id, callback) ->
```
thgreasi: I ended up using the id here since the `name` field is not unique (natural key) for the api keys!
What I don't like at all is that `createApiKey` returns a string and need to call `getAllApiKeys` to pick the latest key by `created_at` (which arguably doesn't feel too robust given that `name`s are not unique).

DOCUMENTATION.md
pimterry: > give me a note if you want to namespace all these methods (but that would be breaking for createApiKey).

I think we should do this. Fortunately, MC is still beta, so we can! Let's go for it and break them out right here.

> Also, in the UI we are going to update both the name and the description with the same action.

`apiKey.update({ name: 'abc', description: 'def' })` would be fine I think. Where both are optional. For methods that'll be used more frequently externally it might be useful to have slightly more readable names for different calls, but I think for api keys we're the main way people will manage them, and that's fine for now.

lib/auth.coffee
pimterry: Ah, this is awkward. According to the spec, it looks like these should actually be unique: https://github.com/resin-io/hq/pull/861/files#diff-18d34b62d6da8f468178ae8c7359b74fR44

> We won't allow users to repeat names across their API keys, they will act as human-readable identifier within each account

It does seem that you're correct and we haven't enforced that though.

For now let's use an id here anyway. Later, I think we should make names unique (checking with Shaun and somehow migrating in that direction), and then later we can make this method take both strings or numbers, and revoke the name/id accordingly.

lib/models/api-key.ts
```diff
@@ -0,0 +1,234 @@
...
+
+import * as Promise from 'bluebird';
+import isString = require('lodash/isString');
+import pick = require('lodash/pick');
+import * as errors from 'resin-errors';
+import * as ResinSdk from '../../typings/resin-sdk';
+import { InjectedDependenciesParam, InjectedOptionsParam } from '../resin';
+import { findCallback, mergePineOptions } from '../util';
+
+const getApikeysModel = function(deps: InjectedDependenciesParam, opts: InjectedOptionsParam) {
```
pimterry: Nit: should be `getApiKeysModel` (capital K)

pimterry: Also, TypeScript models in the SDK! You hero.

lib/models/api-key.ts
```diff
@@ -0,0 +1,234 @@
...
+	 *
+	 * @example
+	 * resin.models.apiKey.getAll(function(error, apiKeys) {
+	 * 	if (error) throw error;
+	 * 	console.log(apiKeys);
+	 * });
+	 */
+	exports.getAll = function(
+		options: ResinSdk.PineOptionsFor<ResinSdk.ApiKey> = {},
+		callback: (error?: Error, apiKeys?: ResinSdk.ApiKey[]) => any,
```
pimterry: Should be `=> void` I think

lib/models/api-key.ts
```diff
@@ -0,0 +1,234 @@
...
+	 * @example
+	 * resin.models.apiKey.update(123, { name: 'updatedName', description: 'updated description' }, function(error, apiKeys) {
+	 * 	if (error) throw error;
+	 * 	console.log(apiKeys);
+	 * });
+	 */
+	exports.update = function(
+		id: number,
+		apiKeyInfo: { name?: string; description?: string },
+		callback: (error?: Error) => any,
```
pimterry: Ditto

lib/models/api-key.ts
```diff
@@ -0,0 +1,234 @@
...
+	 *
+	 * @example
+	 * resin.models.apiKey.revoke(123);
+	 *
+	 * @example
+	 * resin.models.apiKey.revoke(123, function(error) {
+	 * 	if (error) throw error;
+	 * });
+	 */
+	exports.revoke = function(id: number, callback: (error?: Error) => any): Promise<void> {
```
pimterry: Ditto

lib/util/index.d.ts
```diff
@@ -0,0 +1,17 @@
+import * as ResinSdk from '../../typings/resin-sdk';
+
+/** Use with: `findCallback(arguments)`. */
+export function findCallback(args: IArguments): (() => any);
```
pimterry: Ditto

tests/integration/auth.spec.coffee
```diff
@@ -58,7 +58,12 @@ describe 'SDK authentication', ->
 					email: credentials.username,
 					password: 'NOT-THE-CORRECT-PASSWORD'
 
-				m.chai.expect(promise).to.be.rejectedWith('Unauthorized')
+				m.chai.expect(promise).to.be.rejected
+				.then (error) ->
+					m.chai.expect(error).to.have.property('statusCode')
+					.and.satisfy (statuCode) ->
```
pimterry: Typo: `statuCode` -> `statusCode`

tests/integration/auth.spec.coffee
```diff
@@ -58,7 +58,12 @@ describe 'SDK authentication', ->
...
+						statuCode == 401 or # 'Unauthorized'
+						statuCode == 429 # 'Too many requests in this time frame.'
```
pimterry: I don't think this is right - we shouldn't be hitting rate limits here, and if we are we should fix that properly in general, not just for this special case.

tests/integration/models/api-key.spec.coffee
```diff
@@ -0,0 +1,136 @@
...
+		it 'should be able to create a new api key with description', ->
+			resin.models.apiKey.create('apiKey', 'apiKeyDescription')
+			.then (key) ->
+				m.chai.expect(key).to.be.a('string')
+				resin.models.apiKey.getAll()
+			.then (apiKeys) ->
+				m.chai.expect(apiKeys).to.be.an('array')
+				m.chai.expect(apiKeys).to.have.lengthOf(1)
+				m.chai.expect(apiKeys[0]).to.have.property('name', 'apiKey')
+				m.chai.expect(apiKeys[0]).to.have.property('description', 'apiKeyDescription')
```
pimterry: You could simplify this (and most other assertions here) to just:

```coffee
m.chai.expect(apiKeys).to.deep.equal([
  { name: 'apiKey', description: 'apiKeyDescription' }
])
```

I think that's equivalent to what you've got here, but shorter & clearer.

pimterry: Ah, except presumably you get an id property, and that gets tricky.

There's probably still a nice similar solution with some kind of matching plugin (e.g. https://www.npmjs.com/package/chai-deep-match).

Alternatively, it's a bit of a hack, but I think this should work:

```coffee
m.chai.expect(apiKeys).to.deep.equal([
  { id: apiKeys[0].id, name: 'apiKey', description: 'apiKeyDescription' }
]);
```

Or maybe even

```coffee
m.chai.expect(
  _.map(apiKeys, (k) => _.omit(k, 'id')
).to.deep.equal([
  { name: 'apiKey', description: 'apiKeyDescription' }
])
```

tests/integration/auth.spec.coffee
```diff
@@ -34,7 +34,7 @@ describe 'SDK authentication', ->
 				resin.auth.authenticate(credentials)
```
pimterry: Preexisting, but I don't think the name of this test makes much sense. Looks like it's testing that `loginWithToken` works with api keys? We should move that to the `loginWithToken` tests, and rename it.

tests/integration/setup.coffee
```diff
@@ -73,6 +73,14 @@ exports.resetUser = ->
...
 			resin.pine.delete
 				resource: 'user__has__public_key'
+
+			# logins with api keys don't have
+			# access to the api key resource
+			if loggedInWithApiKey then Promise.resolve()
+			else resin.pine.delete
+				resource: 'api_key'
+				# only delete named user api keys
+				options: $filter: name: $ne: null
```
pimterry: This special case flag is kind of nasty. Easier option, remove that extra noise, and change this to just:

```coffee
resin.pine.delete
	resource: 'api_key'
    # only delete named user api keys
    options: $filter: name: $ne: null
# Api keys can't delete api keys, just ignore failures here
.catchReturn()
```

thgreasi:
> I'm not sure changing resin-requests types is a breaking change, since Bluebird promises should mean we've effectively widened the types of request. Any existing code should still work, I think, it's just that now even more promise methods are available, right?

I actually had to change a method in the UI that was typed to return an es6 Promise to return a Bluebird Promise and that was the reason I added the Breaking comment. Should I remove it?

tests/integration/auth.spec.coffee
thgreasi: Agreed, removing this commit for now. I guess that this only helped since I was using `.only` for this section of the tests. Maybe checking just that the promise was rejected is enough for this "fails to login with wrong password" test case.

tests/integration/models/api-key.spec.coffee
thgreasi: I was thinking about using that package for a long time.
I guess that the time to add it to our dependencies has come.
We definitely need this in the billing tests as well.

tests/integration/auth.spec.coffee
thgreasi: There was no section for `loginWithToken` at all!
Moved the two tests that were using session tokens & api keys to the new `loginWithToken` section.

tests/integration/setup.coffee
thgreasi: I would prefer a way to detect whether we are using a session token or an api key but anyway :)

pimterry:
All looks great! I'm not 100% sure what our position is on breaking changes for types only, which don't actually break at runtime, but I'm fine with being cautious and marking that as breaking as well anyway.

:+1: :shipit:

thgreasi:
The part that I liked in this PR a lot, was that the new TS model uses the sdk's exported typings for `pine` and `request`.

Issue #476, opened 2018-03-16
Calls to models.application show do not include collaborators' apps


nurikabe:
Authing using email/password, calls to `models.application` only return results if the auth user is the original owner of the applications; not a collaborator.

This seems to differ from the REST API.  If I curl using the token of a collaborator I can see all applications for which the collaborator has access.

thgreasi:
Hi,
This seems pretty reasonable. Let me do a research about the reason that made us also filter on the current user ID.

thgreasi:
@pimterry I totally agree that this should be in the MC release.
I think I will not be able to pick this up until next week. Will check back with you when I'll have some time for this.

pimterry:
@CameronDiver is going to handle this as part of some CLI work he's doing

PR #492, opened 2018-04-19
Return collaborator applications with models.applications.getAll

CameronDiver:
Change-type: major
Connects-to: #476
Signed-off-by: Cameron Diver <cameron@resin.io>

### DIFF
--- a/lib/models/application.coffee
+++ b/lib/models/application.coffee
@@ -103,7 +103,11 @@ getApplicationModel = (deps, opts) ->
 					mergePineOptions
 						$orderby: 'app_name asc'
 						$filter:
-							user: userId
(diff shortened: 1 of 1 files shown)


lib/models/application.coffee
```diff
@@ -103,7 +103,11 @@ getApplicationModel = (deps, opts) ->
 					mergePineOptions
 						$orderby: 'app_name asc'
 						$filter:
-							user: userId
+							$or:
```
pimterry: I thought `$or` always took an array? FBM if you've tested it, this is neat.

CameronDiver: Also tested and works fine.

Issue #326, opened 2017-05-16
Find a better way to lock our dependencies


pimterry:
Yarn is nice in principle, but:

* Doesn't help for downstream users, so our environment may not match theirs
* Doesn't currently get used on AppVeyor, which uses npm only, so we get inconsistent test results

If we can, we should find a better way, which lets us test on consistent versions that can then get used by downstream users too.

pimterry:
Closing for now - we've dropped Yarn, we'll stick with that for now.

PR #493, opened 2018-04-19
Remove yarn from the SDK

pimterry:
Opening this because I hit issues whilst trying to do an MC release (if you used `npm install` at all then the current code won't build, because of prettier changes), but some discussion would be good.

We've had a yarn lock here in the SDK for ages, and #326 open almost as long, because it causes problems.

Right now it worries me that we could update a broken runtime subdependency (resin-request etc), and because of Yarn all our tests in the SDK would keep passing, but all downstream users would get an unusuable SDK install.  AFAICT there's no nice way to actually lock the final dependencies our end users are using, so it's better for us to have floating dependencies, and at least that way we'll see the same issues they do.

Thoughts?
---- Autogenerated Waffleboard Connection: Connects to #326


LucianBuzzo:
@pimterry I think this is the best move we can make in a bad situation.

pimterry:
Next fun problem that this has highlighted: the tests fail, because the browser tests use PhantomJS (now unmaintained), which doesn't support ES6, and this change has let us pull in an update to `resin-settings-client` which uses a new version of `home-or-tmp`, which includes some ES6 code (`const`).

Clearly this has been broken for ages, but Yarn's been hiding it - I'll get it fixed now.

thgreasi:
PhantomJS? I thought that we used headless chrome for the tests!?

thgreasi:
It seems that the billing tests on appveyor are failing but work on Travis. Expecting free plan and getting custom. Maybe someone messed with that account?

pimterry:
> PhantomJS? I thought that we used headless chrome for the tests!?

We do in the UI, but not here (it predates that setup). We could set that up by updating to the latest resin-config-karma release, but that drops coffeescript support, so we'd have to rewrite all the tests in TypeScript. I'd quite like to do that anyway though, so I'm going to look at it and see how hard it is...

> WRT the es6 in dependencies issue, should we just lock home-or-tmp to a lower version, or should we transpile our browser dist after bundling?

We can't directly lock `home-or-tmp`, as it's a dependency of `resin-settings-client`. We could lock that, but I think it would be really good to stay up to date with our internal modules, if we can.

Transpiling the browser dist is a good idea though. That's a little fiddlier than you'd think, because the SDK currently builds with Browserify, not webpack, but we probably want to fix that anyway.

> It seems that the billing tests on appveyor are failing but work on Travis. Expecting free plan and getting custom. Maybe someone messed with that account?

Yes, I think so, I'll look into it soon.

pimterry:
Tried to rewrite all the tests in TS, but it's not as easy as I'd hoped. I've found an easier route though: https://github.com/resin-io-modules/resin-settings-client/pull/50. Once that's released, this will be mergeable and we can look at testing improvements independently later.

pimterry:
Yessss, with the changes elsewhere this finally now successfully builds. @LucianBuzzo @thgreasi @dimitrisnl can I have a review here please?

.prettierrc
```diff
@@ -1,6 +1,6 @@
 {
-  "single-quote": true,
-  "trailing-comma": "all",
-  "use-tabs": true,
-  "print-width": 120
+  "singleQuote": true,
+  "trailingComma": "all",
+  "useTabs": true,
+  "printWidth": 120
```
dimitrisnl: Maybe make the `printWidth` a bit smaller? Something like 80?

pimterry: Hmm, good idea. I'm not sure where this came from originally, but the prettier docs do specifically recommend against it: https://prettier.io/docs/en/options.html#print-width.

I've updated it (actually just removed this line, since 80 is the default), and I agree it looks better. Done :+1:

.npmrc
```diff
@@ -0,0 +1 @@
+package-lock=false
```
thgreasi: Woohoo!

DOCUMENTATION.md
```diff
@@ -172,7 +172,7 @@ The current array of interceptors to use. Interceptors intercept requests made
 internally and are executed in the order they appear in this array for requests,
 and in the reverse order for responses.
 
-**Kind**: static property of <code>[resin](#resin)</code>  
+**Kind**: static property of [<code>resin</code>](#resin)
```
thgreasi: I'm sick of seeing the docs changing all the time... :nauseated_face:

Issue #491, opened 2018-04-18
Add release tag methods


thgreasi:
Just adding this so we can track the respective PR.
HQ: https://github.com/resin-io/hq/issues/150
HQ: https://github.com/resin-io/hq/pull/281
HQ: https://github.com/resin-io/hq/issues/1213
Depends-On: https://github.com/resin-io/resin-api/issues/948

thgreasi:
Fixed by #494 in the multicontainer branch and upcoming v9 sdk release.

PR #494, opened 2018-04-20
Add ReleaseTag methods

thgreasi:
Resolves: #491
Change-Type: minor
HQ: https://github.com/resin-io/hq/pull/281
HQ: https://github.com/resin-io/hq/issues/1213
Depends-On: https://github.com/resin-io/resin-api/pull/952

### DIFF
--- a/lib/models/release.coffee
+++ b/lib/models/release.coffee
@@ -17,12 +17,20 @@ limitations under the License.
 once = require('lodash/once')
 omit = require('lodash/omit')
 errors = require('resin-errors')
+Promise = require('bluebird')
 
 { findCallback, mergePineOptions } = require('../util')
 
 getReleaseModel = (deps, opts) ->
 	{ pine } = deps
 	applicationModel = once -> require('./application')(deps, opts)
+	tagsModel = once -> require('./tags').tagsModel(
+		deps,
+			associatedResource: 'release'
+			# so that the user receives a ResinReleaseNotFound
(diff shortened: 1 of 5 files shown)


thgreasi:
Mostly identical to device-tags but w/o the ability to fetch tags based on a unique release property.

lib/models/release.coffee
```diff
@@ -180,6 +188,162 @@ getReleaseModel = (deps, opts) ->
...
+			# only doing this so that the user get a ResinReleaseNotFound
+			# error when the release doean't actually exist
+			exports.get(id, $select: 'id').get('id')
+			# start the request on parallel, since we don't need
+			# any extra data to wait for
+			exports.tags.getAll(
+				mergePineOptions
+					$filter: release: id
+				, options
+			)
```
pimterry: We could do these two together, couldn't we? Can't you get the release by id, and expand to get the tags too? I think that will throw properly if the release doesn't exist, and let you grab the tags from there very easily as well.

tests/integration/models/release.spec.coffee
```diff
@@ -115,3 +122,36 @@ describe 'Release Model', ->
 							build_log: 'web log'
 						]
 
+		describe.only 'resin.models.release.tags', ->
```
pimterry: Don't commit this `.only`! This disables all of the rest of the test suite

thgreasi: Oh my...

PR #500, opened 2018-05-03
Fix tag upserts for newer pine

thgreasi:
Resolves: #499
Change-Type: patch


thgreasi:
Closing in favor of #503

PR #503, opened 2018-05-09
Fix ResourceNotFound detection on tag upserts for newer pine

thgreasi:
Resolves: #499
Change-Type: patch


thgreasi:
Keeps the Pine 6 related code for now but I don't have a strong opinion on whether we should completely drop that right now or later.

thgreasi:
The envvars related & billing tests are failing though.
We could update the billing test, but b/c of the translations envvars need services to be first created.

PR #509, opened 2018-05-11
Update supervisor methods to handle MC supervisor changes

pimterry:
Also removed an existing deprecated method - if we're breaking things, might as well clean up a bit.


lib/models/device.coffee
```diff
@@ -895,6 +905,10 @@ getDeviceModel = (deps, opts) ->
 	# @function
 	# @memberof resin.models.device
 	#
+	# @deprecated
+	# This is not supported on multicontainer devices, and will
+	# be removed in future
```
LucianBuzzo: NIT: `removed in future` --> `removed in the future` or `removed in a future release`

lib/models/device.coffee
```diff
@@ -849,6 +855,10 @@ getDeviceModel = (deps, opts) ->
...
```
LucianBuzzo: NIT: `removed in future` --> `removed in the future` or `removed in a future release`

lib/models/device.coffee
```diff
@@ -971,6 +985,159 @@ getDeviceModel = (deps, opts) ->
...
+	exports.startService = (uuidOrId, imageId, callback) ->
+		exports.get uuidOrId,
+			$select: ['id', 'supervisor_version']
+			$expand: belongs_to__application: $select: 'id'
+		.then (device) ->
+			ensureSupervisorCompatibility(
+				device.supervisor_version,
+				MIN_SUPERVISOR_MC_API
+			).then ->
+				appId = device.belongs_to__application[0].id
```
thgreasi: I've started thinking about moving the `getExpanded`,... util methods to the SDK.

lib/models/device.coffee
```diff
@@ -971,6 +985,159 @@ getDeviceModel = (deps, opts) ->
...
 
+	###*
+	# @summary Start service on device
+	# @name startService
+	# @public
+	# @function
+	# @memberof resin.models.device
+	#
+	# @param {String|Number} uuidOrId - device uuid (string) or id (number)
+	# @param {Number} imageId - id of the image to start
```
thgreasi: After reading `Start service on device`, I would expect this to be the service ID :confused: 
Not-for-now: Wouldn't it be great if we could handle the service name as well?
I guess that getting the currently running image from the service name & deviceId isn't straightforward, but I imagine that the consumers might need to do this on their own.

lib/models/device.coffee
```diff
@@ -971,6 +985,159 @@ getDeviceModel = (deps, opts) ->
 		.catch(notFoundResponse, treatAsMissingDevice(uuidOrId))
 		.asCallback(callback)
 
+	###*
+	# @summary Start service on device
```
thgreasi: NIT: How about `Start a service on a device`?

lib/models/device.coffee
```diff
@@ -971,6 +985,159 @@ getDeviceModel = (deps, opts) ->
...
+						appId: appId
+						data: {
+							appId
+							imageId
+						}
+					timeout: CONTAINER_ACTION_ENDPOINT_TIMEOUT
+		.asCallback(callback)
+
+	###*
+	# @summary Stop service on device
```
thgreasi: NIT: How about `Stop a service on a device`?

lib/models/device.coffee
```diff
@@ -971,6 +985,159 @@ getDeviceModel = (deps, opts) ->
...
+						appId: appId
+						data: {
+							appId
+							imageId
+						}
+					timeout: CONTAINER_ACTION_ENDPOINT_TIMEOUT
+		.asCallback(callback)
+
+	###*
+	# @summary Restart service on device
```
thgreasi: NIT: How about `Restart a service on a device`?

lib/models/device.coffee
pimterry: I'm not going to do that right now, but in general I do think you're right. Really imo, those should live in Pine, as nice helper methods for anybody using the pine client.

lib/models/device.coffee
pimterry: It's not that easy. The main problem is that it's possible to have two images from the same service running on one device at the same time, if you're using [handovers](https://docs.resin.io/learn/deploy/release-strategy/update-strategies/#hand-over), or plausibly if something goes wrong somewhere or something. That's important because it's a risky case, and it's a time when you're mostly likely to want to manually kill one of your containers (because your handover has got stuck).

This model is also a _slightly_ nicer in that it makes it hard to hit race conditions, where you try to stop one image while another is downloading and end up stopping a different one, because the supervisor is changing things at the same time you're talking to it. You can avoid that if you specify image ids, but if you can only use services then there's nothing you can do to be more specific.

All that said, I can see it's still not super convenient. What do you think about `startService` + `startServiceByName`? The 2nd would do what you're suggesting (and throw an error if it's ambiguous), but we'd also have the 1st option so you can disambiguate when it matters.

pimterry:
One question left to think about image ids vs service names, but I've pushed changes to fix all the other comments.

lib/models/device.coffee
thgreasi: Yea sure, never meant to have this in this PR :)

lib/models/device.coffee
thgreasi: > All that said, I can see it's still not super convenient. What do you think about startService + startServiceByName? The 2nd would do what you're suggesting (and throw an error if it's ambiguous), but we'd also have the 1st option so you can disambiguate when it matters.

Sounds great :+1:  . I also don't think that this need to block this PR. We could do it in a separate minor release.

pimterry: Created an issue for later: https://github.com/resin-io/resin-sdk/issues/512

Issue #514, opened 2018-05-23
Should allow moving devices across same arch apps


thgreasi:
HQ: https://github.com/resin-io/hq/issues/461
See: https://github.com/resin-io/resin-api/pull/978

PR #515, opened 2018-05-23
Allow moving devices to applications of the same arch type

thgreasi:
Resolves: #514
HQ: https://github.com/resin-io/hq/issues/461
Change-Type: minor

### DIFF
--- a/tests/integration/models/device.spec.coffee
+++ b/tests/integration/models/device.spec.coffee
@@ -971,21 +971,26 @@ describe 'Device Model', ->
 				m.chai.expect(promise).to.be.rejected
 					.and.eventually.have.property('code', 'ResinAmbiguousDevice')
 
-	describe 'given two compatible applications and a single device', ->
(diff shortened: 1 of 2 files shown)


tests/integration/models/device.spec.coffee
```diff
@@ -978,14 +978,19 @@ describe 'Device Model', ->
...
+					deviceType: 'raspberrypi3'
 				application2: resin.models.application.create
 					name: 'BarBaz'
 					applicationType: 'microservices-starter'
-					deviceType: 'raspberry-pi'
+					deviceType: 'raspberrypi3'
+				application3: resin.models.application.create
+					name: 'BazFoo'
+					applicationType: 'microservices-starter'
+					deviceType: 'raspberry-pi2'
```
pimterry: We should update the `describe` above this, which is now wrong (there's now 3 applications)

thgreasi: Yea missed that :laughing:

Issue #516, opened 2018-05-24
Include 'is_active' in the Device interface


dimitrisnl:
Depends-On: https://github.com/resin-io/resin-api/issues/1030

dimitrisnl:
relates-to: https://github.com/resin-io/resin-sdk/issues/513

PR #517, opened 2018-05-28
Include 'is_active' in Device interface

dimitrisnl:
Connects-To: #516
Change-Type: minor
---- Autogenerated Waffleboard Connection: Connects to #516

### DIFF
--- a/typings/resin-sdk.d.ts
+++ b/typings/resin-sdk.d.ts
@@ -410,6 +410,7 @@ declare namespace ResinSdk {
 		is_locked_until__date: string;
 		is_on__commit: string;
 		is_web_accessible: boolean;
+		is_active: boolean;
 		is_online: boolean;
 		last_connectivity_event: string;
 		latitude?: string;


dimitrisnl:
Should be merged once v.5.1.0 gets in production

lib/models/device.coffee
```diff
@@ -562,6 +562,36 @@ getDeviceModel = (deps, opts) ->
...
+	# });
+	#
+	# @example
+	# resin.models.device.isActive('7cf02a6', function(error, isActive) {
+	# 	if (error) throw error;
+	# 	console.log('Is device online?', isActive);
+	# });
+	###
+	exports.isActive = (uuidOrId, callback) ->
+		exports.get(uuidOrId, $select: 'is_active').get('is_active').asCallback(callback)
```
pimterry: What's the use case for this method? We generally don't want to add a separate method in the SDK for each field, unless there's a really good reason somebody would need an easy way to specifically look up this one field.

In the UI for example, I imagine we'd load this as part of the `devices.getAll()/device.get(id)` call when loading the content for the page, not with a separate request for each device involved.

tests/integration/models/device.spec.coffee
```diff
@@ -412,6 +412,24 @@ describe 'Device Model', ->
 				promise = resin.models.device.isOnline(999999)
 				m.chai.expect(promise).to.be.rejectedWith('Device not found: 999999')
 
+		describe.only 'resin.models.device.isActive()', ->
```
pimterry: Don't commit `.only`! This disables the whole of the rest of the SDK test suite.

It's easy to miss ofc. Good opportunity for a shameless plug: https://github.com/pimterry/git-confirm (with `git config --add hooks.confirm.match ".only"`)

dimitrisnl:
Oh, I assumed it was necessary to have one. Will remove it if it won't
serve any purpose.

dimitrisnl:
Kept the bare minimum

Issue #527, opened 2018-06-20
Add deprecation warnings for resin.logs.clear() and resin.logs.historySinceClear()


flesler:
In preparation for #525 for v9

pimterry:
Fixed in 8.2.2

Issue #525, opened 2018-06-19
Use the API's custom endpoint for reading device logs (MC-version)


flesler:
See: https://github.com/resin-io/resin-api/issues/1062

PR #528, opened 2018-06-20
Mark resin.logs.clear() & resin.logs.historySinceClear() as deprecated

flesler:
Change-Type: patch
Connects-To: #527 

In preparation for #525 for v9
---- Autogenerated Waffleboard Connection: Connects to #527

**Question:** Should I regenerate the docs myself or will it be done automatically on release?


LucianBuzzo:
@flesler Can you rebuild the docs please?

flesler:
@LucianBuzzo I had asked above if I should or not. Now that I was able to fix the build locally I just did. All good now?

lib/logs.coffee
```diff
@@ -193,6 +194,7 @@ getLogs = (deps, opts) ->
 	# @function
 	# @public
 	# @memberof resin.logs
+	# @deprecated use resin.logs.history() instead, since .clear() is deprecated
```
thgreasi: How about
> this method will be removed in the next major release, please use resin.logs.history() instead

lib/logs.coffee
```diff
@@ -67,6 +67,7 @@ getLogs = (deps, opts) ->
 	# @summary Event fired when the logs have been cleared
 	# @event clear
 	# @memberof resin.logs.LogSubscription
+	# @deprecated since .clear() is deprecated, so is this event
```
thgreasi: How about simply
> this event will be removed in the next major release

lib/logs.coffee
```diff
@@ -244,6 +246,7 @@ getLogs = (deps, opts) ->
 	# @function
 	# @public
 	# @memberof resin.logs
+	# @deprecated .clear() will be removed in the next major release
```
thgreasi: Ultra-NIT: I would use `this method` to be consistent with the rest deprecation warning of the SDK

lib/logs.coffee
flesler: Changed

flesler:
@thgreasi deleted the test in another commit

PR #532, opened 2018-06-26
Add user__is_member_of__application in User interface

dimitrisnl:
Connects-To: #531
Change-Type: minor
---- Autogenerated Waffleboard Connection: Connects to #531


dimitrisnl:
@thgreasi Should I be worried of the failing tests? They don't point towards my change.

thgreasi:
@dimitrisnl Yea, I would saw that we can merge this since it's just a typing addition and we have already identified the API change that causes the tests to fail and a fix is already merged and pending deploy.

PR #529, opened 2018-06-20
Use the API's custom endpoint for reading device logs

flesler:

PR #534, opened 2018-06-29
Upgrade uglify & include (now-supported) streaming dependencies

pimterry:
Change-Type: patch

These upgrades let us bundle correctly when using dependencies with ES6 features (mainly `const`), and this also means we can stop excluding dependencies like rindle/zlib/progress-stream that might be useful, and will be required once the new resin-request is released (i.e. https://github.com/resin-io-modules/resin-request/pull/116).


pimterry:
This is required for #529, but I'm doing it as a separate PR because we need to merge it before resin-request gets released (or all SDK builds will fail), and we can't merge #529 until the API parts for it are all in production.

PR #537, opened 2018-07-06
Print a 'please update' warning if any legacy SDK logs methods are used

### LINKED ISSUE #530: Print a 'please upgrade' warning when calling logs methods in v8 methods
This is necessary as these are going to stop working very shortly.

We can implement now, but this should not be released until there's a v9 release available that does support the new logging mechanism.

pimterry:
Fixes #530. These will stop working in about a month, so we need to push people hard to update.

This should not be released until #529 is ready to go (because otherwise the 'please upgrade' message is useless), and then we should merge this one just a few minutes before the other.

Conveniently this will also affect all future CLI installs, which is why I've left the 'please update' message studiously ambiguous.

Logs with `console.error`, so it should be a) nice & noticeable in the browser b) stderr in node (so hopefully obviously, and separate from console.log'd log output)


pimterry:
https://github.com/resin-io/resin-sdk/pull/529 is ready - shipping these warnings first

PR #540, opened 2018-07-11
*BREAKING*: the version option in os.getConfig() is now required

### LINKED ISSUE #533: os.getConfig() should require a version
https://www.flowdock.com/app/rulemotion/resin-frontend/threads/ByWClPVHC-aS68iP1ff6-KEeb4n

Defaulting to oldest/newest versions causes problems, we should require usage to be explicit.

pimterry:
We need to make sure this breaking change gets in before the MC release, and that's now ultra-imminent...

Fixes #533 
Change-Type: major

### DIFF
--- a/lib/models/os.coffee
+++ b/lib/models/os.coffee
@@ -322,7 +322,8 @@ getOsModel = (deps, opts) ->
 	# @memberof resin.models.os
 	#
 	# @param {String|Number} nameOrId - application name (string) or id (number).
-	# @param {Object} [options={}] - OS configuration options to use.
+	# @param {Object} options - OS configuration options to use.
+	# @param {String} options.version - Required: the OS version of the image.
 	# @param {String} [options.network='ethernet'] - The network type that
 	# the device will use, one of 'ethernet' or 'wifi'.
 	# @param {Number} [options.appUpdatePollInterval] - How often the OS checks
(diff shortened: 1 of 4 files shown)


lib/models/os.coffee
```diff
@@ -322,7 +322,8 @@ getOsModel = (deps, opts) ->
 	# @memberof resin.models.os
 	#
 	# @param {String|Number} nameOrId - application name (string) or id (number).
-	# @param {Object} [options={}] - OS configuration options to use.
+	# @param {Object} options - OS configuration options to use.
+	# @param {String} options.version - Required: the OS version of the image.
```
thgreasi: * We should also update the examples.
* `resin.models.os.getConfig('MyApp', { version })` ~~seems a bit weird. Could go with it, but since we are breaking things, how about changing to use a single object param like we did in a few other places in the MC branch?~~ Didn't have in mind that it can accept other options as well... Since we are already on this, how about adding those as optional `options.param`s?

pimterry: > We should also update the examples.

Very true, fixed

> Since we are already on this, how about adding those as optional options.params?

Yeah, I'm not totally sure about this either way. Right now it doesn't quite match our API in most other places, where options are purely optional, which isn't great. I think I'd rather keep it in the options object though, for a few reasons:

* It's easier to migrate for existing users (many of whom will hopefully already be passing a version anyway)
* I expect that most people will be passing at least some other options in here anyway, not just version alone.
* This way it matches the actual API that way (which includes the version in the options object like any other option)
* With a version + a parameters option, we'd have to decide what to do if you pass a version argument normally _and_ in the parameters (since we'd be squishing them straight back into one object immediately anyway)
* Extra parameters are noisy, and I want less of them :smiley:.

PR #543, opened 2018-07-18
Pin bluebird typings to the latest working version

thgreasi:
v3.5.22 uses TS 2.8 conditional types, which break our build with the TS version that we are currently on.

Change-type: patch


thgreasi:
Oh, I forgot to update the yarn.lock that we still have in master!

PR #548, opened 2018-07-20
Support 'count' in the SDK, and update LogMessage typings

pimterry:
Count's now sent to the API from both subscribe & history requests. I've removed the documentation about a 'default' value here, which was making assumptions about API behaviour.

We talked a little about adding a new default value for count - imo if we do, we should do it at the API level, rather than the frontend. That's simpler, and ensures it's consistent everywhere, even for users talking to the API directly.

Typings are updated to include `isStdErr`, and properly reflect the types the new API endpoint now returns (according to https://github.com/resin-io/resin-api/blob/master/src/lib/device-logs/struct.ts#L26-L35)


flesler:
[FlowDock discussion](https://www.flowdock.com/app/rulemotion/resin-tech/threads/GXl3-qdGVudlt1D96hPrgKr3oXU)

PR #556, opened 2018-08-02
Add support for 'all' in logs requests

### LINKED ISSUE #555: Support an 'all' option for the logs history count

pimterry:
Tested with history in the dashboard, seems to work fine against current production.

~~We still shouldn't merge this until the API supports it though, as it's not actually supported in reality, and the current default for subscribe isn't 'all', so we can't even sneakily pretend it's supported already there.~~

Fixes #555


lib/logs.coffee
```diff
@@ -135,7 +135,7 @@ getLogs = (deps, opts) ->
 	#
 	# @param {String|Number} uuidOrId - device uuid (string) or id (number)
 	# @param {Object} [options] - options
-	# @param {Number} [options.count] - number of historical messages to include
+	# @param {Number} [options.count] - number of historical messages to include (or 'all')
```
thgreasi: Should we also change the `{Number}` to `{Number|String}` (or even `{Number|'all'}` if it's allowed)?

lib/logs.coffee
```diff
@@ -186,7 +186,7 @@ getLogs = (deps, opts) ->
 	# @param {String|Number} uuidOrId - device uuid (string) or id (number)
 
 	# @param {Object} [options] - options
-	# @param {Number} [options.count] - number of log messages to return
+	# @param {Number} [options.count] - number of log messages to return (or 'all')
```
thgreasi: Ditto

pimterry:
JSdoc types & defaults now also updated.

Needs to wait until the API has been released, but then this just needs a rebase and it's good to go I think.

PR #560, opened 2018-08-07
Fix bug in log stream setup error reporting

### LINKED ISSUE #559: Logs stream should report setup errors correctly

pimterry:
Fixes #559

Tiny typo, but it results in an uncaught exception and no reported error for any failures in the initial logs subscription request.

Change-type: patch
Signed-off-by: Tim Perry <tim@resin.io>


thgreasi:
I feel like a primitive that the coffeescript linter didn't catch that...
After so much TS I'm used to taking for granted such checks...

Issue #462, opened 2018-01-18
Narrow the catch clauses in auth methods to better report errors


pimterry:
Right now lots of auth methods assume that any error means we're not logged in (e.g. https://github.com/resin-io/resin-sdk/blob/master/lib/auth.coffee#L245-L246). It would be good to narrow that a little - it'd be better to accurately report if we fail to connect to the server, for example.

PR #566, opened 2018-08-09
462 narrow catch clauses

amdomanska:


lib/auth.coffee
```diff
@@ -23,6 +23,15 @@ getAuth = (deps, opts) ->
 	twoFactor = require('./2fa')(deps, opts)
 	exports = {}
 
+	normalizeError = (err) ->		#todo
+		# console.log(err.code)
```
pimterry: We should drop these leftover comments

lib/auth.coffee
```diff
@@ -23,6 +23,15 @@ getAuth = (deps, opts) ->
 	twoFactor = require('./2fa')(deps, opts)
 	exports = {}
 
+	normalizeError = (err) ->		#todo
```
pimterry: Can we make this name a tiny bit more specific? It's not clear from this exactly what this does. `normalizeAuthError` would be a good start I think, but suggestions welcome.

tests/integration/auth.spec.coffee
```diff
@@ -3,7 +3,7 @@ m = require('mochainon')
 
 { resin, sdkOpts, credentials, givenLoggedInUser, givenLoggedInUserWithApiKey } = require('./setup')
 
-describe 'SDK authentication', ->
+describe.only 'SDK authentication', ->
```
pimterry: This is good for local testing, but you shouldn't commit it, as it'll disable all the other tests

tests/integration/auth.spec.coffee
```diff
@@ -42,6 +42,13 @@ describe 'SDK authentication', ->
 
 				m.chai.expect(promise).to.be.rejectedWith('Unauthorized')
 
+		describe 'resin.auth.getToken()', ->
+
+			it 'shoud be rejected', ->
```
pimterry: Typo: `shuod` -> `should`

lib/auth.coffee
```diff
@@ -240,10 +249,9 @@ getAuth = (deps, opts) ->
 	exports.getToken = (callback) ->
 		auth.getKey()
 		.then (savedToken) ->
-			throw new errors.ResinNotLoggedIn() if not savedToken?
```
pimterry: Why don't we need this line? Does `getKey` throw for us if it's undefined?

tests/integration/auth.spec.coffee
dimitrisnl: If the tests pass successfully, it would be nice to remove this limitation.

lib/auth.coffee
dimitrisnl: There are some forgotten logs.

DOCUMENTATION.md
```diff
@@ -3492,22 +3492,15 @@ This method registers a new api key for the current user with the name given.
 
 **Example**  
 ```js
-resin.models.apiKey.create(apiKeyName).then(function(apiKey) {
-	console.log(apiKey);
-});
+resin.models.apiKey.create(apiKeyName).then(function(apiKey) {	console.log(apiKey);});
```
pimterry: These `^M` characters are because of line ending issues, which I think means you're running on Windows and you need extra git settings. Looks like you need to follow this: https://help.github.com/articles/dealing-with-line-endings/#platform-windows, which means running:

```
git config --global core.autocrlf true
```

Then if you readd this file, this diff should go away.

lib/auth.coffee
```diff
@@ -23,6 +23,15 @@ getAuth = (deps, opts) ->
 	twoFactor = require('./2fa')(deps, opts)
 	exports = {}
 
+	normalizeError = (err) ->		#todo
+		# console.log(err.code)
+		if err.statusCode == 401
+			return new errors.ResinNotLoggedIn()
+		else if err.code == 'ResinMalformedToken' and not savedToken?
```
pimterry: `savedToken` isn't defined in this function. I don't think you need that check anyway - if we get a malformed token error somehow, then I think we know we're definitely not logged in.

tests/integration/auth.spec.coffee
dimitrisnl: 🚓 Typo there Tim

lib/auth.coffee
amdomanska: All done. Sorry for that.

tests/integration/auth.spec.coffee
amdomanska: I'm going to squash all these commits to one. Next time will do before creating the PR. Thank you :)

lib/auth.coffee
amdomanska: That will but not only when it's null or undefined but also when it's incorrect because of other reasons (eg. token set to a number instead of string). In such case the ResinMalformedToken error will be thrown.

amdomanska: I edited the commits. Do you think it's ok now?

pimterry:
The overall commit format & structure now looks good! The only remaining thing there is the first line content, which is now just `Fixes #462`.

Once this PR gets merged, that first line is what will become the description of the change in the changelog: https://github.com/resin-io/resin-sdk/blob/master/CHANGELOG.md. We use a bot called versionbot which automatically does that, and also bumps the version, tags the release in git, and thereby triggers the job that publishes the built code to npm.

For the first line of your commit, in general it should be a short clear imperative description of the result of your changes in this version. Fixes ABC is technically correct, but doesn't make the changelog very easy to skim.

I'd suggest something like: `Report authentication errors more accurately`.

You can then put `Fixes #462` in the body of the commit, if you'd like, up to you.

pimterry:
@amdomanska it looks like somehow you've got commits on this branch that are on master, but also duplicated in this branch. Not sure how that's happened, but we should remove them - the only commit here should be yours, and at the bottom of the build results, it should say the branch is up to date.

There's also a couple of my comments remaining. Github's started collapsing them as outdated, but I think that's just because of the rebase.

It sounds like you're moving on to Etcher soon, so I'm not sure what your plan is, or if you're working on wrapping this up already. Feel free to ping me though for the next couple of hours or on Monday if you want a hand on the last bits here.

amdomanska:
@pimterry I'm definately going to finish it (today or Monday). I'll make sure I checked all your and Dimitrios' comments and make it finally work.

About that malformed token - are you sure there is no possibility that token will be wrong (eg. number instead of string) when we are still logged in? We just wanted to make sure that in such situation we've got ResinMalformedToken error thrown @LucianBuzzo

amdomanska:
@pimterry I tried to squash all these comments to one and it looks quite fine now BUT it still says "pimterry authored" and I have no idea how to change it, sorry!

dimitrisnl:
The commit seems to be credited to Tim

screen shot 2018-08-10 at 9 47 54 pm

amdomanska:
Normally it adds me as an author correctly. Looks like I totally messed that up while squashing the commits and amending them. Am I really supposed to manually resolve the conflicts with versions in CHANGELOG and package.json? Doesn't look like a good idea but that's what git wants now from me.

dimitrisnl:
@amdomanska Ah I see. You should probably check out these two files, since VB will take care of them.

Did you squash the Version Bot commits too?

pimterry:
Looks great! Unfortunately the tests won't pass as production's in a bad state for now with all the orgs/billing big deploys going on. Everything should be all good later today though.

amdomanska:
I was just wondering what is going on. Thank for the info and the review.

Issue #434, opened 2017-11-20
Should provide staged-releases related methods


thgreasi:
These are probably going to change a bit after MC gets released.
See: https://github.com/resin-io-playground/staged-releases
HQ: https://github.com/balena-io/balena/issues/297

<img src="https://frontapp.com/assets/img/icons/favicon-32x32.png" height="16" width="16" alt="Front logo" /> [Front conversations](https://app.frontapp.com/open/top_b2ar)

PR #572, opened 2018-08-26
Add release tracking methods

amdomanska:
Resolves: #434
HQ: https://github.com/balena-io/balena/issues/297
Change-type: minor
---- Autogenerated Waffleboard Connection: Connects to #434

### DIFF
--- a/DOCUMENTATION.md
+++ b/DOCUMENTATION.md
@@ -46,6 +46,11 @@ If you feel something is missing, not clear or could be improved, please don't h
             * [.purge(appId)](#balena.models.application.purge) ⇒ <code>Promise</code>
             * [.shutdown(appId, [options])](#balena.models.application.shutdown) ⇒ <code>Promise</code>
             * [.reboot(appId, [options])](#balena.models.application.reboot) ⇒ <code>Promise</code>
+            * [.willTrackNewReleases(nameOrId)](#balena.models.application.willTrackNewReleases) ⇒ <code>Promise</code>
+            * [.isTrackingLatestRelease(nameOrId)](#balena.models.application.isTrackingLatestRelease) ⇒ <code>Promise</code>
+            * [.pinToRelease(nameOrId, fullReleaseHash)](#balena.models.application.pinToRelease) ⇒ <code>Promise</code>
+            * [.getTargetReleaseHash(nameOrId)](#balena.models.application.getTargetReleaseHash) ⇒ <code>Promise</code>
+            * [.trackLatestRelease(nameOrId)](#balena.models.application.trackLatestRelease) ⇒ <code>Promise</code>
             * [.enableDeviceUrls(nameOrId)](#balena.models.application.enableDeviceUrls) ⇒ <code>Promise</code>
             * [.disableDeviceUrls(nameOrId)](#balena.models.application.disableDeviceUrls) ⇒ <code>Promise</code>
             * [.grantSupportAccess(nameOrId, expiryTimestamp)](#balena.models.application.grantSupportAccess) ⇒ <code>Promise</code>
@@ -122,6 +127,10 @@ If you feel something is missing, not clear or could be improved, please don't h
             * [.grantSupportAccess(uuidOrId, expiryTimestamp)](#balena.models.device.grantSupportAccess) ⇒ <code>Promise</code>
             * [.revokeSupportAccess(uuidOrId)](#balena.models.device.revokeSupportAccess) ⇒ <code>Promise</code>
             * [.lastOnline(device)](#balena.models.device.lastOnline) ⇒ <code>String</code>
+            * [.isTrackingApplicationRelease(uuidOrId)](#balena.models.device.isTrackingApplicationRelease) ⇒ <code>Promise</code>
+            * [.getTargetReleaseHash(uuidOrId)](#balena.models.device.getTargetReleaseHash) ⇒ <code>Promise</code>
+            * [.pinToRelease(uuidOrId, fullReleaseHashOrId)](#balena.models.device.pinToRelease) ⇒ <code>Promise</code>
+            * [.trackApplicationRelease(uuidOrId)](#balena.models.device.trackApplicationRelease) ⇒ <code>Promise</code>
         * [.apiKey](#balena.models.apiKey) : <code>object</code>
             * [.create(name, [description])](#balena.models.apiKey.create) ⇒ <code>Promise</code>
             * [.getAll([options])](#balena.models.apiKey.getAll) ⇒ <code>Promise</code>
@@ -153,6 +162,7 @@ If you feel something is missing, not clear or could be improved, please don't h
             * [.get(id, [options])](#balena.models.release.get) ⇒ <code>Promise</code>
             * [.getWithImageDetails(id, [options])](#balena.models.release.getWithImageDetails) ⇒ <code>Promise</code>
             * [.getAllByApplication(nameOrId, [options])](#balena.models.release.getAllByApplication) ⇒ <code>Promise</code>
+            * [.getLatestByApplication(nameOrId, [options])](#balena.models.release.getLatestByApplication) ⇒ <code>Promise</code>
         * [.service](#balena.models.service) : <code>object</code>
             * [.var](#balena.models.service.var) : <code>object</code>
                 * [.getAllByService(id, [options])](#balena.models.service.var.getAllByService) ⇒ <code>Promise</code>
@@ -329,6 +339,11 @@ balena.models.device.get(123).catch(function (error) {
         * [.purge(appId)](#balena.models.application.purge) ⇒ <code>Promise</code>
         * [.shutdown(appId, [options])](#balena.models.application.shutdown) ⇒ <code>Promise</code>
         * [.reboot(appId, [options])](#balena.models.application.reboot) ⇒ <code>Promise</code>
+        * [.willTrackNewReleases(nameOrId)](#balena.models.application.willTrackNewReleases) ⇒ <code>Promise</code>
+        * [.isTrackingLatestRelease(nameOrId)](#balena.models.application.isTrackingLatestRelease) ⇒ <code>Promise</code>
+        * [.pinToRelease(nameOrId, fullReleaseHash)](#balena.models.application.pinToRelease) ⇒ <code>Promise</code>
+        * [.getTargetReleaseHash(nameOrId)](#balena.models.application.getTargetReleaseHash) ⇒ <code>Promise</code>
+        * [.trackLatestRelease(nameOrId)](#balena.models.application.trackLatestRelease) ⇒ <code>Promise</code>
         * [.enableDeviceUrls(nameOrId)](#balena.models.application.enableDeviceUrls) ⇒ <code>Promise</code>
         * [.disableDeviceUrls(nameOrId)](#balena.models.application.disableDeviceUrls) ⇒ <code>Promise</code>
         * [.grantSupportAccess(nameOrId, expiryTimestamp)](#balena.models.application.grantSupportAccess) ⇒ <code>Promise</code>
@@ -405,6 +420,10 @@ balena.models.device.get(123).catch(function (error) {
         * [.grantSupportAccess(uuidOrId, expiryTimestamp)](#balena.models.device.grantSupportAccess) ⇒ <code>Promise</code>
         * [.revokeSupportAccess(uuidOrId)](#balena.models.device.revokeSupportAccess) ⇒ <code>Promise</code>
         * [.lastOnline(device)](#balena.models.device.lastOnline) ⇒ <code>String</code>
+        * [.isTrackingApplicationRelease(uuidOrId)](#balena.models.device.isTrackingApplicationRelease) ⇒ <code>Promise</code>
+        * [.getTargetReleaseHash(uuidOrId)](#balena.models.device.getTargetReleaseHash) ⇒ <code>Promise</code>
+        * [.pinToRelease(uuidOrId, fullReleaseHashOrId)](#balena.models.device.pinToRelease) ⇒ <code>Promise</code>
+        * [.trackApplicationRelease(uuidOrId)](#balena.models.device.trackApplicationRelease) ⇒ <code>Promise</code>
     * [.apiKey](#balena.models.apiKey) : <code>object</code>
         * [.create(name, [description])](#balena.models.apiKey.create) ⇒ <code>Promise</code>
         * [.getAll([options])](#balena.models.apiKey.getAll) ⇒ <code>Promise</code>
@@ -436,6 +455,7 @@ balena.models.device.get(123).catch(function (error) {
         * [.get(id, [options])](#balena.models.release.get) ⇒ <code>Promise</code>
         * [.getWithImageDetails(id, [options])](#balena.models.release.getWithImageDetails) ⇒ <code>Promise</code>
         * [.getAllByApplication(nameOrId, [options])](#balena.models.release.getAllByApplication) ⇒ <code>Promise</code>
+        * [.getLatestByApplication(nameOrId, [options])](#balena.models.release.getLatestByApplication) ⇒ <code>Promise</code>
     * [.service](#balena.models.service) : <code>object</code>
         * [.var](#balena.models.service.var) : <code>object</code>
             * [.getAllByService(id, [options])](#balena.models.service.var.getAllByService) ⇒ <code>Promise</code>
@@ -490,6 +510,11 @@ balena.models.device.get(123).catch(function (error) {
     * [.purge(appId)](#balena.models.application.purge) ⇒ <code>Promise</code>
     * [.shutdown(appId, [options])](#balena.models.application.shutdown) ⇒ <code>Promise</code>
     * [.reboot(appId, [options])](#balena.models.application.reboot) ⇒ <code>Promise</code>
+    * [.willTrackNewReleases(nameOrId)](#balena.models.application.willTrackNewReleases) ⇒ <code>Promise</code>
+    * [.isTrackingLatestRelease(nameOrId)](#balena.models.application.isTrackingLatestRelease) ⇒ <code>Promise</code>
+    * [.pinToRelease(nameOrId, fullReleaseHash)](#balena.models.application.pinToRelease) ⇒ <code>Promise</code>
+    * [.getTargetReleaseHash(nameOrId)](#balena.models.application.getTargetReleaseHash) ⇒ <code>Promise</code>
+    * [.trackLatestRelease(nameOrId)](#balena.models.application.trackLatestRelease) ⇒ <code>Promise</code>
     * [.enableDeviceUrls(nameOrId)](#balena.models.application.enableDeviceUrls) ⇒ <code>Promise</code>
     * [.disableDeviceUrls(nameOrId)](#balena.models.application.disableDeviceUrls) ⇒ <code>Promise</code>
     * [.grantSupportAccess(nameOrId, expiryTimestamp)](#balena.models.application.grantSupportAccess) ⇒ <code>Promise</code>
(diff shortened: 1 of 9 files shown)


DOCUMENTATION.md
```diff
@@ -3493,21 +3611,15 @@ This method registers a new api key for the current user with the name given.
 **Example**  
 ```js
 resin.models.apiKey.create(apiKeyName).then(function(apiKey) {
-	console.log(apiKey);
-});
+	console.log(apiKey);});
```
amdomanska: I'm going to fix all these eols on Thursday on my computer.

amdomanska:
I created the "follow-up" issue:
https://github.com/resin-io/resin-sdk/issues/573

lib/models/application.coffee
```diff
@@ -694,6 +694,128 @@ getApplicationModel = (deps, opts) ->
...
+	#
+	# @example
+	# resin.models.application.getRelease('MyApp').then(function(release) {
+	# 	console.log(release);
+	# });
+	###
+
+	exports.getRelease = (nameOrId, callback) ->
+		exports.get(nameOrId)
+    .get('commit')
```
pimterry: Indentation is slightly funny here

lib/models/application.coffee
```diff
@@ -694,6 +694,128 @@ getApplicationModel = (deps, opts) ->
...
+	# @returns {Promise}
+	#
+	# @example
+	# resin.models.application.getRelease('MyApp').then(function(release) {
+	# 	console.log(release);
+	# });
+	###
+
+	exports.getRelease = (nameOrId, callback) ->
+		exports.get(nameOrId)
```
pimterry: If we only want the `commit` field, it'd be good to `$select: 'commit'` here. It works anyway, but that way we avoid extra work for the API & DB, and avoid transferring lots of unnecessary data.

lib/models/application.coffee
```diff
@@ -694,6 +694,128 @@ getApplicationModel = (deps, opts) ->
...
+	#
+	# @param {String|Number} nameOrId - application name (string) or id (number)
+	#
+	# @description this method requires rolling release to be disabled prior to call
+	#
+	# @returns {Promise}
+	#
+	# @example
+	# resin.models.application.disableRollingRelease('MyApp').then(
+	# 	resin.models.application.isRollingReleaseEnabled('MyApp').then(function(isEnabled) {
```
pimterry: This example doesn't use the function it's documenting.

lib/models/application.coffee
```diff
@@ -694,6 +694,128 @@ getApplicationModel = (deps, opts) ->
...
+		.get('should_track_latest_release').asCallback(callback)
+
+	###*
+	# @summary Set application to particular commit
+	# @name setApplicationToRelease
+	# @public
+	# @function
+	# @memberof resin.models.application
+	#
+	# @param {String|Number} nameOrId - application name (string) or id (number)
```
pimterry: This is missing the `@param` for the commit hash.

lib/models/application.coffee
```diff
@@ -694,6 +694,128 @@ getApplicationModel = (deps, opts) ->
...
+	#
+	# @example
+	# resin.models.application.disableRollingRelease('MyApp').then(
+	# 	resin.models.application.isRollingReleaseEnabled('MyApp').then(function(isEnabled) {
+	# 		console.log(isEnabled);
+	# 	})
+	# );
+	###
+
+	exports.setApplicationToRelease = (nameOrId, fullCommitHash, callback) ->
```
pimterry: Can we rename this to `fullReleaseHash`? Just to be clearer that this is the release identifier.

Not a big deal right now, but it'll be useful to make sure we're talking in terms of releases. Later on, we're probably going to want to also allow release ids here too, and in future more and more releases will come from `resin push`, not actual commits.

lib/models/application.coffee
```diff
@@ -694,6 +694,128 @@ getApplicationModel = (deps, opts) ->
...
+	# @param {String|Number} nameOrId - application name (string) or id (number)
+	# @returns {Promise}
+	#
+	# @example
+	# resin.models.application.getRelease('MyApp').then(function(release) {
+	# 	console.log(release);
+	# });
+	###
+
+	exports.getRelease = (nameOrId, callback) ->
```
pimterry: This returns the hash, right? Can we call it `getReleaseHash` to make that clearer?

lib/models/device.coffee
```diff
@@ -101,13 +101,26 @@ getDeviceModel = (deps, opts) ->
 
 	# Internal method for uuid/id disambiguation
 	# Note that this throws an exception for missing uuids, but not missing ids
+	# method for staged-releases get-device-id.sh
```
pimterry: I think this comment is probably just your own notes, right? I think this can go.

lib/models/device.coffee
```diff
@@ -101,13 +101,26 @@ getDeviceModel = (deps, opts) ->
...
 
+	###*
+	# @summary Get device id for a specific device on a resin.io application
+	# @param {String} uuid - uuid of specific device
+	# @memberof resin.models.device
+	# @returns {Number} DeviceId
+	# @example
+	# resin.models.device.getDeviceId(device.uuid)
+	###
+	exports.getDeviceId = (uuid, callback) ->
```
pimterry: Why do we need this as a public method? I'm not sure it's something users should be calling. It's part of staged releases, but I think just to use internally, not something users should need to care about.

tests/integration/models/application.spec.coffee
```diff
@@ -13,6 +13,79 @@ describe 'Application Model', ->
...
+
+		it 'should disable and enable rolling release', ->
+
+			resin.models.application.disableRollingRelease(@application.id)
+			.then =>
+				promise = resin.models.application.isRollingReleaseEnabled(@application.id)
+				m.chai.expect(promise).to.eventually.be.false
+
+			.then =>
+
```
pimterry: Some of these new lines are a little odd - personally I'd skip these two here.

tests/integration/models/application.spec.coffee
```diff
@@ -13,6 +13,79 @@ describe 'Application Model', ->
...
+				m.chai.expect(promise).to.eventually.be.true
+
+		describe 'setApplicationToRelease()', ->
+
+			it 'should throw the error if rolling release is enabled', ->
+
+				resin.models.application.enableRollingRelease(@application.id)
+				.then =>
+					promise = resin.models.application.isRollingReleaseEnabled(@application.id)
+					m.chai.expect(promise).to.eventually.be.true
```
pimterry: We don't need to retest this here I think. We've got a separate test above that already checks that `enableRollingRelease` works, so we can skip this step imo.

tests/integration/models/application.spec.coffee
```diff
@@ -13,6 +13,79 @@ describe 'Application Model', ->
...
+				.then =>
+					promise = resin.models.application.setApplicationToRelease(@application.id, 'new-release-commit')
+					m.chai.expect(promise).to.be.rejectedWith('Rolling release enabled')
+
+			it 'should set the application to specific release', ->
+
+				resin.models.application.disableRollingRelease(@application.id)
+				.then =>
+					promise = resin.models.application.isRollingReleaseEnabled(@application.id)
+					m.chai.expect(promise).to.eventually.be.false
```
pimterry: Same here

tests/integration/models/application.spec.coffee
```diff
@@ -13,6 +13,79 @@ describe 'Application Model', ->
 
 	givenLoggedInUser()
 
+	describe 'rolling release related methods', ->
```
pimterry: I think this should live inside a `describe 'given ...'` block. We generally use top-level describe blocks to describe the test environment of preexisting data everywhere, and then group by method and functionality like this only within that, where necessary.

If you put this inside `given a single application` block, you'd be able to skip everything before `userId =` since that's already there. I.e. something like:

```coffee
 	describe 'given a single application', ->
 
 		beforeEach ->
 			resin.models.application.create
 				name: 'FooBar'
 				deviceType: 'raspberry-pi'
 			.then (application) =>
 				@application = application

# ^^^ this part already exists below
		...

		describe 'given two releases', ->

			beforeEach ->
				userId = @application.user.__id
 				resin.pine.post
					resource: 'release'
					body:
						belongs_to__application: @application.id
						is_created_by__user: userId
						commit: 'old-release-commit'
						status: 'success'
						source: 'cloud'
						composition: {}
						start_timestamp: 1234
 				.then =>
					resin.pine.post
						resource: 'release'
						body:
							belongs_to__application: @application.id
							is_created_by__user: userId
							commit: 'new-release-commit'
							status: 'success'
							source: 'cloud'
							composition: {}
							start_timestamp: 54321

			# Put your tests here
```

lib/models/application.coffee
```diff
@@ -694,6 +694,128 @@ getApplicationModel = (deps, opts) ->
 			throw err
 		.asCallback(callback)
 
+	###*
+	# @summary Check if rolling release is enabled for specific application
+	# @name isRollingReleaseEnabled
```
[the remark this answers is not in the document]
shaunmulligan: @pimterry yeah I agree it  would be good to have consistency. I think talking about it as `enable/disable latest release tracking` seems good to me and keeps us inline with our current API and I think also inline with the UI designs we have for this feature

lib/models/device.coffee
thgreasi: Given that we were trying to not use device ids in any of our methods, I agree to not add this as part of the public methods.

lib/models/application.coffee
thgreasi: I guess that this comment is no longer valid, since the commit hash isn't part of the `isTrackingLatestRelease()` method call signature

thgreasi:
I've ended up with the following methods:
* `application.isTrackingLatestRelease`
* `application.getReleaseHash`
* `application.setToRelease`
* `application.disableLatestReleaseTracking`
* `application.enableLatestReleaseTracking`
* `device.isTrackingApplicationRelease`
* `device.getTrackedReleaseHash` (to not confuse the hash with the value of `is_on__commit`)
* `device.setToRelease`
* `device.enableApplicationReleaseTracking`

Which make me thinking:
* should we rename `application.getReleaseHash` to `application.getTrackedReleaseHash` for symmetry?
* should we drop `application.disableLatestReleaseTracking` for symmetry?
  This would mean that in order to stop tracking the latest release, you should use `application.setToRelease` which makes sense especially if we also agree the following bullet
* should we rename `application.enableLatestReleaseTracking` & `device.enableApplicationReleaseTracking` to `application.trackLatestRelease` & `device.trackApplicationRelease` respectively?
* should we rename `setToRelease` to `pinToRelease`?

amdomanska:
@thgreasi 
ad1: makes sense
ad2: I would say `application.disableLatestReleaseTracking` should stay for users that are on current release and don't want to update the release in the future. Making them to use `setToRelease` with the current release they are already set sounds counterintuitive. 
ad3: what about `application.trackLatestRelease` and  `device.trackLatestApplicationRelease `?

thgreasi:
@amdomanska WRT point 3 `enableApplicationReleaseTracking`/`trackApplicationRelease` clears the `should_be_running__release` field of the device, making it to track the release that the application `commit` points to. `device.trackLatestApplicationRelease` does sound different than what was described above (tracking the latest is different than tracking whatever the app tracks).

PS: Afaik, the API doesn't have support for `device.trackLatestApplicationRelease` atm even if we wanted to provide that in the SDK.

lib/models/device.coffee
```diff
@@ -1917,6 +1917,170 @@ getDeviceModel = (deps, opts) ->
...
+	#
+	# @example
+	# resin.models.device.setToRelease('7cf02a6, 'f7caf4ff80114deeaefb7ab4447ad9c661c50847', function(error) {
+	# 	if (error) throw error;
+	# 	...
+	# });
+	###
+	exports.setToRelease = (uuidOrId, fullReleaseHashOrId, callback) ->
+		Promise.try ->
+			if isId(uuidOrId) and isId(fullReleaseHashOrId)
```
thgreasi: This will be the branch that the UI will be using for the group action, so that it will need to do as few requests as possible.

thgreasi:
@resin-ci retest

shaunmulligan:
@amdomanska and @thgreasi whats the status on this PR, is it blocked on something?

thgreasi:
@shaunmulligan It's waiting for reviews. I will reping everyone in FD.
I was trying this branch on the dashboard's console yesterday, and it was working nicely. It was great beng able to do staged releases w/o the .sh scripts (using this js api).

thgreasi:
@shaunmulligan now with npm pre-publishing of resinCI you can npm i the result of this branch right away :)
```
npm i resin-sdk@10.1.0-434-stage-releases-methods-314488952d45adc7aaa7353259925745da3c2a37
```

lib/models/application.coffee
```diff
@@ -694,6 +694,180 @@ getApplicationModel = (deps, opts) ->
...
+	# 	...
+	# });
+	#
+	# @example
+	# resin.models.application.setToRelease('MyApp', 'f7caf4ff80114deeaefb7ab4447ad9c661c50847', function(error) {
+	# 	if (error) throw error;
+	# 	...
+	# });
+	###
+	exports.setToRelease = (nameOrId, fullReleaseHash, callback) ->
```
pimterry: Nit: How about `pinToRelease`? (just spotted - you actually suggested this! Good suggestion :+1:)

lib/models/application.coffee
```diff
@@ -694,6 +694,180 @@ getApplicationModel = (deps, opts) ->
...
+	# @name setToRelease
+	# @public
+	# @function
+	# @memberof resin.models.application
+	#
+	# @description Configures the application to run a particular release
+	# and not get updated when the latest release changes.
+	#
+	# @param {String|Number} nameOrId - application name (string) or id (number)
+	# @param {String} fullReleaseHash - the hash of a successful release (string) or id (number)
```
pimterry: This isn't quite right - I don't think we accept ids here, only hashes.

lib/models/application.coffee
```diff
@@ -694,6 +694,180 @@ getApplicationModel = (deps, opts) ->
...
+	# resin.models.application.getTrackedReleaseHash(123).then(function(release) {
+	# 	console.log(release);
+	# });
+	#
+	# @example
+	# resin.models.application.getTrackedReleaseHash('MyApp', function(release) {
+	# 	console.log(release);
+	# });
+	###
+	exports.getTrackedReleaseHash = (nameOrId, callback) ->
```
pimterry: I think `tracked` is a little confusing here. We use it to talk about tracking the latest release elsewhere, to describe following one thing (the latest), that changes what it points to. It's a small thing, but it does feel a bit unclear. What we want to know here is not what we're following, but what the current target of that is.

How about `getCurrentReleaseHash`? That matches the current dashboard designs too (which put `Current` next to the current release).

Alternatively imo `Target` would work well, but we should use that consistently everywhere, and we don't use it right now.

lib/models/device.coffee
```diff
@@ -1917,6 +1917,170 @@ getDeviceModel = (deps, opts) ->
...
+	# resin.models.device.getTrackedReleaseHash('7cf02a6).then(function(release) {
+	# 	console.log(release);
+	# });
+	#
+	# @example
+	# resin.models.device.getTrackedReleaseHash('7cf02a6, function(release) {
+	# 	console.log(release);
+	# });
+	###
+	exports.getTrackedReleaseHash = (uuidOrId, callback) ->
```
pimterry: Ditto `current`/`target`

lib/models/device.coffee
```diff
@@ -1917,6 +1917,170 @@ getDeviceModel = (deps, opts) ->
...
+	# @fulfil {String} - The release hash of the currently tracked release
+	# @returns {Promise}
+	#
+	# @example
+	# resin.models.device.getTrackedReleaseHash('7cf02a6).then(function(release) {
+	# 	console.log(release);
+	# });
+	#
+	# @example
+	# resin.models.device.getTrackedReleaseHash('7cf02a6, function(release) {
```
pimterry: Both these examples are missing a closing quote on the first argument

lib/models/device.coffee
```diff
@@ -1917,6 +1917,170 @@ getDeviceModel = (deps, opts) ->
...
+	# 	...
+	# });
+	#
+	# @example
+	# resin.models.device.setToRelease('7cf02a6, 'f7caf4ff80114deeaefb7ab4447ad9c661c50847', function(error) {
+	# 	if (error) throw error;
+	# 	...
+	# });
+	###
+	exports.setToRelease = (uuidOrId, fullReleaseHashOrId, callback) ->
```
pimterry: `pin` would be nice here too

lib/models/device.coffee
```diff
@@ -1917,6 +1917,170 @@ getDeviceModel = (deps, opts) ->
...
+	# @name setToRelease
+	# @public
+	# @function
+	# @memberof resin.models.device
+	#
+	# @description Configures the device to run a particular release
+	# and not get updated when the current application release changes.
+	#
+	# @param {String|Number} uuidOrId - device uuid (string) or id (number)
+	# @param {String} fullReleaseHash - the hash of a successful release (string) or id (number)
```
pimterry: This one does support ids, so should be `{String|Number}` here too

lib/models/device.coffee
```diff
@@ -1917,6 +1917,170 @@ getDeviceModel = (deps, opts) ->
...
+					belongs_to__application:
+						$select: 'id'
+						$expand:
+							owns__release:
+								$top: 1
+								$select: 'id'
+								$filter:
+									"#{releaseFilterProperty}": fullReleaseHashOrId
+								# if there is a successful release for a specific commit hash
+								# then it will be the last created
```
pimterry: Is that guaranteed? Of course you can't have another successful release after the first one, but can you have a failed, or still in progress release? I don't think the model would reject those.

Also, if there are no successful releases, this will return a failed one instead, which will probably do bad things later on.

Why not just filter for successful releases?

lib/models/device.coffee
```diff
@@ -1917,6 +1917,170 @@ getDeviceModel = (deps, opts) ->
...
+	# @param {String|Number} uuidOrId - device uuid (string) or id (number)
+	# @returns {Promise}
+	#
+	# @example
+	# resin.models.device.enableApplicationReleaseTracking('7cf02a6).then(function() {
+	# 	...
+	# });
+	#
+	# @example
+	# resin.models.device.enableApplicationReleaseTracking('7cf02a6, function(error) {
```
pimterry: These are missing quotes too

lib/models/device.coffee
```diff
@@ -1917,6 +1917,170 @@ getDeviceModel = (deps, opts) ->
...
+	# 	if (error) throw error;
+	# 	...
+	# });
+	###
+	exports.enableApplicationReleaseTracking = (uuidOrId, callback) ->
```
pimterry: As in your comment - I do think renaming this to `trackApplicationRelease` (and the app equivalent to `trackLatestRelease`) is a good plan.

lib/models/application.coffee
```diff
@@ -694,6 +694,180 @@ getApplicationModel = (deps, opts) ->
...
+	# 	...
+	# });
+	#
+	# @example
+	# resin.models.application.disableLatestReleaseTracking('MyApp', function(error) {
+	# 	if (error) throw error;
+	# 	...
+	# });
+	###
+	exports.disableLatestReleaseTracking = (nameOrId, callback) ->
```
pimterry: Agree with your comment - we should remove this.

lib/models/application.coffee
thgreasi: Also changed that in the device model

lib/models/application.coffee
thgreasi: Opps! Nice spot, bad copy paste

lib/models/application.coffee
thgreasi: I agree that we can do better than `tracked`. The new designs only use it as `track: latest`.
We will probably keep it for later when we add release channels.
The only thing I don't like about `current` is that it isn't accurate in the context of a device, were we should better use the same method name for consistency. (A device can be made to truck release B but is still running release A. In that case `getCurrentReleaseHash` probably means A but the method we implement here is meant to return B, so Tracked/Target make more sense.)

lib/models/device.coffee
thgreasi: Taking another look at this, I don't see a reason to not filter only the succeeded releases.
I went with this approach so that the user will get the API error if the release exists but isn't successful instead of ResinReleaseNotFound.
I will add the success filter, since getting a ReleaseNotFoundError isn't totally wrong, since the description explicitly mentions that it should be a successful release.
Makes sense?

lib/models/application.coffee
```diff
@@ -678,6 +678,145 @@ getApplicationModel = (deps, opts) ->
...
+	# 	if (error) throw error;
+	# 	...
+	# });
+	###
+	exports.trackLatestRelease = (nameOrId, callback) ->
+		getId(nameOrId).then (applicationId) ->
+			return pine.patch
+				resource: 'application'
+				id: applicationId
+				body: should_track_latest_release: true
```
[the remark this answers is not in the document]
thgreasi: Found the thread about doing it in an API hook. I guess since v5 is already out, it might be late to expect that to happen soon.
See: https://www.flowdock.com/app/rulemotion/resin-frontend/threads/k5B1N0BEPFHVQJYv-9wdegSs1Ji

PR #576, opened 2018-09-17
Update to resin-pine 8.x

Page-:
This will only break things if you are passing custom options,
the relevant breakages are:
* Remove deprecated expand behaviour, eg `$expand: { x: 'y' }`
* Remove deprecated automatic $ prefixing for `options`,
    eg `select: 'id'`

Change-type: major
Signed-off-by: Pagan Gazzard <page@resin.io>


thgreasi:
That's not going to be passing on resin-ci until we get secrets in there.
We should also revisit the typings as well to reflect the resin-pine changes. Might find some time to give it try later this week if no one does.

Page-:
What do you mean @thgreasi ?  I think those changes were already in the typings from before as the typings never included the deprecated formats

thgreasi:
Yea you are right @Page- , they indeed do not include any of these :)

Page-:
Done @pimterry

pimterry:
I'd love to squeeze this in with other breaking changes, but I'm not sure we've got any coming in the very short term, unless you're planning some for orgs? If they're not super soon though (like, within the next week max), imo we can just do a major bump now and another later on, that's better than holding this for ages.

pimterry:
This bump has ended up being included in the breaking rename PR as we've moved to balena-pine, so I'm going to close this.

PR #580, opened 2018-09-26
tests-billing: Update paid billing tests for the dynamic billing era

thgreasi:
Resolves: #579
Change-type: patch

### DIFF
--- a/tests/integration/models/billing.spec.coffee
+++ b/tests/integration/models/billing.spec.coffee
@@ -143,25 +143,34 @@ describe 'Billing Model', ->
 				resin.models.billing.getPlan()
 				.then (plan) ->
 					m.chai.expect(plan).to.deep.match
-						title: 'TestDev'
-						name: 'TestDev plan'
-						code: 'testdev'
-						tier: 'testdev'
+						title: 'Team member'
+						name: 'Team member plan'
+						code: 'free'
+						tier: 'free'
 						addOns: []
+						addonPlan:
+							addOns: []
+							billing:
+								charges: []
+								currency: 'USD'
+								totalCostCents: '0'
+							code: 'addons'
+							currentPeriodEndDate: ''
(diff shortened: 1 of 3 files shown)


tests/integration/models/billing.spec.coffee
```diff
@@ -172,8 +181,6 @@ describe 'Billing Model', ->
 							name: 'Standard support'
 
 					m.chai.expect(plan).to.have.property('currentPeriodEndDate').that.is.a('string')
-					m.chai.expect(plan).to.have.property('id').that.is.a('string')
```
thgreasi: This one is no longer present, but since we don't have it in the typings, I guess we never used it on the Dashboard

tests/integration/models/billing.spec.coffee
```diff
@@ -143,25 +143,34 @@ describe 'Billing Model', ->
 				resin.models.billing.getPlan()
 				.then (plan) ->
 					m.chai.expect(plan).to.deep.match
-						title: 'TestDev'
-						name: 'TestDev plan'
-						code: 'testdev'
-						tier: 'testdev'
+						title: 'Team member'
+						name: 'Team member plan'
+						code: 'free'
```
dimitrisnl: Is it really free?

The API definition doesn't have a code https://github.com/resin-io/resin-api/blob/master/src/utils/definitions/billing-plans-definition.ts#L148

thgreasi: I guess that it's filled as a default value but you are right this does seem strange.

pimterry: This does seem odd. Something to look at for the API rather than here though.

thgreasi:
@pimterry Yes, I did bring this up last week and mentioned that it's blocking us.

thgreasi:
@resin-ci retest

PR #584, opened 2018-10-01
tests: Use a random test user when no env vars are present

thgreasi:
* This changes the test suit to register a random test user when the respective env vars are not provided, and deletes it after the tests complete. It still uses the `RESINTEST_USERNAME/...` env vars if present, since while developing this I ended up having the staging api rate limit me b/c of the extra user registrations that the suit did. As a result, for practical reasons, I think we should still keep them to not add friction when developing.
* Removes the `RESINTEST_REGISTER_*` env vars and generated a random user as well.
* Drops travis tests, as a way to reduce the load to the API. This shouldn't reduce the effectiveness of the CI's used, since resinCI will be testing the environment that travis was testing before.
* Still uses appveyor since
  * it's nice to test the sdk on a windows machine as well
  * it is responsible for running the paid user tests, until resinCI adds support for secrets
* Adds the required parameters to headless chrome so that it works on resinCI

Resolves: #582
Change-type: minor
See: https://www.flowdock.com/app/rulemotion/resin-frontend/threads/JIU7143SphYOgcVeQVh5n69eycK
See: resin-io/resin-concourse#248
See: https://github.com/GoogleChrome/puppeteer/blob/master/docs/troubleshooting.md#running-on-alpine
Signed-off-by: Thodoris Greasidis <thodoris@resin.io>

PR #586, opened 2018-10-09
tests: Add support for TypeScript test cases

thgreasi:
Spitted these from a local branch that was combining them with #584, since the later one seems blocked.
* Bumps resin-config-karma to v2.2 to support TS tests (which also adds some flags to headless chrome to make it resinCI friendly)
* Adds the configuration required to run TS tests on node & karma along with the required tsconfig & tslint changes.
* Converts the OS tests to TypeScript

Resolves: #583
Change-type: minor


thgreasi:
@resin-ci retest

Page-:
Not a review per se, but I'd like to note that I want to keep the availability of non-typescript cases in order to test we have appropriate type guards in place for non-typescript users of the sdk (who won't have the typescript compiler guarding against passing the wrong types)

thgreasi:
@Page- I'm not sure that we have test cases atm that do not use invalid pine/sdk arguments but we can always support those in the TS world by using `any` I guess :)

tests/typings/chai-samsam.d.ts
```diff
@@ -0,0 +1,6 @@
+/* tslint:disable:no-namespace */
```
pimterry: I think this means our TSLint settings are wrong. According to https://palantir.github.io/tslint/rules/no-namespace/ we should be setting

```
"no-namespace": [true, "allow-declarations"]
```

That would disallow namespaces in real code, but keep it valid in type definitions like these.

tslint.json
```diff
@@ -25,6 +26,7 @@
     "no-console": [false],
     "no-empty-interface": false,
     "no-string-literal": false,
+    "no-unused-expression-chai": false,
```
pimterry: Doesn't this disable the rule?

thgreasi: Nice catch. While revisiting this I realized that I was actually disabling the `no-unused-expression` rule for every kind of expression.
Removed this and used `const { expect } = m.chai;` to make tslint detect that it's a chai expression.

thgreasi:
That's interesting. Some `before()` calls of some tests are failing with `ResinRequestError: Request error: It is necessary that each organization that has an application1 that has an app name1, has at most one application that has an app name2 that is of the application1.`
cc: @richbayliss

package.json
```diff
@@ -66,16 +68,15 @@
...
     "lint-staged": "^5.0.0",
-    "mocha": "^3.0.0",
-    "mochainon": "^1.0.0",
+    "mocha": "^5.2.0",
+    "mochainon": "^2.0.0",
     "prettier": "^1.8.2",
     "readable-stream": "^2.3.6",
     "require-npm4-to-publish": "^1.0.0",
-    "resin-config-karma": "^1.0.4",
+    "resin-config-karma": "^2.2.0",
```
pimterry: Not strictly necessary, but this would be a nice moment to switch it to [balena-config-karma](npmjs.com/package/balena-config-karma) (it's the same source as 2.3.1, just republished)

thgreasi: Will do :+1:

PR #595, opened 2018-10-19
application.getAll: Use the my_application endpoint

thgreasi:
With this change the application.getAll() will use the my_application 
endpoint instead of defining pine filters. This will also stops it from
returning dependent applications.

Change-type: major
HQ: https://github.com/resin-io/balena/pull/1140


thgreasi:
@resin-ci retest

PR #596, opened 2018-10-20
app.getAppByOwner: Use the slug for filtering

thgreasi:
Also adds typings for the application slug property.

Change-type: minor
Depends-on: resin-io/resin-api#1570


thgreasi:
Should first merge #595

lib/models/application.coffee
```diff
@@ -278,17 +278,10 @@ getApplicationModel = (deps, opts) ->
...
-						user:
-							$any:
-								$alias: 'u',
-								$expr: $eq: [
-									$tolower: $: 'username'
-									owner
-								]
+						$or: [
+							{ slug: "#{owner}/#{appName}" }
+							{ slug: "#{appName}" }
```
pimterry: I think we should only have the first case here. I think there's a good argument that calling `getAppByOwner` in open-source (or using abc/appname in the CLI) should return nothing. You're filtering by owner, and the apps don't have the owner you're asking for (because they have no owner at all).

PR #601, opened 2018-11-05
Make OS version optional during configuration

dfunckt:
Change-type: minor
Depends-on: https://github.com/balena-io/open-balena-api/pull/13


dfunckt:
Turns out we're not going to need this -- version will still be a required argument and clients should still ensure its provided. Closing.

Issue #602, opened 2018-11-15
Update typings to include filters like $ne $tolower 


thgreasi:
See: https://www.flowdock.com/app/rulemotion/resin-frontend/threads/czUGMrohN7nL2s8MNl1rXlnulWh

PR #603, opened 2018-11-15
Extend filter typings to include more comparator functions

thgreasi:
Resolves: #602
Resolves: #608
Change-type: minor
See: https://github.com/balena-io/pinejs-client-js/blob/7e582491765f9037a75754dfa2d0227e3365de33/core.ts#L973-L1036

### DIFF
--- a/typings/pinejs-client-core.d.ts
+++ b/typings/pinejs-client-core.d.ts
@@ -40,7 +40,12 @@ type ResourceObjFilter<T> = {
 	[k in keyof T]?: object | number | string | boolean
 };
 
+type Filter<T> = ResourceObjFilter<T> & FilterExpressions<T>;
+
 interface FilterArray<T> extends Array<Filter<T>> {}
+
+type FilterOperationValue<T> = Filter<T> | FilterArray<T>;
+type FilterFunctionValue<T> = Filter<T> | FilterArray<T>;
 
 type FilterExpressions<T> = {
 	$raw?: RawFilter;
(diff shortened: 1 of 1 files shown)


typings/pinejs-client-core.d.ts
```diff
@@ -40,8 +40,13 @@ type ResourceObjFilter<T> = {
 	[k in keyof T]?: object | number | string | boolean
 };
 
+type Filter<T> = ResourceObjFilter<T> & FilterExpressions<T>;
+
 interface FilterArray<T> extends Array<Filter<T>> {}
 
+type FilterOperationValue<T> = Filter<T> | FilterArray<T>;
+type FilterFunctionValue<T> = Filter<T> | FilterArray<T>;
```
pimterry: These are identical, right? Why do we need two separate types?

thgreasi: @pimterry We don't really but I'm trying to converge to the pinejs client typings, so that it will eventually become easier to bring in updates (until we can have a single source of typings).

thgreasi: See: https://github.com/balena-io/pinejs-client-js/blob/7e582491765f9037a75754dfa2d0227e3365de33/core.ts#L973-L976

pimterry: Ah, gotcha, that makes sense. Might be worth adding a comment here linking there, so we know what to aim for in future. Otherwise all LGTM though!

Issue #606, opened 2018-11-15
Method to disable rolling updates for deployed machines


Lucaszw:
Modified via feedback from PR #607 

It would be nice to have the functionailty fo ```application.disable_rolling_updates``` available in the node.js API.

Something like:
```js
application.disableTrackLatestRelease(nameOrId)
```

https://github.com/balena-io-projects/staged-releases/blob/master/disable-rolling-release-on-fleet.sh

```sh
curl -X PATCH "https://api.$BASE_URL/v4/application($APP_ID)" -H "Authorization: Bearer $authToken" -H "Content-Type: application/json" --data-binary '{"should_track_latest_release":false}'
```

https://github.com/balena-io/balena-sdk-python/blob/master/balena/models/application.py

```python
    def disable_rolling_updates(self, app_id):
        """
        Disable Rolling update on application.
        Args:
            name (str): application id.
        Returns:
            OK/error.
        Raises:
            ApplicationNotFound: if application couldn't be found.
        Examples:
            >> > balena.models.application.disable_rolling_updates('106640')
            'OK'
        """

        params = {
            'filter': 'id',
            'eq': app_id
        }
        data = {
            'should_track_latest_release': False
        }

        return self.base_request.request(
            'application', 'PATCH', params=params, data=data,
            endpoint=self.settings.get('pine_endpoint')
        )
```

Lucaszw:
This looks like it can be done w/ the .update() methods; closing for now

thgreasi:
A way to achieve this was added by https://github.com/balena-io/balena-sdk/pull/572 in v11.1.0 .
You can now achieve this by using `balena.models.application.pinToRelease('MyApp', 'f7caf4ff80114deeaefb7ab4447ad9c661c50847');`

PR #607, opened 2018-11-15
606-disableTrackLatestRelease

Lucaszw:
Resolves: #606 
Change-type: minor
See: https://github.com/balena-io-projects/staged-releases/blob/master/disable-rolling-release-on-fleet.sh

### DIFF
--- a/lib/models/application.coffee
+++ b/lib/models/application.coffee
@@ -875,6 +875,62 @@ getApplicationModel = (deps, opts) ->
 		.asCallback(callback)
 
 	###*
+	# @summary Configure a specific application to stop tracking the latest available release
+	# @name disableTrackLatestRelease
+	# @public
+	# @function
+	# @memberof balena.models.application
+	#
+	# @description The application's current release will not be updated by successfully built releases.
+	#
+	# @param {String|Number} nameOrId - application name (string) or id (number)
+	# @returns {Promise}
+	#
+	# @example
+	# balena.models.application.disableTrackLatestRelease('MyApp').then(function() {
+	# 	...
+	# });
+	#
+	# @example
+	# balena.models.application.disableTrackLatestRelease(123).then(function() {
+	# 	...
+	# });
+	#
+	# @example
+	# balena.models.application.disableTrackLatestRelease('MyApp', function(error) {
+	# 	if (error) throw error;
+	# 	...
+	# });
+	###
+	exports.disableTrackLatestRelease = (nameOrId, callback) ->
+		releaseModel().getLatestByApplication(nameOrId,
+			$select: ['commit', 'belongs_to__application']
+		)
+		.then (latestRelease) ->
+			if not latestRelease
+				return Promise.props(
+					applicationId: getId(nameOrId)
+					commit: null
+				)
+
+			return {
+				applicationId: latestRelease.belongs_to__application.__id
+				commit: latestRelease.commit
+			}
+		.then ({ applicationId, commit }) ->
+			body =
+				should_track_latest_release: false
+
+			if commit
+				body.commit = commit
+
+			return pine.patch
+				resource: 'application'
+				id: applicationId
+				body: body
+		.asCallback(callback)
+
+	###*
 	# @summary Enable device urls for all devices that belong to an application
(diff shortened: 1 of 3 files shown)


pimterry:
Hi @Lucaszw could you tell me a little more about your use case for this?

I think this method was originally omitted because we assumed users would typically want to disable tracking & pin to a specific release, rather than just disabling release tracking in general, and you can already do the former with `application.pinToRelease`. I'd love to know more about what you're doing, and why this extra method is useful to you.

lib/models/application.coffee
```diff
@@ -874,6 +874,62 @@ getApplicationModel = (deps, opts) ->
...
+	# 	...
+	# });
+	#
+	# @example
+	# balena.models.application.disableTrackLatestRelease('MyApp', function(error) {
+	# 	if (error) throw error;
+	# 	...
+	# });
+	###
+	exports.disableTrackLatestRelease = (nameOrId, callback) ->
```
thgreasi: I think that the name of the method doesn't fully describe what it's doing under the hood.
It sounds like it should just pin the application to its current release but the implementation actually pins it to the latest release instead, possibly resulting the whole fleet to download updates if a successful release was published in between two calls of this method.

nazrhom:
@resin-ci retest

Lucaszw:
We are looking to have the capability to stop a fleet of deployed machines (on different release commits) from automatically updating on new releases. 

With additional consideration, we are more looking to add "disableRollingUpdates" from the python API to the node.js API (see: https://www.balena.io/docs/reference/sdk/python-sdk/#function-disable_rolling_updates-app_id)

I.e. Add this function to the Node.js Library.

https://github.com/balena-io/balena-sdk-python/blob/master/balena/models/application.py
```python

 def disable_rolling_updates(self, app_id):
        """
        Disable Rolling update on application.
        Args:
            name (str): application id.
        Returns:
            OK/error.
        Raises:
            ApplicationNotFound: if application couldn't be found.
        Examples:
            >> > balena.models.application.disable_rolling_updates('106640')
            'OK'
        """

        params = {
            'filter': 'id',
            'eq': app_id
        }
        data = {
            'should_track_latest_release': False
        }

        return self.base_request.request(
            'application', 'PATCH', params=params, data=data,
            endpoint=self.settings.get('pine_endpoint')
        )

```

I will close this PR, and create a new one.

thgreasi:
> a fleet of deployed machines (on different release commits) 

I guess that some of those devices are pinned to specific releases, is that correct?
Changing the app's tracked release only affects those that are not pinned and the pinned ones should be unaffected (regardless of the app's `commit` or `should_track_latest_release`).

That could be done with the current methods by using:
```
const currentReleaseHash = await balena.models.application.getTargetReleaseHash('myApp');
balena.models.application.pinToRelease('myApp', currentReleaseHash);
```
On the other hand of course you can avoid working with the release hash by making the method to be a simple PATCH, like the one you mentioned in #606 (or similar to what the python sdk does).
I like the idea of having a separate `pinToCurrentRelease()` method that normalized the app identification parameter and then does the aforementioned PATCH.

Lucaszw:
> I guess that some of those devices are pinned to specific releases, is that correct?

That is correct.
We have many deployed machines on different releases, and we want every machine under the current app (regardless of the hash) to be "pinned" to the current release on the device. 

I.e. Maybe something more along the lines of ```application.pinAll(appId)``` that pins each device to its current release, not the current the release on the app level.

I.e. functionality that would have the same outcome of the following (js pseudo code)
```javascript
model.device.getAllByApplication().then((devices) => {
   model.devices.forEach((device) => {
     model.device.pinToRelease(device.id, device.commit);
  });
})

```

thgreasi:
@Lucaszw Your last message confused me a bit.
This PR was about a method in the application model and later you mentioned that you were interested at having the python SDK's `disable_rolling_updates()` method in the JS sdk.
On the other hand the provided pseudocode is different from the other implementations discussed earlier and just each individual device of an application (to what it's currently running).

What's the use case of that? Is this for example so that your fleet only downloads updates in specific time frames?
Is this something you need to do often?

Issue #610, opened 2018-12-07
Should be able to fetch devices with service details


karaxuna:
See: https://github.com/balena-io/resin-ui/pull/2490

PR #611, opened 2018-12-07
sdk.models.application: getAllWithDeviceServiceDetails method added

karaxuna:
Resolves: #610
Change-type: minor

### DIFF
--- a/lib/models/application.coffee
+++ b/lib/models/application.coffee
@@ -131,6 +131,51 @@ getApplicationModel = (deps, opts) ->
 		.asCallback(callback)
 
 	###*
+	# @summary Get applications and their devices, along with each device's
+	# associated services' essential details
+	# @name getAllWithDeviceServiceDetails
+	# @public
+	# @function
+	# @memberof balena.models.application
+	#
+	# @description
+	# This method does not map exactly to the underlying model: it runs a
+	# larger prebuilt query, and reformats it into an easy to use and
+	# understand format. If you want more control, or to see the raw model
+	# directly, use `application.getAll(options)` instead.
+	#
+	# @param {Object} [options={}] - extra pine options to use
+	# @fulfil {Object[]} - applications
+	# @returns {Promise}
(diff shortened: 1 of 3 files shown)


lib/models/application.coffee
```diff
@@ -130,6 +130,51 @@ getApplicationModel = (deps, opts) ->
 
 		.asCallback(callback)
 
+	###*
+	# @summary Get applications and their deives, along with each device's
```
sradevski: NIT: Misspelled word, "their deives" -> "their devices".

karaxuna:
@thgreasi can we create another issue for these tests? This issue is blocking me on https://github.com/balena-io/resin-ui/issues/2422 and the last one blocks me on https://github.com/balena-io/resin-ui/pull/2490. Writing tests for this issue will take some time for me considering the fact that I'm not that familiar with sdk repo.

thgreasi:
@karaxuna You should be able to start the balena sdk upgrade in the UI w/o this PR.
Perhaps you can write the tests for this while we will be reviewing the sdk upgrade PR.
If the tests end up blocking you, then we can probably open an SDK for them and merge this a bit sooner, as far as the new methods works fine in the UI PR.

lib/models/application.coffee
```diff
@@ -130,6 +130,51 @@ getApplicationModel = (deps, opts) ->
...
+
+		serviceOptions = mergePineOptions
+			$expand: [
+				owns__device: getCurrentServiceDetailsPineOptions()
+			]
+		, options
+
+		exports.getAll(serviceOptions)
+		.then (apps) ->
+			return apps.map (app) ->
```
thgreasi: NIT: we probably don't need to use `.map` here. Iterating with `forEach` to set the `owns__device` property on the existing objects and returning the `apps` should be fine and would save an unnecessary array creation.

karaxuna:
@resin-ci retest

PR #617, opened 2019-01-12
Lazy-load the sdk code, so only setting shared options can be faster

Page-:
Change-type: patch


lib/util/index.coffee
```diff
@@ -15,15 +15,10 @@ omit = require('lodash/omit')
...
 
 # Use window (web)/self (web worker)/global (node) as appropriate
-exports.globalEnv = if typeof window != 'undefined'
-	window
-else if typeof self != 'undefined'
-	self
-else if typeof global != 'undefined'
-	global
-else null # If we can't guarantee global state, don't fake it: fail instead.
+exports.globalEnv = globalEnv
```
thgreasi: Since the only other place that this get's imported is `lib/logs.coffee`, how about changing the import there and completely dropping this re-export?

Page-: done

PR #618, opened 2019-01-14
Adds peer dependencies of resin-register-device

joelgallant:


thgreasi:
Hi,
The underlying issue was properly resolved by #662 in v11.13.0 (by updating some dependencies) and this shouldn't be an issue any more.
Thanks for bringing this up and contributing to this module :+1:

Issue #539, opened 2018-07-10
Should support the v6 endpoint (Orgs)


thgreasi:
Placeholder Issue
HQ: https://github.com/resin-io/hq/pull/887
Depends-On: https://github.com/resin-io/resin-api/pull/1079

PR #621, opened 2019-01-15
**BREAKING** Add support for the v6 organizations model

thgreasi:
Resolves: #539
Depends-on: https://github.com/balena-io/resin-api/pull/2018
Change-type: major


thgreasi:
@balena-ci retest

lib/auth.coffee
```diff
@@ -281,6 +281,52 @@ getAuth = (deps, opts) ->
...
+	# 	console.log(organizationId);
+	# });
+	#
+	# @example
+	# balena.auth.getPersonalOrganizationId(function(error, organizationId) {
+	# 	if (error) throw error;
+	# 	console.log(organizationId);
+	# });
+	###
+	exports.getPersonalOrganizationId = (callback) ->
```
sradevski: Should we have a `getPersonalOrganization` method that accepts a `$select` instead, so the caller can decide which fields they need (in this case just ID)? This applies to all methods we expose from the SDK, just thinking which approach would be better here, no action required. We can then have some commonly used cases (like `getPersonalOrganizationId` exposed as methods that simply utilize this more general `getPersonalOrganization`.

thgreasi: @sradevski  I was thinking about this, since I we also need this in the UI, but that doesn't sound like that should be part of the `.auth` namespace.
If we are going to add it, then it probably deserves a proper namespace in `models`, which I'm not sure we should do, since it will not work with the open-balena-api :thinking:
On the other hand the `getPersonalOrganizationId()` just returns an id which allows us to safely cache it along with the user details.
How would you feel about having such a method in the `.auth` namespace?

sradevski: @thgreasi I agree that it doesn't belong in the `auth` namespace, I was thinking more like an `organization` model that `getPersonalOrganization` and `getPersonalOrganizationId` that the `auth` namespace can use. Anyway, just thinking out loud, I am fine without doing that for this PR

thgreasi: Always glad to hear your thoughts on these :)
We can always add it later if it ends up being a crucial method.

thgreasi:
@balena-ci rebase

lib/models/application.js
```diff
@@ -372,20 +380,20 @@ const getApplicationModel = function (deps, opts) {
 		},
 
 		/**
-		 * @summary Get a single application using the appname and owner's username
+		 * @summary Get a single application using the appname and the handle of the owner organization
```
dfunckt, line 383:
```suggestion
		 * @summary Get a single application using the appname and the handle of the owning organization
```

lib/models/application.js
```diff
@@ -372,20 +380,20 @@ const getApplicationModel = function (deps, opts) {
...
 		 * @name getAppByOwner
 		 * @public
 		 * @function
 		 * @memberof balena.models.application
 		 *
 		 * @param {String} appName - application name
-		 * @param {String} owner - The owner's username
+		 * @param {String} owner - The handle of the owner organization
```
dfunckt, line 390:
```suggestion
		 * @param {String} owner - The handle of the owning organization
```

lib/models/application.js
```diff
@@ -492,6 +500,7 @@ const getApplicationModel = function (deps, opts) {
 		 * @param {String} [options.applicationType] - application type slug e.g. microservices-starter
 		 * @param {String} options.deviceType - device type slug
 		 * @param {(Number|String)} [options.parent] - parent application name or id
+		 * @param {(String|Number|null)} [options.organization] - handle (string) or id (number) of the organization that the application will belong to or null
```
dfunckt, line 503: `organization` can't be `null`.

lib/models/application.js
```diff
@@ -537,46 +546,88 @@ const getApplicationModel = function (deps, opts) {
...
+					});
+				})
+				.then(function ([dt]) {
+					if (dt == null) {
+						throw new errors.BalenaInvalidDeviceType(deviceType);
+					}
+					return dt.id;
 				});
 
+			const organizationPromise = !organization
```
dfunckt, line 578: `organization` must not be `null`.

lib/models/organization.ts
```diff
@@ -0,0 +1,216 @@
...
+	isNotFoundResponse,
+	mergePineOptions,
+	treatAsMissingOrganization,
+} from '../util';
+
+const getOrganizationModel = function (deps: InjectedDependenciesParam) {
+	const { pine } = deps;
+
+	const getId = (handleOrId: string | number) =>
+		get(handleOrId, { $select: 'id' }).then(({ id }) => id);
```
dfunckt, line 33: Using async/await on new code is best, but I guess you're limited by the ES version?

lib/models/organization.ts
```diff
@@ -0,0 +1,216 @@
...
+	 * 	console.log(organization);
+	 * });
+	 */
+	const get = function (
+		handleOrId: string | number,
+		options: BalenaSdk.PineOptions<BalenaSdk.Organization> = {},
+	) {
+		return Promise.try(() => {
+			if (handleOrId == null) {
+				throw new errors.BalenaOrganizationNotFound(handleOrId);
```
dfunckt, line 141: Is there a more appropriate error to throw here? `handleOrId` being `null` is more like an "invalid argument" or "bad request" kind of error.

lib/models/application.js
thgreasi: I actually just checked today the API source for this
See: https://github.com/balena-io/balena-api/blob/b84ec51f8c20896c037297ab7d623a7bfb9c998b/src/models/hooks/resources/application.ts#L84-L102

but it's best to anyway start stricter :+1:

lib/models/organization.ts
thgreasi: The concern is that the transpiled code size will greatly increase after the async-degenerator code gets added.
We should though re-evaluate the cost of this vs the convenience, but I would avoid doing this now.

lib/models/organization.ts
thgreasi: Indeed, BalenaInvalidParameterError is the way to go :+1:

lib/models/application.js
dfunckt: But we've gone over this several times in calls and whatnot and agreed it had to be an argument that is explicitly provided. Do I misremember? That hook will also have to change but that's unrelated.

lib/models/organization.ts
dfunckt: Yeah, agreed, it only makes sense to switch when the minimum supported Node version has full async/await support.

lib/balena.js
```diff
@@ -122,6 +122,12 @@ const getSdk = function (opts) {
 		}
 	}
 
+	if ('apiKey' in opts) {
```
sradevski, line 125: Can you remind me why we still need to support this?

lib/models/application-invite.ts
```diff
@@ -164,7 +165,7 @@ const getApplicationInviteModel = function (
 								$top: 1,
 								$select: ['id'],
 								$filter: {
-									name: roleName,
+									name: roleName as ApplicationMembershipRoles,
```
sradevski, line 168: Shouldn't we add this typing to `ApplicationInviteOptions` instead? I see it's destructured from there.

lib/models/application.js
```diff
@@ -1034,36 +1114,31 @@ const getApplicationModel = function (deps, opts) {
...
+							$top: 1,
+							$filter: {
+								status: 'success',
+							},
+							$orderby: 'created_at desc',
+						},
+					},
 				})
-				.then(function ({ applicationId, commit }) {
+				.then(function (application) {
```
sradevski, line 1131: NIT: Convert to arrow function

lib/models/organization.ts
```diff
@@ -0,0 +1,216 @@
...
+		get(handleOrId, { $select: 'id' }).then(({ id }) => id);
+
+	/**
+	 * @summary Creates a new organization
+	 * @name create
+	 * @public
+	 * @function
+	 * @memberof balena.models.organization
+	 *
+	 * @description This method creates a new organization with current user as an administrator.
```
sradevski, line 42:
```suggestion
	 * @description This method creates a new organization with the current user as an administrator.
```

lib/models/organization.ts
```diff
@@ -0,0 +1,216 @@
...
+						resource: 'organization',
+						id: handleOrId,
+						options: mergePineOptions({}, options),
+					})
+					.tap((organization) => {
+						if (organization == null) {
+							throw new errors.BalenaOrganizationNotFound(handleOrId);
+						}
+					});
+			} else {
```
sradevski, line 155: The else is redundant here

lib/util/dependent-resource.js
```diff
@@ -102,26 +99,27 @@ export function buildDependentResource(
...
-					// If we never checked that the resource actually exists
-					// then we should reject an appropriate error.
-					if (!isId(parentParam)) {
-						return;
-					}
-					return getResourceId(parentParam);
-				}),
+					})
+					.tapCatch(isUnauthorizedResponse, function () {
+						// On Pine 7, when the post throws a 401
```
sradevski, line 114: is this comment still valid?

lib/models/application.js
thgreasi: No no, you are right, nice catch.

lib/balena.js
thgreasi: Our dependencies still support using apiKeys since they are more generic than the SDK.
Since the SDK just passes the options to balena-request & balena-pine if we don't explicitly remove it, it will still reach downstream and keep working.
This way our users will notice that the `apiKey` doesn't work and will have to switch to `loginWIthToken`.

sradevski: Oh ok, but we should remove the option from balena-request and balena-pine as well and just force users to move to using headers

lib/util/dependent-resource.js
thgreasi: Yes it is.
Let me update the comment.

Issue #622, opened 2019-01-16
Should support tags on the open-balena-api


thgreasi:
See: https://github.com/balena-io/open-balena-api/pull/45

PR #623, opened 2019-01-16
Support tags on the open-balena-api

thgreasi:
Properly detects unique violation errors so that upserts
wort both on balenaCloud & open-balena-api / plain pine
instances. Adds support for tags on the open-balena-api
and fixes a device service envvar issue.

Resolves: #622
See: https://github.com/balena-io/open-balena-api/pull/45
Change-type: minor

### DIFF
--- a/lib/util/index.coffee
+++ b/lib/util/index.coffee
@@ -84,9 +84,14 @@ exports.noApplicationForKeyResponse =
 	statusCode: 500
 	body: 'No application found to associate with the api key'
 
-exports.uniqueKeyViolated =
-	code: 'BalenaRequestError'
-	body: 'Unique key constraint violated'
+exports.isUniqueKeyViolationResponse = ({code, body}) ->
+	code == 'BalenaRequestError' &&
+	(
+		# api translated response
+		body == 'Unique key constraint violated' ||
+		# pine response (tested on pine 10)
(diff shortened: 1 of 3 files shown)


lib/util/index.coffee
```diff
@@ -84,9 +84,14 @@ exports.noApplicationForKeyResponse =
...
-exports.uniqueKeyViolated =
-	code: 'BalenaRequestError'
-	body: 'Unique key constraint violated'
+exports.isUniqueKeyViolationResponse = ({code, body}) ->
+	code == 'BalenaRequestError' &&
+	(
+		# api translated response
+		body == 'Unique key constraint violated' ||
+		# pine response (tested on pine 10)
+		/^".*" must be unique\.$/.test(body)
```
dfunckt: This is good enough as it is (I mean the “.*” bit), but you could also provide as an argument the actual field names you expect constraints to fail and make it a lot more robust. Cloud API needs to translate constraint violations out of bounds, hence the need to cast such a wide net with the regular expression.

thgreasi: Maybe we can re-iterate on this on the orgs sdk, since the v6 endpoint will probably no longer translate the pine error to `Unique key constraint violated`.

dfunckt: I see. Yeah, it’s good as it is :)

PR #629, opened 2019-01-16
util: Create a generic upsert pine helper method

thgreasi:
That's on top of #628 so that the diff is cleaner and is planned to end-up in the next major sdk version (v12).
This might better at some point end up in PineClient (if @Page- likes it).
For now, since we rely on an API translation for the detection of the unique complex key violation, we better make it part of the SDK, which is aware of both plain pine behavior (as in the OSS API) and the cloud API one w/ the translation.

Resolves: #627
Change-type: minor

### DIFF
--- a/lib/models/device.coffee
+++ b/lib/models/device.coffee
@@ -40,7 +40,6 @@ deviceStatus = require('resin-device-status')
 	treatAsMissingDevice,
 	LOCKED_STATUS_CODE,
 	timeSince,
-	isUniqueKeyViolationResponse
 } = require('../util')
 { normalizeDeviceOsVersion } = require('../util/device-os-version')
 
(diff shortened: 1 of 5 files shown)


lib/models/device.coffee
```diff
@@ -2779,21 +2779,17 @@ getDeviceModel = (deps, opts) ->
...
-							$filter:
-								service_install: serviceInstallId
-								name: key
-						body:
-							value: value
+				,
+				[
+					'service_install'
+					'name'
+				]
```
thgreasi: It would be nice if these could be part of the pine params, but I thought I should better be sure that we are not going to cause any prop name conflicts w/ PineClient. We should do so though if this is going to be part of PineClient.

thgreasi:
@sradevski with unit tests, do you mean offline mocked tests or just having a test in the utils test cases like `describe 'upsert' () -> it 'should work'` that checks upserting on a specific resource?
I have the second in mind, since I would like to start exposing upsert, along with some of the rest utility methods that we use in the sdk. Would you be fine with it if add such upsert e2e tests at that point?

PR #630, opened 2019-01-18
Upgrade Typescript to v3.0 & improve typing for post & patch request body

thgreasi:
Resolves: #480
Resolves: #481
Change-type: major


package.json
```diff
@@ -76,7 +76,7 @@
     "lint-staged": "^5.0.0",
     "mocha": "^3.0.0",
     "mochainon": "^2.0.0",
-    "prettier": "^1.8.2",
+    "prettier": "^1.11.1",
```
JackuB: Just wondering: is there any reason we are not using package-lock in this project as well?

typings/pinejs-client-core.d.ts
```diff
@@ -114,30 +117,53 @@ type BaseExpandFor<T> = { [k in keyof T]?: object } | StringKeyof<T>;
...
+	apiPrefix?: string;
+	method?: ODataMethod;
+	resource?: string;
 	id?: number;
-	body?: object;
+	url?: string;
+	passthrough?: AnyObject;
+	passthroughByMethod?: {
+		GET: AnyObject;
+		POST: AnyObject;
```
JackuB: No PUT?

typings/pinejs-client-core.d.ts
```diff
@@ -114,30 +117,53 @@ type BaseExpandFor<T> = { [k in keyof T]?: object } | StringKeyof<T>;
...
-	$top?: number;
-	$skip?: number;
 }
 
-export interface PineOptionsFor<T> extends PineOptions {
+export interface PineOptionsFor<T> extends ODataOptionsBase {
 	$select?: Array<StringKeyof<T>> | StringKeyof<T> | '*';
 	$filter?: Filter<T>;
-	$expand?: Expand<T>;
+	$expand?: Expand<T> | string;
```
sradevski: Should this be part of the `export type Expand<T>....` defined above?

tslint.json
```diff
@@ -5,7 +5,7 @@
     "tslint-no-unused-expression-chai"
   ],
   "rules": {
-    "indent": [ true, "tabs" ],
+    "indent": [ false, "tabs" ],
```
sradevski: NIT: It is probably a good idea to move to `resin-lint` instead, but it doesn't have to be part of this PR if you don't have the time.

package.json
thgreasi: We shouldn't use lock files on most libraries, since the lock files do not reach our users & we will not be testing against the actual node modules that our consumers get with an `npm i balena-sdk`. For example, let's say that `memoizee` 0.4.10 regresses and our lock file is on 0.4.9, b/c of the package.json being on "^0.4.9", we will never notice that the sdk is broken for our consumers.
If there were "devDependencies only" lock files, we could use them though to make our ci more predictable.

typings/pinejs-client-core.d.ts
thgreasi: I'm not sure, I just got that from pinejs-client
See: https://github.com/balena-io/pinejs-client-js/blob/7e582491765f9037a75754dfa2d0227e3365de33/core.ts#L1089

typings/pinejs-client-core.d.ts
thgreasi: Oh actually it seems that this is a leftover from the rebase, since we now have `type BaseExpandFor<T> = { [k in keyof T]?: object } | StringKeyof<T>;`.
Nice heads-up @sradevski

package.json
JackuB: :+1: true, haven't realised this is a library. Maybe the strategy is to keep a lockfile for our CI (since we would need to rerun our CI for every deep dependency update anyway), don't publish it and keep transitive dependencies in check with some tool like greenkeeper

typings/pinejs-client-core.d.ts
thgreasi: Updated

typings/pinejs-client-core.d.ts
thgreasi: @JackuB checked up on this with Page and it ends up that PUT should indeed be there and as a result [this PR was opened](https://github.com/balena-io/pinejs-client-js/pull/69).
Mirrored those changes in here.

thgreasi:
I've changed the base branch of this to be the `v12.0.0-organizations-dev` feature branch. Just like we did for past major releases, I'm planning to:
* start merging small PRs in there
* open a WIP/DONOTMERGE `v12.0.0-organizations-dev` - `master` PR
  * which will make resinCI publish pre-release packages whenever a PR gets merged to the feature branch,which we can then link to other components.
* once the /v5 endpoint of the API gets released to prod, the `v12.0.0-organizations-dev` branch tests should start passing again. We should then rebase it, and once all the tests pass and everything for v12 milestone is there, we can then and only then merge `v12.0.0-organizations-dev` to `master`.

PR #632, opened 2019-01-28
Import lodash methods independently to reduce bundle size

### LINKED ISSUE #631: Should import lodash methods independently to reduce bundle size
Connects-to: #250

thgreasi:
Also runs newest prettier on the sources so that the tests still pass on CI. Added those changes in a separate commit to make it easier to review & track.

Resolves #631
Change-type: patch


thgreasi:
Impressively, the unminified bundle size got reduced by more than a MB, I guess lodash has a ton of doc comments in there.

Issue #635, opened 2019-02-03
Should provide a method to override device update locks


thgreasi:
See: https://github.com/balena-io/resin-ui/blob/0e0a5c18642b772f3d03e1193731bb87bb6cc4ca/src/services/models/device-model-service.ts#L132-L139

thgreasi:
Resolved by https://github.com/balena-io/balena-sdk/pull/636

PR #636, opened 2019-02-03
Device: Set update lock

karaxuna:
Connects-to: #635
Change-type: minor

### DIFF
--- a/lib/models/device.coffee
+++ b/lib/models/device.coffee
@@ -1760,6 +1760,96 @@ getDeviceModel = (deps, opts) ->
 						uuid: uuid
 		.asCallback(callback)
 
+	setLockOverriden = (uuidOrId, shouldOverride, callback) ->
+		getId(uuidOrId).then (deviceId) ->
+			value = if shouldOverride then '1' else '0'
+			return request.send
+				method: 'POST'
+				url: "/device/#{deviceId}/lock-override"
+				baseUrl: apiUrl
+				body:
+					value: value
+		.asCallback(callback)
+
+	###*
+	# @summary Enable lock override
+	# @name enableLockOverride
+	# @public
+	# @function
+	# @memberof balena.models.device
+	#
+	# @param {String|Number} uuidOrId - device uuid (string) or id (number)
+	# @returns {Promise}
+	#
+	# @example
+	# balena.models.device.enableLockOverride('7cf02a6');
+	#
+	# @example
+	# balena.models.device.enableLockOverride(123);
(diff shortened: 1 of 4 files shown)


lib/models/device.coffee
```diff
@@ -1760,6 +1760,70 @@ getDeviceModel = (deps, opts) ->
...
+	#
+	# @example
+	# balena.models.device.setLockOverriden(123, true);
+	#
+	# @example
+	# balena.models.device.setLockOverriden('7cf02a6', true, function(error) {
+	# 	if (error) throw error;
+	# });
+	###
+	exports.setLockOverriden = (uuidOrId, shouldOverride, callback) ->
```
sradevski: I personally find methods like `enableLockOverride` and `disableLockOverride` more intuitive to use and more difficult to call with the wrong argument, so maybe we can name it like that? The `setLockOverriden` one can be private and stay pretty much the same (except maybe pass `'1'` and `'0'`) and be called by the two methods I mentioned above.

lib/models/device.coffee
```diff
@@ -1760,6 +1760,96 @@ getDeviceModel = (deps, opts) ->
...
+	#
+	# @example
+	# balena.models.device.isLockOverriden(123);
+	#
+	# @example
+	# balena.models.device.isLockOverriden('7cf02a6', function(error) {
+	# 	if (error) throw error;
+	# });
+	###
+	exports.isLockOverriden = (uuidOrId, callback) ->
```
thgreasi: How about naming this as `hasLockOverride` similarly to `hasDeviceUrl`?

typings/balena-sdk.d.ts
```diff
@@ -839,6 +845,9 @@ declare namespace BalenaSdk {
 					expiryTimestamp: number,
 				): Promise<void>;
 				revokeSupportAccess(uuidOrId: string | number): Promise<void>;
+				enableLockOverride(uuidOrId: string | number): Promise<void>;
+				disableLockOverride(uuidOrId: string | number): Promise<void>;
+				isLockOverriden(uuidOrId: string | number): Promise<boolean>;
```
thgreasi: Reminder
> How about naming this as `hasLockOverride` similarly to `hasDeviceUrl`?

thgreasi:
Let's also title the commit with something like `Add device lock override methods`.

thgreasi:
The tests for the new methods do pass but the test suit times out on both travis & appveyor.
Merging as an admin.

Issue #578, opened 2018-09-22
Should be able to trigger host OS updates (hup)


thgreasi:
HQ: https://github.com/resin-io/hq/issues/1215
See: https://www.flowdock.com/app/rulemotion/resin-frontend/threads/sYrLPINqwDMg1t5Zb9-lcW9IAWi?filter=search&query=hup
See: https://www.flowdock.com/app/rulemotion/r-architecture/threads/lukg_hxfLkIRON5-TTAhmycEvxk
See: https://www.flowdock.com/app/rulemotion/resin-frontend/threads/m90nwXCsWvXikh2fHSM_uZToGoa

PR #638, opened 2019-02-22
Add OS update methods

thgreasi:
Resolves: #578
HQ: https://github.com/resin-io/hq/issues/1215
Depends-on: https://github.com/balena-io-modules/balena-hup-action-utils/pull/1
Change-type: minor

### DIFF
--- a/lib/models/device.coffee
+++ b/lib/models/device.coffee
@@ -24,25 +24,31 @@ find = require('lodash/find')
 some = require('lodash/some')
 includes = require('lodash/includes')
 map = require('lodash/map')
+rSemver = require('resin-semver')
 semver = require('semver')
 errors = require('balena-errors')
 deviceStatus = require('resin-device-status')
 
 {
-	onlyIf,
-	isId,
-	findCallback,
-	getCurrentServiceDetailsPineOptions,
-	generateCurrentServiceDetails,
-	mergePineOptions,
-	notFoundResponse,
-	noDeviceForKeyResponse,
-	treatAsMissingDevice,
-	LOCKED_STATUS_CODE,
-	timeSince,
+	onlyIf
+	isId
+	findCallback
+	getCurrentServiceDetailsPineOptions
+	getOsUpdateHelper: _getOsUpdateHelper
+	generateCurrentServiceDetails
+	mergePineOptions
+	notFoundResponse
+	noDeviceForKeyResponse
+	treatAsMissingDevice
+	LOCKED_STATUS_CODE
+	timeSince
 	isUniqueKeyViolationResponse
 } = require('../util')
-{ normalizeDeviceOsVersion } = require('../util/device-os-version')
+{ hupActionHelper } = require('../util/device-actions/os-update/utils')
+{
+	getDeviceOsSemverWithVariant
+	normalizeDeviceOsVersion
+} = require('../util/device-os-version')
 
 # The min version where /apps API endpoints are implemented is 1.8.0 but we'll
 # be accepting >= 1.8.0-alpha.0 instead. This is a workaround for a published 1.8.0-p1
@@ -60,11 +66,12 @@ CONTAINER_ACTION_ENDPOINT_TIMEOUT = 50000
 
 getDeviceModel = (deps, opts) ->
 	{ pine, request } = deps
-	{ apiUrl, dashboardUrl } = opts
+	{ apiUrl, dashboardUrl, deviceUrlsBase } = opts
 
 	registerDevice = require('resin-register-device')({ request })
 	configModel = once -> require('./config')(deps, opts)
 	applicationModel = once -> require('./application')(deps, opts)
+	osModel = once -> require('./os')(deps, opts)
 	auth = require('../auth')(deps, opts)
 
 	{ buildDependentResource } = require('../util/dependent-resource')
@@ -98,6 +105,15 @@ getDeviceModel = (deps, opts) ->
 	# Infer dashboardUrl from apiUrl if former is undefined
 	if not dashboardUrl?
 		dashboardUrl = apiUrl.replace(/api/, 'dashboard')
+
+	getDeviceUrlsBase = once Promise.method ->
+		if deviceUrlsBase?
+			return deviceUrlsBase
+		return configModel().getAll().get('deviceUrlsBase')
+
+	getOsUpdateHelper = once ->
+		getDeviceUrlsBase().then (deviceUrlsBase) ->
+			_getOsUpdateHelper(deviceUrlsBase, request)
 
 	# Internal method for uuid/id disambiguation
 	# Note that this throws an exception for missing uuids, but not missing ids
@@ -1691,7 +1707,7 @@ getDeviceModel = (deps, opts) ->
 			if not hasDeviceUrl
 				throw new Error("Device is not web accessible: #{uuidOrId}")
 
-			configModel().getAll().get('deviceUrlsBase').then (deviceUrlsBase) ->
+			getDeviceUrlsBase().then (deviceUrlsBase) ->
 				exports.get(uuidOrId, $select: 'uuid').get('uuid').then (uuid) ->
 					return "https://#{uuid}.#{deviceUrlsBase}"
 		.asCallback(callback)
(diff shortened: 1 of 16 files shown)


lib/models/device.coffee
```diff
@@ -60,14 +61,14 @@ CONTAINER_ACTION_ENDPOINT_TIMEOUT = 50000
...
-	{ apiUrl, dashboardUrl } = opts
+	{ apiUrl, dashboardUrl, deviceUrlsBase } = opts
 
 	registerDevice = require('resin-register-device')({ request })
 	configModel = once -> require('./config')(deps, opts)
 	applicationModel = once -> require('./application')(deps, opts)
 	auth = require('../auth')(deps, opts)
 
-	{ buildDependentResource } = require('../util/dependent-resource')
+	{ buildDependentResource } = requiredre('../util/dependent-resource')
```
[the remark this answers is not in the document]
thgreasi: Oh I though that WIP PRs do not send emails...
Sorry about that, I will add some tests and re-ping you afterwards.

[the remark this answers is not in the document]
thgreasi: And this is obviously a typo :)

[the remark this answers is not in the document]
sradevski: @thgreasi I assumed you have tested it therefore it couldn't be just a typo :smile:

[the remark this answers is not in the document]
thgreasi: @sradevski I actually thought about pushing and trying opening a draft PR to see how that works, so at that point it was still a WIP that I wanted a placeholder PR for.
Sorry for the noise, I thought that they were not supposed to send reviewer emails while this is in a draft.

typings/utils.d.ts
```diff
@@ -0,0 +1,8 @@
+// Functionally the same as Exclude, but for strings only.
+type Diff<T extends string, U extends string> = ({ [P in T]: P } &
```
Page-: I'm not quite sure what this `Diff` is doing, but I think you'd be better off with:
```ts
export type ExcludeField<T, F extends keyof T> = Pick<T, Exclude<keyof T, F>>;
```
where I presume `Diff` is doing the equivalent of `Exclude<keyof T, F>`?

typings/balena-sdk.d.ts
```diff
@@ -571,26 +577,30 @@ declare namespace BalenaSdk {
 		release: NavigationResource<Release>;
 	}
 
+	// See: https://github.com/balena-io/resin-proxy/issues/51#issuecomment-274251469
+	interface OsUpdateActionResult {
+		status: 'idle' | 'in_progress' | 'update_done' | 'error';
```
Page-: These are changing/have changed so it'd be good to stick with the newer form (or accept both), cc @xginn8

lib/models/os.coffee
```diff
@@ -384,6 +392,86 @@ getOsModel = (deps, opts) ->
...
+			current = find(allVersions, (v) -> rSemver.compare(v, currentVersion) == 0) ||
+				currentVersion
+
+			versions = allVersions.filter (v) ->
+				# avoid the extra call to validateDeviceType,
+				# since getSupportedVersions is already does that
+				hupActionHelper.isSupportedOsUpdate(deviceType, currentVersion, v)
+
+			recommended = first(
+				reject(versions, (v) -> rSemver.prerelease(v))
```
Page-: Can this be
```suggestion
				reject(versions, rSemver.prerelease)
```
?

lib/models/device.coffee
```diff
@@ -2170,6 +2180,87 @@ getDeviceModel = (deps, opts) ->
...
+			osUpdateHelper.getOsUpdateStatus(uuid)
+		.catch (err) ->
+			if err.statusCode != 400
+				throw err
+
+			# as an attempt to reduce the requests for this method
+			# check whether the device exists only when the request rejects
+			# so that it's rejected with the appropriate BalenaDeviceNotFound error
+			return exports.get(uuid, $select: 'id')
+			.then ->
```
Page-:
```suggestion
			.throw(err)
```

typings/utils.d.ts
thgreasi: The SDK is still using TS 2.6.2 unfortunately, where `Exclude` isn't available, so I had to use a backwards compatible implementation that is based on mapped types.
Bumping TS w/o a major version can cause CI tests to fails on consumer projects that haven't caught up yet with newer TS versions.
On the other hand the `v12-dev` branch (the next major) is already updated to use `Exclude` as you suggest.
See: https://github.com/balena-io/balena-sdk/blob/v12-dev/typings/utils.d.ts

PS: `Diff` was the most popular community typing helper afaict for doing this, so I also used the same name for reference.

typings/balena-sdk.d.ts
thgreasi: I saw the UI PR after I opened this one.
Will update the typings here, thanks for the reminder :+1:

lib/models/device.coffee
thgreasi: At last the ideal place to use this :)

lib/models/device.coffee
```diff
@@ -99,6 +100,15 @@ getDeviceModel = (deps, opts) ->
 	if not dashboardUrl?
 		dashboardUrl = apiUrl.replace(/api/, 'dashboard')
 
+	getDeviceUrlsBase = once ->
+		if deviceUrlsBase?
+			return deviceUrlsBase
```
sradevski: Is this also a Promise? Otherwise I assume it will throw when you call `.then` on it.

package.json
```diff
@@ -100,6 +100,7 @@
     "abortcontroller-polyfill": "^1.1.9",
     "balena-auth": "^3.0.0",
     "balena-errors": "^4.0.0",
+    "balena-hup-action-utils": "1.0.0-initial-code-737718128eb1b664f5d161d182052f4400b3f344",
```
sradevski: Don't forget to point to the released version if that gets released before this PR.

typings/balena-sdk.d.ts
```diff
@@ -571,26 +577,31 @@ declare namespace BalenaSdk {
 		release: NavigationResource<Release>;
 	}
 
+	// See: https://github.com/balena-io/resin-proxy/issues/51#issuecomment-274251469
+	interface OsUpdateActionResult {
+		// 'update_done' will be obsolete with https://github.com/balena-io/resin-proxy/pull/334
+		status: 'idle' | 'in_progress' | 'update_done' | 'resinhup_done';
```
sradevski: I think `resinhup_done` was changed to just `done` judging by a PR in the dashboard.

lib/models/device.coffee
thgreasi: Oh nice catch @sradevski :+1: 
This was defenitelly going to cause issues if someone used the new constructor parameter!
Will update that shortly :+1:

typings/balena-sdk.d.ts
thgreasi: Yea I also just saw that new change in the respective Proxy PR.
Will update this as well, thanks for the heads up :+1:

thgreasi:
@resin-ci retest

lib/models/device.coffee
imrehg: So this is still pending? Or was resolved in an interation?

lib/models/device.coffee
```diff
@@ -2170,6 +2186,140 @@ getDeviceModel = (deps, opts) ->
...
+	# @param {String} targetOsVersion - semver-compatible version for the target device
+	# @throws Exception if update isn't supported
+	# @returns {undefined}
+	###
+	exports._checkOsUpdateTarget = ({ uuid, device_type, is_online, os_version, os_variant }, targetOsVersion) ->
+		if not is_online
+			throw new Error("The device is offline: #{uuid}")
+
+		if not uuid
+			throw new Error('The uuid of the device is not available')
```
imrehg: Is this a message shown to the user? Just wondering about the wording, and in what case this would happen. Is it like "Device not found with this UUID?" Also, should it come before "not is_online", a the UUID check is a more basic test, right?

tests/integration/models/device.spec.coffee
```diff
@@ -1369,3 +1450,235 @@ describe 'Device Model', ->
...
+						os_version
+						os_variant
+					}
+
+					m.chai.expect(-> balena.models.device._checkOsUpdateTarget(
+						mockDevice
+						'2.29.2+rev1.prod'
+					)).to.throw('Invalid current balenaOS version')
+
+			it 'should throw when the current os version is a very old v1', ->
```
imrehg: Is it because it's "old", or is it because we exclude devices that are OS version < 1.8.0, as they had a file system fingerprinting bug, that would prevent proper updates? :)  (that's the reason it's not self serve for those versions)

tests/integration/models/device.spec.coffee
```diff
@@ -1369,3 +1450,235 @@ describe 'Device Model', ->
...
+						os_version
+						os_variant
+					}
+
+					m.chai.expect(-> balena.models.device._checkOsUpdateTarget(
+						mockDevice
+						'2.29.2+rev1.prod'
+					)).to.throw('Updates cannot be performed on development balenaOS variants')
+
+			it 'should throw when the device is running a pre-release version', ->
```
imrehg: I don't think it should throw an error, these versions should be updatable, IMHO.

lib/models/device.coffee
thgreasi: That's resolved @imrehg . Wrapping with `Promise.method` ensures that `getDeviceUrlsBase()` will always return a promise.

lib/models/device.coffee
thgreasi: > Is this a message shown to the user?

No. Since `_checkOsUpdateTarget()` isn't a public/documented method, that is called by `startOsUpdate()`, this check is there so that we will notice if the device object passed in is missing the `uuid`, which is required for the rest of the errors. In short, this will only show up in the tests cases, in case we do a bad refactor.

> should it come before "not is_online", a the UUID check is a more basic test, right?

Yes, that's correct :+1:

tests/integration/models/device.spec.coffee
thgreasi: `< 1.8.0` is what we are actually checking, but I though to not make it part of the rules, since this is already checked in the hup utils module. Now that I'm thinking it again, let's change it, so that we remember how/why the test case was written.

tests/integration/models/device.spec.coffee
thgreasi: The proxy & the hup utils actually prohibit it, so I'm just replicating the same check in here, to make sure whether an update changes the behavior.
If that's indeed something supported, then we should first update it in the proxy.
See: https://github.com/balena-io/resin-proxy/blob/5efa5a905866fef31086c195e9d9ec4642eb1ee1/src/common/version-utils.ts#L66-L70
See: https://github.com/balena-io-modules/balena-hup-action-utils/blob/5032cf146ae8171781305881f4beb2ef1ef4ce8c/lib/index.ts#L80-L87

thgreasi:
@imrehg I actually took some extra time to better group the hup types & also add a secondary device type to test for each hup type. I hope you like it more this way :)

PR #648, opened 2019-03-12
Add getSupervisorState device method

xginn8:
HQ: balena-io/balena#1663
Change-type: minor
Signed-off-by: Matthew McGinn <mamcgi@gmail.com>

### DIFF
--- a/lib/models/device.coffee
+++ b/lib/models/device.coffee
@@ -52,7 +52,7 @@ deviceStatus = require('resin-device-status')
 
 # The min version where /apps API endpoints are implemented is 1.8.0 but we'll
 # be accepting >= 1.8.0-alpha.0 instead. This is a workaround for a published 1.8.0-p1
-# prerelase supervisor version, which precedes 1.8.0 but comes after 1.8.0-alpha.0
+# prerelease supervisor version, which precedes 1.8.0 but comes after 1.8.0-alpha.0
 # according to semver.
 MIN_SUPERVISOR_APPS_API = '1.8.0-alpha.0'
 
@@ -1359,6 +1359,46 @@ getDeviceModel = (deps, opts) ->
 					appId: device.belongs_to__application[0].id
 					data:
 						force: Boolean(options.force)
+		.asCallback(callback)
+
+	###*
+	# @summary Get the supervisor state on a device
+	# @name getSupervisorState
+	# @public
+	# @function
+	# @memberof balena.models.device
+	#
+	# @param {String|Number} uuidOrId - device uuid (string) or id (number)
+	# @returns {Promise}
+	#
+	# @example
+	# balena.models.device.getSupervisorState('7cf02a6').then(function(state) {
+	#	 	console.log(state);
+	# });
+	#
+	# @example
+	# balena.models.device.getSupervisorState(123).then(function(state) {
+	#	 	console.log(state);
+	# });
+	#
+	# @example
+	# balena.models.device.getSupervisorState('7cf02a6', function(error, state) {
+	# 		if (error) throw error;
+	#	 	console.log(state);
+	# });
+	###
+	exports.getSupervisorState = (uuidOrId, callback) ->
+		exports.get uuidOrId,
+			$select: 'uuid'
+		.then ({ uuid }) ->
+			return request.send
+				method: 'POST'
+				url: '/supervisor/v1/device'
+				baseUrl: apiUrl
+				body:
+					uuid: uuid
+					method: 'GET'
+		.get('body')
 		.asCallback(callback)
 
 	###*
--- a/typings/balena-sdk.d.ts
+++ b/typings/balena-sdk.d.ts
@@ -452,6 +452,19 @@ declare namespace BalenaSdk {
 		note?: string;
 	}
 
+	interface SupervisorStatus {
+		api_port: string;
+		ip_address: string;
+		os_version: string;
+		supervisor_version: string;
+		update_pending: boolean;
+		update_failed: boolean;
+		update_downloaded: boolean;
+		status?: string | null;
+		commit?: string | null;
+		download_progress?: string | null;
+	}
+
 	interface ServiceInstance {
 		created_at: string;
 		id: number;
@@ -874,6 +887,9 @@ declare namespace BalenaSdk {
 					uuidOrId: string | number,
 					{ force }: { force?: boolean },
 				): Promise<void>;
+				getSupervisorState(
+					uuidOrId: string | number,
+				): Promise<SupervisorStatus>;
 				getDisplayName(deviceTypeName: string): string;
 				getDeviceSlug(deviceTypeName: string): string;
 				generateUniqueKey(): string;
--- a/DOCUMENTATION.md
+++ b/DOCUMENTATION.md
@@ -111,6 +111,7 @@ If you feel something is missing, not clear or could be improved, please don't h
             * [.shutdown(uuidOrId, [options])](#balena.models.device.shutdown) ⇒ <code>Promise</code>
             * [.purge(uuidOrId)](#balena.models.device.purge) ⇒ <code>Promise</code>
             * [.update(uuidOrId, [options])](#balena.models.device.update) ⇒ <code>Promise</code>
+            * [.getSupervisorState(uuidOrId)](#balena.models.device.getSupervisorState) ⇒ <code>Promise</code>
             * [.getDisplayName(deviceTypeSlug)](#balena.models.device.getDisplayName) ⇒ <code>Promise</code>
             * [.getDeviceSlug(deviceTypeName)](#balena.models.device.getDeviceSlug) ⇒ <code>Promise</code>
             * [.getSupportedDeviceTypes()](#balena.models.device.getSupportedDeviceTypes) ⇒ <code>Promise</code>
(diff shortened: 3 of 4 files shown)


lib/models/device.coffee
```diff
@@ -1345,6 +1345,45 @@ getDeviceModel = (deps, opts) ->
...
+	# function(error) {
+	# 	if (error) throw error;
+	# });
+	###
+	exports.getSupervisorState = (uuidOrId, callback) ->
+		callback = findCallback(arguments)
+
+		exports.get uuidOrId,
+			$select: 'uuid'
+		.then (device) ->
```
thgreasi: since the uuid is all that we need we could do
```suggestion
		.then ({ uuid }) ->
```

lib/models/device.coffee
```diff
@@ -1345,6 +1345,45 @@ getDeviceModel = (deps, opts) ->
...
+			return request.send
+				method: 'POST'
+				url: '/supervisor/v1/device'
+				baseUrl: apiUrl
+				body:
+					uuid: device.uuid
+					method: 'GET'
+		.then ( response ) ->
```
thgreasi: This could be
```suggestion
		.get('body')
```
otherwise please format as:
```suggestion
		.then (response) ->
```
But TBH I prefer the `.get()` variant.

lib/models/device.coffee
```diff
@@ -1345,6 +1345,45 @@ getDeviceModel = (deps, opts) ->
...
+	#
+	# @example
+	# balena.models.device.getSupervisorState('7cf02a6');
+	#
+	# @example
+	# balena.models.device.getSupervisorState(123);
+	#
+	# @example
+	# balena.models.device.getSupervisorState('7cf02a6',
+	# function(error) {
```
thgreasi: Let's not add a newline here and keep the oneline format that the rest methods use.

lib/models/device.coffee
```diff
@@ -1345,6 +1345,45 @@ getDeviceModel = (deps, opts) ->
...
+	# 	if (error) throw error;
+	# });
+	###
+	exports.getSupervisorState = (uuidOrId, callback) ->
+		callback = findCallback(arguments)
```
thgreasi: We don't really need `findCallback`, since if the callback is provided, it will be the second argument.
```suggestion
```
The purpose of `findCallback`, is for the case that the method accepts optional arguments, which can make callback arg position vary, like in `device.getAll()`.

lib/models/device.coffee
```diff
@@ -1345,6 +1345,45 @@ getDeviceModel = (deps, opts) ->
...
+	# @name getSupervisorState
+	# @public
+	# @function
+	# @memberof balena.models.device
+	#
+	# @param {String|Number} uuidOrId - device uuid (string) or id (number)
+	# @returns {Promise}
+	#
+	# @example
+	# balena.models.device.getSupervisorState('7cf02a6');
```
thgreasi: Please also add a `.then()` like we have in the rest of the methods.
```suggestion
	# balena.models.device.getSupervisorState('7cf02a6').then(function(state) {
	# 	console.log(state);
	# });
```

lib/models/device.coffee
```diff
@@ -1345,6 +1345,45 @@ getDeviceModel = (deps, opts) ->
...
+	#
+	# @example
+	# balena.models.device.getSupervisorState(123);
```
thgreasi: Please also add a `.then()` like we have in the rest of the methods.
```suggestion
	# balena.models.device.getSupervisorState(123).then(function(state) {
	# 	console.log(state);
	# });
```

lib/models/device.coffee
xginn8: I used the `update` method as a template, which has no `.then()`, but happy to add it if you prefer.

xginn8:
> Looks great and only requesting coding style & docs changes to keep it consistent with the rest of the code +1
> Please also add:
> 
>     * some typing for this method in `typings/balena-sdk.d.ts`
> 
>     * a test case in the `tests/integration/models/device.spec.coffee` under the `'given a single application with a single offline device'` section, so that we can document somewhat how a response from that endpoint looks like, even if in this case the request gets rejected since the device isn't actually real.

Done. I also decided to treat the inaccessible device as a `BalenaDeviceNotFound` to be consistent with other sdk methods.

lib/models/device.coffee
xginn8: Addressed in latest push

lib/models/device.coffee
xginn8: Addressed in latest push

lib/models/device.coffee
```diff
@@ -1345,6 +1345,41 @@ getDeviceModel = (deps, opts) ->
...
+	# @returns {Promise}
+	#
+	# @example
+	# balena.models.device.getSupervisorState('7cf02a6').then(function(state) { console.log(state); });;
+	#
+	# @example
+	# balena.models.device.getSupervisorState(123).then(function(state) { console.log(state); });;
+	#
+	# @example
+	# balena.models.device.getSupervisorState('7cf02a6', function(error) { if (error) throw error; });
```
thgreasi: I'm sorry if I was confusing about the newline part. I only meant it for the `, function`.
```suggestion
	# balena.models.device.getSupervisorState('7cf02a6', function(error, state) {
		if (error) throw error;
		console.log(state);
	});
```

lib/models/device.coffee
```diff
@@ -1345,6 +1345,41 @@ getDeviceModel = (deps, opts) ->
...
+	# @name getSupervisorState
+	# @public
+	# @function
+	# @memberof balena.models.device
+	#
+	# @param {String|Number} uuidOrId - device uuid (string) or id (number)
+	# @returns {Promise}
+	#
+	# @example
+	# balena.models.device.getSupervisorState('7cf02a6').then(function(state) { console.log(state); });;
```
thgreasi:
```suggestion
	# balena.models.device.getSupervisorState('7cf02a6').then(function(state) {
		console.log(state);
	});
```
also watch the double `;;`

lib/models/device.coffee
```diff
@@ -1345,6 +1345,41 @@ getDeviceModel = (deps, opts) ->
...
+	#
+	# @example
+	# balena.models.device.getSupervisorState(123).then(function(state) { console.log(state); });;
```
thgreasi:
```suggestion
	# balena.models.device.getSupervisorState(123).then(function(state) {
		console.log(state);
	});
```
also watch the double `;;`

lib/models/device.coffee
```diff
@@ -1345,6 +1345,41 @@ getDeviceModel = (deps, opts) ->
...
+		.then ({ uuid }) ->
+			return request.send
+				method: 'POST'
+				url: '/supervisor/v1/device'
+				baseUrl: apiUrl
+				body:
+					uuid: uuid
+					method: 'GET'
+		.get('body')
+		.catch(notFoundResponse, treatAsMissingDevice(uuidOrId))
```
thgreasi: Just to be clear, `exports.get()` will actually reject with a BalenaDeviceNotFound. The rest of the methods that use `treatAsMissingDevice` are using the plain `getId` which isn't checking the API if the parameter provided is a number, and that's why they definitely need the `treatAsMissingDevice`.
This shouldn't be necessary, but I'm fine with it, as far as the `'/supervisor/v1/device'` doesn't return a 404 in any other case than the device not being in the DB.

typings/balena-sdk.d.ts
```diff
@@ -857,6 +863,7 @@ declare namespace BalenaSdk {
 					uuidOrId: string | number,
 					{ force }: { force?: boolean },
 				): Promise<void>;
+				getSupervisorState(uuidOrId: string | number): Promise<void>;
```
thgreasi: I don't think that `void `describes the result of this method.

lib/models/device.coffee
xginn8: Updated, thank you for clarifying!

lib/models/device.coffee
xginn8: Updated, thank you for clarifying!

typings/balena-sdk.d.ts
xginn8: Updated, good catch!

typings/balena-sdk.d.ts
```diff
@@ -874,6 +880,7 @@ declare namespace BalenaSdk {
 					uuidOrId: string | number,
 					{ force }: { force?: boolean },
 				): Promise<void>;
+				getSupervisorState(uuidOrId: string | number): Promise<string>;
```
thgreasi: In my tests this returned
```
{
  "api_port": "48484",
  "ip_address": "192.168.1.103",
  "os_version": "Resin OS 2.12.3+rev1",
  "supervisor_version": "7.1.14",
  "update_pending": false,
  "update_failed": false,
  "update_downloaded": false,
  "commit": "861833c22f53251956ec25f7162d9882d3848ef2",
  "status": "Idle",
  "download_progress": null
}
```

lib/models/device.coffee
```diff
@@ -1361,6 +1361,48 @@ getDeviceModel = (deps, opts) ->
...
+	# });
+	#
+	# @example
+	# balena.models.device.getSupervisorState('7cf02a6', function(error, state) {
+	# 		if (error) throw error;
+	#	 	console.log(state);
+	# });
+	###
+	exports.getSupervisorState = (uuidOrId, callback) ->
+
```
thgreasi: NIT: Let's also remove this empty line.

lib/models/device.coffee
thgreasi: You will have to remove this line, since it is causing the `it 'should be rejected if the device exists but is inaccessible'` test case to fail.

xginn8:
@resin-ci retest

thgreasi:
@xginn8 instead of `retest` isn't going to help here.
You need to open the appveyor check and click restart.
image
image
Let me try this once.

thgreasi:
The requires tests have passed @xginn8 . Feel free to merge this :+1:

PR #653, opened 2019-03-26
diagnostics: add getSupervisorTargetState support

xginn8:
HQ: balena-io/balena#1663
Change-type: minor
Signed-off-by: Matthew McGinn <mamcgi@gmail.com>

##### Contributor checklist
- [x] Includes tests
- [ ] Includes typings

### DIFF
--- a/lib/models/device.coffee
+++ b/lib/models/device.coffee
@@ -1362,6 +1362,42 @@ getDeviceModel = (deps, opts) ->
 		.asCallback(callback)
 
 	###*
+	# @summary Get the target supervisor state on a device
+	# @name getSupervisorTargetState
+	# @public
+	# @function
+	# @memberof balena.models.device
+	#
+	# @param {String|Number} uuidOrId - device uuid (string) or id (number)
+	# @returns {Promise}
+	#
+	# @example
+	# balena.models.device.getSupervisorTargetState('7cf02a6').then(function(state) {
+	# 	console.log(state);
+	# });
+	#
+	# @example
+	# balena.models.device.getSupervisorTargetState(123).then(function(state) {
+	# 	console.log(state);
+	# });
+	#
+	# @example
+	# balena.models.device.getSupervisorTargetState('7cf02a6', function(error, state) {
+	# 	if (error) throw error;
+	# 	console.log(state);
+	# });
+	###
+	exports.getSupervisorTargetState = (uuidOrId, callback) ->
+		exports.get uuidOrId,
+			$select: 'uuid'
+		.then ({ uuid }) ->
+			return request.send
+				url: "/device/v2/#{uuid}/state"
+				baseUrl: apiUrl
+		.get('body')
+		.asCallback(callback)
+
+	###*
 	# @summary Get the supervisor state on a device
 	# @name getSupervisorState
 	# @public
@@ -1373,18 +1409,18 @@ getDeviceModel = (deps, opts) ->
 	#
 	# @example
 	# balena.models.device.getSupervisorState('7cf02a6').then(function(state) {
-	#	 	console.log(state);
+	#	console.log(state);
 	# });
 	#
 	# @example
 	# balena.models.device.getSupervisorState(123).then(function(state) {
-	#	 	console.log(state);
+	# 	console.log(state);
 	# });
 	#
 	# @example
 	# balena.models.device.getSupervisorState('7cf02a6', function(error, state) {
-	# 		if (error) throw error;
-	#	 	console.log(state);
+	# 	if (error) throw error;
+	# 	console.log(state);
 	# });
 	###
 	exports.getSupervisorState = (uuidOrId, callback) ->
--- a/tests/integration/models/device.spec.coffee
+++ b/tests/integration/models/device.spec.coffee
@@ -1028,6 +1028,31 @@ describe 'Device Model', ->
 					m.chai.expect(_.find(result, { name: 'A' }).value).equal('a')
 					m.chai.expect(_.find(result, { name: 'B' }).value).equal('b')
 
+		describe 'balena.models.device.getSupervisorTargetState()', ->
+
+			it 'should be rejected if the device does not exist', ->
+				promise = balena.models.device.getSupervisorTargetState('asdfghjkl')
+				m.chai.expect(promise).to.be.rejectedWith('Device not found: asdfghjkl')
+
+			it 'should reflect the device\'s target state', ->
+				balena.models.device.getSupervisorTargetState(@device.id).then (state) =>
+					# first, check the name
+					m.chai.expect(state.local.name).to.be.a('string')
+					m.chai.expect(state.local.name).to.equal(@device.device_name)
+
+					# next, check application types and some values
+					m.chai.expect(state.local.apps).to.be.an('object')
+					m.chai.expect(state.local.apps[@application.id].name).to.equal(@application.app_name)
+					m.chai.expect(state.local.apps[@application.id].name).to.be.a('string')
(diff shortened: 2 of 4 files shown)


lib/models/device.coffee
```diff
@@ -1361,6 +1361,42 @@ getDeviceModel = (deps, opts) ->
...
+	#
+	# @example
+	# balena.models.device.getSupervisorTargetState(123).then(function(state) {
+	#	 	console.log(state);
+	# });
+	#
+	# @example
+	# balena.models.device.getSupervisorTargetState('7cf02a6', function(error, state) {
+	# 		if (error) throw error;
+	#	 	console.log(state);
```
thgreasi: Let's change the examples to not use 2 tabs for their indentation.

tests/integration/models/device.spec.coffee
```diff
@@ -1028,6 +1028,18 @@ describe 'Device Model', ->
...
+		describe.only 'balena.models.device.getSupervisorTargetState()', ->
+
+			it 'should be rejected if the device does not exist', ->
+				promise = balena.models.device.getSupervisorTargetState('asdfghjkl')
+				m.chai.expect(promise).to.be.rejectedWith('Device not found: asdfghjkl')
+
+			it 'should reflect the device\'s state', ->
+				balena.models.device.getSupervisorTargetState(@device.id).then (state) =>
+					m.chai.expect(state.name).to.equal(@device.name)
+					m.chai.expect(state.apps[@application.id].name).to.equal(@application.name)
```
thgreasi: It would be nice to also include some extra `.to.be.a('number')` or `.to.be.a('string')` checks, that will more or less check what the typings describe about this object.
This way we will notice if suddenly the state endpoint response format changes.

lib/models/device.coffee
xginn8: Done! That was consistent with the `getSupervisorState` method, but I fixed that too.

tests/integration/models/device.spec.coffee
xginn8: @thgreasi i cannot figure out why `@application` isn't defined in https://github.com/balena-io/balena-sdk/pull/653/files#diff-499041c13cee047a446c4f6ee3a36c14R1045, yet seems to be defined in the above definition (https://github.com/balena-io/balena-sdk/pull/653/files#diff-499041c13cee047a446c4f6ee3a36c14R1027)

tests/integration/models/device.spec.coffee
```diff
@@ -1028,6 +1029,28 @@ describe 'Device Model', ->
...
+			it 'should be rejected if the device does not exist', ->
+				promise = balena.models.device.getSupervisorTargetState('asdfghjkl')
+				m.chai.expect(promise).to.be.rejectedWith('Device not found: asdfghjkl')
+
+			it 'should reflect the device\'s name', ->
+				balena.models.device.getSupervisorTargetState(@device.id).then (state) ->
+					m.chai.expect(state.local.name).to.be.a('string')
+
+			it 'should reflect the device\'s apps', ->
+				balena.models.device.getSupervisorTargetState(@device.id).then (state) ->
```
thgreasi: I think that the issue is that you need `=>` instead of `->` so that the `this` is still the same inside the promise handler:
```suggestion
				balena.models.device.getSupervisorTargetState(@device.id).then (state) =>
```

tests/integration/models/device.spec.coffee
```diff
@@ -1028,6 +1028,25 @@ describe 'Device Model', ->
...
+				m.chai.expect(promise).to.be.rejectedWith('Device not found: asdfghjkl')
+
+			it 'should reflect the device\'s name', ->
+				balena.models.device.getSupervisorTargetState(@device.id).then (state) ->
+					m.chai.expect(state.local.name).to.be.a('string')
+
+			it 'should reflect the device\'s apps', ->
+				balena.models.device.getSupervisorTargetState(@device.id).then (state) =>
+					m.chai.expect(state.local.apps).to.be.an('object')
+					m.chai.expect(state.local.apps[@application.id.toString()].name).to.equal(@application.app_name)
```
thgreasi: You probably can drop the `.toString()`s in here to make it a bit shorter, since JS will do that for you.

tests/integration/models/device.spec.coffee
```diff
@@ -1028,6 +1028,25 @@ describe 'Device Model', ->
...
 
+		describe 'balena.models.device.getSupervisorTargetState()', ->
+
+			it 'should be rejected if the device does not exist', ->
+				promise = balena.models.device.getSupervisorTargetState('asdfghjkl')
+				m.chai.expect(promise).to.be.rejectedWith('Device not found: asdfghjkl')
+
+			it 'should reflect the device\'s name', ->
+				balena.models.device.getSupervisorTargetState(@device.id).then (state) ->
+					m.chai.expect(state.local.name).to.be.a('string')
```
sradevski: In general for this and the other test case, it would be great if we can test for something more concrete than the type of a field. Since we know the name of the device, we should check if the name matches, not just that it is a string. In the other test cases as well, if some of the values are predictable, it is better to test against concrete values to avoid false positives.

xginn8: @sradevski here's the full response from the API in this case:

```
{"local":{"name":"icy-moon","config":{"RESIN_SUPERVISOR_NATIVE_LOGGER":"true","RESIN_SUPERVISOR_POLL_INTERVAL":"900000"},"apps":{"1424972":{"name":"FooBar","services":{},"volumes":{},"networks":{}}}},"dependent":{"apps":{},"devices":{}}}
```

I have added more exact checks (device name, app name, config options) however we will need to do a fair bit more leg work to set up volumes or networks and confirm that they have been returned properly.

sradevski: If it requires too much work, it might not be necessary. It was just my initial impression, but not something that is required.

xginn8:
> Just drop the `.toString()`s in the test cases and this is +1
> I expected that you would still keep the typings as a separate commit (b/c of the different author) but I don't have strong feeling for this.

Sorry about the misunderstanding @thgreasi! Noted for next time, I definitely did not mean to step on your toes with that commit.

tests/integration/models/device.spec.coffee
```diff
@@ -1028,6 +1028,32 @@ describe 'Device Model', ->
...
+				balena.models.device.getSupervisorTargetState(@device.id).then (state) =>
+					m.chai.expect(state.local.apps).to.be.an('object')
+					m.chai.expect(state.local.apps[@application.id].name).to.equal(@application.app_name)
+					m.chai.expect(state.local.apps[@application.id].name).to.be.a('string')
+					m.chai.expect(state.local.apps[@application.id].services).to.be.an('object')
+					m.chai.expect(state.local.apps[@application.id].volumes).to.be.an('object')
+					m.chai.expect(state.local.apps[@application.id].networks).to.be.an('object')
+
+			it 'should reflect the device\'s config', ->
+				balena.models.device.getSupervisorTargetState(@device.id).then (state) ->
```
thgreasi: Since this and the above two tests are doing the same call & the state endpoint is known to be slow, let's combine them to same some seconds in the test suit execution time & be more kind with the API.
```
        ✓ should reflect the device's name (1161ms)
        ✓ should reflect the device's apps (1125ms)
        ✓ should reflect the device's config (1138ms)
```

xginn8: great catch, i've combined these tests and commented the different sections accordingly.

thgreasi:
LGTM, feel free to merge once the tests pass :+1:

Issue #667, opened 2019-04-17
Should refactor getSupportedVersions test to be more helpful when failing


thgreasi:
Placeholder issue.

PR #668, opened 2019-04-17
Improve the verbosity & stability of os.getSupportedVersions tests

thgreasi:
Without the `expectSorted` change the tests might fail on staging when the `versions` array contains items of the same normalized semver value, depending on the length of the array.
Found this edge case when sorting the versions from
https://api.balena-staging.com/device-types/v1/raspberry-pi/images/
in which case `"2.0.0+rev1"` & `"2.0.0.rev1"` (which have the same normalized semver value) swapped positions after resorting.
Demo: https://runkit.com/embed/690i3z8wz9q2

Resolves: #667
Change-type: patch

---
##### Contributor checklist

- [x] Includes tests
- [ ] Includes typings

### DIFF
--- a/tests/integration/models/os.spec.coffee
+++ b/tests/integration/models/os.spec.coffee
@@ -91,13 +91,20 @@ describe 'OS model', ->
 		describe 'given a valid device slug', ->
 
 			areValidVersions = (osVersions) ->
-				sortedVersions = _.clone(osVersions.versions)
-				sortedVersions.sort(osVersionRCompare)
-				return osVersions and
-					osVersions.versions and osVersions.versions.length and
-					_.isEqual(osVersions.versions, sortedVersions)
-					osVersions.latest and osVersions.recommended and osVersions.default and
-					osVersions.default is osVersions.recommended
+				m.chai.expect(osVersions).to.be.an('object')
+				m.chai.expect(osVersions).to.have.property('versions').that.is.an('array')
+				m.chai.expect(osVersions.versions).to.not.have.lengthOf(0)
+
+				sortedVersions = osVersions.versions.slice().sort(osVersionRCompare)
+				m.chai.expect(osVersions.versions).to.deep.equal(sortedVersions)
+
+				m.chai.expect(osVersions).to.have.property('latest').that.is.a('string')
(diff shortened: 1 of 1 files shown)


tests/integration/models/os.spec.coffee
```diff
@@ -90,14 +90,31 @@ describe 'OS model', ->
...
-				sortedVersions = _.clone(osVersions.versions)
-				sortedVersions.sort(osVersionRCompare)
-				return osVersions and
-					osVersions.versions and osVersions.versions.length and
-					_.isEqual(osVersions.versions, sortedVersions)
-					osVersions.latest and osVersions.recommended and osVersions.default and
-					osVersions.default is osVersions.recommended
+				m.chai.expect(osVersions).to.be.an('object')
+				m.chai.expect(osVersions).to.have.property('versions').that.is.an('array')
+				m.chai.expect(osVersions.versions).to.be.an('array')
```
karaxuna: Isn't this covered on previous line? `that.is.an('array')`

tests/integration/models/os.spec.coffee
```diff
@@ -90,14 +90,31 @@ describe 'OS model', ->
...
+				m.chai.expect(osVersions.versions).to.not.have.lengthOf(0)
+
+				expectSorted(osVersions.versions, osVersionRCompare)
+
+				m.chai.expect(osVersions).to.have.property('latest').that.is.a('string')
+				m.chai.expect(osVersions).to.have.property('recommended').that.is.a('string')
+				m.chai.expect(osVersions).to.have.property('default').that.is.a('string')
+				m.chai.expect(osVersions.default).to.equal(osVersions.recommended)
```
karaxuna: Indentation issue?

tests/integration/models/os.spec.coffee
thgreasi: @karaxuna yea that's true :+1: I will push an update asap

tests/integration/models/os.spec.coffee
thgreasi: @karaxuna I don't see one. Can you be a bit more specific?

karaxuna: Oh sorry, screen size was the problem :)

Issue #664, opened 2019-04-17
Should use the open-balena-api endpoints for device type & version info


thgreasi:
We currently use image maker endpoints for these.

Depends-on: https://github.com/balena-io/resin-api/issues/1829
HQ: https://github.com/balena-io/balena/issues/1744

PR #669, opened 2019-04-17
Use the open-balena-api endpoints for device type & version info

thgreasi:
That's on top of https://github.com/balena-io/balena-sdk/pull/668, so the final PR will not require any changes in the tests.

Resolves: #664
Depends-on: balena-io/resin-api#1829
HQ: balena-io/balena#1744
Change-type: minor

---
##### Contributor checklist

- [x] Includes tests
- [ ] Includes typings

### DIFF
--- a/lib/models/os.coffee
+++ b/lib/models/os.coffee
@@ -63,13 +63,23 @@ getOsModel = (deps, opts) ->
 			if not isValid
 				throw new errors.BalenaInvalidDeviceType('No such device type')
 
-	getDownloadSize = imgMakerHelper.buildApiRequester
-		buildUrl: ({ deviceType, version }) -> "/size_estimate?deviceType=#{deviceType}&version=#{version}"
-		postProcess: ({ body }) -> body.size
+	getDownloadSize = (deviceType, version) ->
+		request.send
+			method: 'GET'
+			url: "/device-types/v1/#{deviceType}/images/#{version}/download-size"
+			baseUrl: apiUrl
+			sendToken: false
+		.get('body')
+		.get('size')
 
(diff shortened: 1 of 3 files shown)


lib/models/os.coffee
```diff
@@ -40,6 +41,11 @@ semver = require('semver')
 
 BALENAOS_VERSION_REGEX = /v?\d+\.\d+\.\d+(\.rev\d+)?((\-|\+).+)?/
 
+DEVICE_TYPES_ENDPOINT_CACHING_INTERVAL = 10 * 60 * 1000 # 10 minutes
```
thgreasi: Similarly to how the image maker helper does its own caching.

thgreasi:
@balena-ci retest

tests/integration/models/os.spec.coffee
```diff
@@ -121,12 +121,37 @@ describe 'OS model', ->
 				promise = balena.models.os.getSupportedVersions('raspberrypi')
 				m.chai.expect(promise).to.eventually.satisfy(areValidVersions)
 
+			it 'should cache the results', ->
+				balena.models.os.getSupportedVersions('raspberry-pi')
+				.then (result1) ->
```
sradevski: `result1` is also coming from cache, isn't it? There are calls for `raspberrypi` without clearing the cache before this test, which would set the cache. This still tests what is supposed to test, and I don't think it can result in flaky tests, just thought I should mention it in case you didn't take it into consideration.

thgreasi: Yes that's true @sradevski ,
Maybe we can use a different device type, or call the `_clear...()` helper before.
Just give me a note and I can change it if you think we should do so.

Issue #670, opened 2019-04-24
tests: Should start batching test cases to run on the same state to improve their speed


thgreasi:
Atm we delete all applications on a beforeEach.
We should make the tests work on shared state to improve their speed.
See: https://app.frontapp.com/open/cnv_13stakz

PR #671, opened 2019-04-24
tests: Batch test case states, to make the tests faster

thgreasi:
Resolves: #670
See: https://app.frontapp.com/open/cnv_13stakz
Change-type: patch

---
##### Contributor checklist

- [x] Includes tests
- [x] Includes typings

### DIFF
--- a/tests/integration/models/device.spec.coffee
+++ b/tests/integration/models/device.spec.coffee
@@ -1020,8 +1020,10 @@ describe 'Device Model', ->
 					.then =>
 						configVarModel.getAllByDevice(@device[deviceParam])
 					.then (result) ->
-						m.chai.expect(_.find(result, { name: "BALENA_A_#{deviceParamUpper}" }).value).equal('a')
-						m.chai.expect(_.find(result, { name: "BALENA_B_#{deviceParamUpper}" }).value).equal('b')
+						m.chai.expect(_.find(result, { name: "BALENA_A_#{deviceParamUpper}" })).to.be.an('object')
+							.that.has.property('value', 'a')
+						m.chai.expect(_.find(result, { name: "BALENA_B_#{deviceParamUpper}" })).to.be.an('object')
+							.that.has.property('value', 'b')
 					.then =>
 						Promise.all [
 							configVarModel.remove(@device[deviceParam], "BALENA_A_#{deviceParamUpper}")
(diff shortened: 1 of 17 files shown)


tests/integration/setup.coffee
```diff
@@ -84,8 +84,8 @@ exports.resetUser = ->
 
 exports.credentials = buildCredentials()
 
-exports.givenLoggedInUserWithApiKey = ->
-	beforeEach ->
+exports.givenLoggedInUserWithApiKey = (beforeFn = beforeEach) ->
```
sradevski: This might cause some confusion down the road, so maybe it is better to be explicit about the function we are passing (instead of a default value). I know it is a bit more verbose, but it is immediately obvious from the tests whether it runs on before or beforeEach. I don't have a strong opinion on it, so I will leave it up to you to decide.

tests/integration/models/device.spec.coffee
```diff
@@ -500,28 +510,30 @@ describe 'Device Model', ->
 
 			it 'should be able to set the location of a device by id', ->
 				balena.models.device.setCustomLocation @device.id,
-					latitude: 41.383333
-					longitude: 2.183333
+					latitude: 42.383333
```
sradevski: I am curious why you changes the long/lat values :)

xginn8: Setcases >> Barcelona!

thgreasi: Previously, all method variants were setting the same value and as a result we had to re-create the device, in order to test that the value was properly set by all test cases.
Now, by setting different values by each method variant, we can test that each call was successful w/o the need of re-creating the device.

tests/integration/setup.coffee
thgreasi: Totally agree on that. I initially used a default value, since I was originally thinking about just changing the device.spec, which was only of the larger parts of our test suit.
Since this escalated and touched most of the test cases, it totally makes sense to make this mandatory and throw if missing. Will do :+1:

thgreasi:
I should have warned to review each commit on each own & that some test cases were moved a bit, so that we could use less `before` statements and keep more tests under it.

tests/integration/models/device.spec.coffee
thgreasi: Also, I didn't check the original coordinates, but Barcelona makes sense, since Tim was there :)

Issue #681, opened 2019-05-07
Should use balena-semver insteat of the custom osVersionRCompare


thgreasi:
Placeholder issue

PR #682, opened 2019-05-07
Upgrade resin-semver to balena-semver

thgreasi:
Resolves: #680
Change-type: minor
Signed-off-by: Thodoris Greasidis <thodoris@balena.io>

---
##### Contributor checklist

- [ ] Includes tests
- [ ] Includes typings
- [ ] Includes updated documentation
- [ ] Includes updated build output

PR #683, opened 2019-05-07
Use balena-semver insteat of the custom osVersionRCompare

thgreasi:
Resolves: #681
Change-type: minor

---
##### Contributor checklist

- [ ] Includes tests
- [ ] Includes typings

### DIFF
--- a/tests/integration/models/os.spec.coffee
+++ b/tests/integration/models/os.spec.coffee
@@ -5,7 +5,7 @@ bSemver = require('balena-semver')
 
 { balena, givenLoggedInUser, IS_BROWSER } = require('../setup')
 
-{ osVersionRCompare } = require('../../../build/util')
+{ osVersionRCompare } = require('../../osVersionRCompare')
 
 eventuallyExpectProperty = (promise, prop) ->
 	m.chai.expect(promise).to.eventually.have.property(prop)
--- a/tests/integration/models/os.spec.coffee
+++ b/tests/integration/models/os.spec.coffee
@@ -106,6 +106,9 @@ describe 'OS model', ->
 				m.chai.expect(osVersions).to.have.property('versions').that.is.an('array')
 				m.chai.expect(osVersions.versions).to.not.have.lengthOf(0)
 
+				expectSorted(osVersions.versions, bSemver.rcompare)
+				# we keep this to ensure balena-semver has the exact same behavior
+				# and will drop it in a later PR
 				expectSorted(osVersions.versions, osVersionRCompare)
 
 				m.chai.expect(osVersions).to.have.property('latest').that.is.a('string')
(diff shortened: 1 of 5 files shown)


thgreasi:
That's on top of https://github.com/balena-io/balena-sdk/pull/682 .
I will rebase and mark as ready for review once the other PR is merged.

tests/integration/models/os.spec.coffee
```diff
@@ -5,7 +5,7 @@ bSemver = require('balena-semver')
 
 { balena, givenLoggedInUser, IS_BROWSER } = require('../setup')
 
-{ osVersionRCompare } = require('../../../build/util')
+{ osVersionRCompare } = require('../../osVersionRCompare')
```
sradevski: It seems this is not being used anywhere.

tests/osVersionRCompare.coffee
```diff
@@ -0,0 +1,40 @@
...
+	if rev?
+		parseInt(rev, 10)
+	else
+		0
+
+isDevelopmentVersion = (version) ->
+	/(\.|\+|-)dev/.test(version)
+
+# TODO: Drop this once we make sure balena-semver is doign the exact same thing
+exports.osVersionRCompare = (versionA, versionB) ->
```
sradevski: I assume if all the tests pass using `balena-semver`, we can go ahead and remove this now, as it doesn't serve any further purpose. This only needs to be tested once (which is already done by this PR), so I guess that is what you meant to drop it once it does the same thing.

tests/integration/models/os.spec.coffee
thgreasi: But it is, on line 112
See: https://github.com/balena-io/balena-sdk/pull/683/files/c9a5b16f16e939fda96dd599dc308d3190bae8ab#diff-4d287dbe36ded3f6fd75fa6cc8ff9b36R112

tests/osVersionRCompare.coffee
thgreasi: I thought that I should have a merged PR as a proof first :)
But yea I can add an extra commit for that and rely have the appveyor build as a prood if you prefer it this way :)
Just give a note!

tests/integration/models/os.spec.coffee
sradevski: Ok, then my searching sucks, I think I searched for `osVersionCompare`.

tests/osVersionRCompare.coffee
sradevski: Not really important, anyway is fine, whatever it is easier for you. Maybe having a follow-up PR is cleaner and provides a bit more searchability (doubt it will matter, but still)

thgreasi:
See: https://github.com/balena-io/balena-sdk/issues/686

Issue #689, opened 2019-05-21
application.getAll() has a now unnecessary getUserId() call


thgreasi:
Placeholder issue.

PR #691, opened 2019-05-21
application.getAll(): Drop unnecessary getUserId() call

thgreasi:
Resolves: #689
Change-type: patch

---
##### Contributor checklist

- [ ] Includes tests
- [ ] Includes typings


thgreasi:
@sradevski that's what I also expected but the coffeelint as we currently have it thought otherwise...
I will add to my todo list to transition to resin-lint and eventually to TS :)

Issue #699, opened 2019-06-01
Should fix device.getStatus return type typings


thgreasi:
Placeholder issue

PR #700, opened 2019-06-01
Fix device.getStatus() return type typings

thgreasi:
See: https://github.com/balena-io/balena-sdk/blob/cac50add85367b853ec759242d887eef2e746d39/lib/models/device.coffee#L2006-L2009
Resolves: #699
Change-type: patch

Resolves: # 
HQ: <url> 
See: <url> 
Depends-on: <url> 

---
##### Contributor checklist

- [ ] Includes tests
- [ ] Includes typings


thgreasi:
The only CI tests failing are related to user registration rejection where the rejection is `Request error: Too many requests in this time frame.` instead of the expected one. Merging since that's expected after the ratelimiting changes in the API.

Issue #698, opened 2019-06-01
Should expose a method to combine the os version with the os variant


thgreasi:
Placeholder issue

PR #701, opened 2019-06-01
Add device.getOsVersion() that returns the complete os release version

thgreasi:
Resolves: #698
Change-type: minor

Resolves: # 
HQ: <url> 
See: <url> 
Depends-on: <url> 

---
##### Contributor checklist

- [ ] Includes tests
- [ ] Includes typings

### DIFF
--- a/lib/models/device.coffee
+++ b/lib/models/device.coffee
@@ -2100,6 +2100,26 @@ getDeviceModel = (deps, opts) ->
 		return timeSince(lce)
 
 	###*
+	# @summary Get the OS version (version number and variant combined) running on a device
+	# @name getOsVersion
+	# @public
+	# @function
+	# @memberof balena.models.device
+	#
+	# @param {Object} device - A device object
+	# @returns {String}
+	#
+	# @example
+	# balena.models.device.get('7cf02a6').then(function(device) {
+	# 	console.log(device.os_version); // => 'balenaOS 2.26.0+rev1'
+	# 	console.log(device.os_variant); // => 'prod'
+	# 	balena.models.device.getOsVersion(device); // => '2.26.0+rev1.prod'
+	# })
+	###
+	exports.getOsVersion = (device) ->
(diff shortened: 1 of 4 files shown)


lib/models/device.coffee
```diff
@@ -2099,6 +2099,26 @@ getDeviceModel = (deps, opts) ->
...
+	# @returns {String}
+	#
+	# @example
+	# balena.models.device.get('7cf02a6').then(function(device) {
+	# 	console.log(device.os_version); // => 'balenaOS 2.26.0+rev1'
+	# 	console.log(device.os_variant); // => 'prod'
+	# 	balena.models.device.getOsVersion(device); // => '2.26.0+rev1.prod'
+	# })
+	###
+	exports.getOsVersion = (device) ->
```
thgreasi: I couldn't really decide a name for this, so I'm looking forward for suggestions as well.
The name should make clear that this is the complete os release version number but also not confuse the user about the `balenaOS` prefix.
The other name that I had in mind was `getOsRelease`, trying to describe that this semver is from the complete OS release version, but I couldn't convince my self to use it.

sradevski: I think `getOsRelease` might be even more confusing (is that a release ID, a version?). I think `getOsVersion` is fine if we want to keep it short, maybe the variant is the only part that would be unexpected as part of the return value, but it is documented that it is included so it shouldn't be an issue. We can potentially add a second `options` param to specify what to include/exclude if necessary in the future.

lib/models/device.coffee
```diff
@@ -2099,6 +2099,26 @@ getDeviceModel = (deps, opts) ->
 
 		return timeSince(lce)
 
+	###*
+	# @summary Get the composite version number of the OS running on a device
```
sradevski: Maybe we need to be more concrete of what "composite" means?

Something like `Get the OS version number and variant running on a device` (please redefine it to whatever you think is clearer).

Issue #693, opened 2019-05-29
device.getWithServiceDetails: should not use the unnecessary $tolower filter


thgreasi:
```
getWithDeviceServiceDetails: 1312.061279296875ms
w/o $tolower: 1255.397216796875ms
```
See: https://www.flowdock.com/app/rulemotion/resin-frontend/threads/achNnDm_EVLZQyaleMTdqxgPq1G

PR #702, opened 2019-06-03
getCurrentServiceDetailsPineOptions: Drop $tolower to make it faster

thgreasi:
After checking the API code, it seems that it should be just lower case `'deleted'`.
It seems that's only the JS & Python SDK tests that have even created records with `'Deteled'` status.

Resolves: #693
Change-type: pacth

---
##### Contributor checklist

- [ ] Includes tests
- [ ] Includes typings

### DIFF
--- a/lib/util/index.coffee
+++ b/lib/util/index.coffee
@@ -200,11 +200,7 @@ exports.getCurrentServiceDetailsPineOptions = ->
 			]
 			$filter:
 				# We filter out deleted images entirely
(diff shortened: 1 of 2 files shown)


thgreasi:
The only CI tests failing are related to user registration rejection where the rejection is `Request error: Too many requests in this time frame.` instead of the expected one. Merging since that's expected after the ratelimiting changes in the API.

PR #705, opened 2019-06-03
Temporarily skip the auth.register() tests

thgreasi:
This temporarily disables the user registration tests,
given the current rate limiting and the upcoming
changes, so that we can have properly passing SDK
PRs until we find a way to implement the tests.

Also opened an SDK issue so that we don't forget to undo this and continue with the proper implementation.

See: https://github.com/balena-io/balena-sdk/issues/704
Change-type: patch

Resolves: # 
HQ: <url> 
See: <url> 
Depends-on: <url> 

---
##### Contributor checklist

- [ ] Includes tests
- [ ] Includes typings

### DIFF
--- a/tests/integration/auth.spec.coffee
+++ b/tests/integration/auth.spec.coffee
@@ -83,7 +83,7 @@ describe 'SDK authentication', ->
 				m.chai.expect(promise).to.be.rejected
 					.and.eventually.have.property('code', 'BalenaNotLoggedIn')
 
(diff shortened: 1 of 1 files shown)


sradevski:
Just make sure to keep a note to reverse the change when everything is done.

thgreasi:
@sradevski already have in issue for this :)
See: https://github.com/balena-io/balena-sdk/issues/704

PR #706, opened 2019-06-07
Fix downstream typing errors due to `Dictionary` helper

thgreasi:
Also add a test for the typings, using the minimum TS version that we support, so that this doesn't happen again. Moreover this protects us from the case that we start using newer TS version features and break the compatibility with the min supported TS version of each major SDK release.

Change-type: patch

---
##### Contributor checklist

- [ ] Includes tests
- [ ] Includes typings

### DIFF
--- a/package.json
+++ b/package.json
@@ -29,9 +29,10 @@
   "scripts": {
     "clean": "rimraf build",
     "test": "npm run build && npm run test:all",
-    "test:all": "npm run test:node && npm run test:browser",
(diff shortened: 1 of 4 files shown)


thgreasi:
AppVeyor errored with
> 
Error: Invalid tag name "$npm_package_config_ts_min_version": Tags may not have any characters that encodeURIComponent encodes.

So I guess I can't use a package.json config value for the TS version number...

PR #719, opened 2019-06-21
util/device-os-version: Make compatible w/ newer typings

thgreasi:
TypeScript will emit the same code, but imo this was
was a nice way to avoid the Array<string | number>
cast (that's what the new typings say for the
`prerelease` prop).

Change-type: patch
See: https://www.typescriptlang.org/play/#src=%5B...build%2C%20...versionInfo.prerelease%5D

Resolves: # 
HQ: <url> 
See: <url> 
Depends-on: <url> 

---
##### Contributor checklist

- [ ] Includes tests
- [ ] Includes typings


thgreasi:
@balena-ci retest

PR #720, opened 2019-06-21
docs: fix up broken self-referential links

xginn8:
Change-type: patch


xginn8:
@thgreasi can you merge this for me as admin? it seems builds generally are broken and i want to avoid repeatedly slamming the API with tests if possible

Issue #464, opened 2018-01-22
Users should be able to use a commit hash instead of a build ID


thgreasi:
[`models.release.get()`](https://github.com/resin-io/resin-sdk/blob/master/DOCUMENTATION.md#resin.models.build.get) currently supports only a numeric param.

We should also implement this for `getWithImageDetails`.

It could be handy to be able to use commit hashes instead.

HQ: https://github.com/resin-io/hq/issues/1166

Note that sometimes commit hashes can be ambiguous (there can only be one _successful_ release per hash, but there may also be failed releases). Debatable how we should handle that: could return a list, or we could always filter to only include the successful release (if available) or the latest release (if not).

thgreasi:
See: https://www.flowdock.com/app/rulemotion/resin-projects/threads/PdDHAT92vyFQcyNtYCH3w9i-TJU

PR #725, opened 2019-07-05
Support retrieving  releases & managing release tags by commit

thgreasi:
Resolves: #464 
Change-type: minor
HQ: https://github.com/resin-io/hq/issues/1166

---
##### Contributor checklist

- [ ] Includes tests
- [ ] Includes typings

### DIFF
--- a/lib/models/release.coffee
+++ b/lib/models/release.coffee
@@ -14,12 +14,13 @@ See the License for the specific language governing permissions and
 limitations under the License.
 ###
 
+isEmpty = require('lodash/isEmpty')
 once = require('lodash/once')
 omit = require('lodash/omit')
 errors = require('balena-errors')
 Promise = require('bluebird')
 
-{ findCallback, mergePineOptions } = require('../util')
+{ isId, findCallback, mergePineOptions } = require('../util')
 
 getReleaseModel = (deps, opts) ->
 	{ pine } = deps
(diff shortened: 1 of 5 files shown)


lib/models/release.coffee
```diff
@@ -55,21 +56,46 @@ getReleaseModel = (deps, opts) ->
...
-			id: id
-			options: mergePineOptions({}, options)
-		.tap (release) ->
-			if not release?
-				throw new errors.BalenaReleaseNotFound(id)
+		return Promise.try ->
+			if not commitOrId?
+				throw new errors.BalenaReleaseNotFound(commitOrId)
+
+			if isId(commitOrId)
```
pdcastro: Are "short" commit hashes acceptable in "commitOrId"? Because commit hashes are hex strings, and it's not "that rare" for a 7-char hex string to be composed of decimal digits and thus return true as `isId`.

thgreasi:
@pdcastro the sdk doesn't yet run on balena-ci since we atm need secrets :(
AppVeyor tests do pass, so this should be fine:
image

lib/models/release.coffee
thgreasi: @pdcastro `isId` is using `_.isNumber` which returns false for strings, so an all numeric short commit hash will not be used as a release id.
I still see your concern, but I will make sure to handle this in the CLI (next Friday probably).

pdcastro: Cool 👍

tests/integration/models/release.spec.coffee
```diff
@@ -68,6 +76,26 @@ describe 'Release Model', ->
...
+				.then (release) =>
+					m.chai.expect(release).to.deep.match
+						status: 'success',
+						source: 'cloud',
+						commit: 'new-release-commit',
+						id: @currentRelease.id
+						belongs_to__application: __id: @application.id
+
+			it 'should get the requested release by shorter commit', ->
+				balena.models.release.get(@currentRelease.commit.slice(0, 8))
```
sradevski: In the dashboard we show 7 characters of the commit hash, it's not that important for the tests but we should probably have a common method for getting a short hash for a commit (already did that in an upcoming staged releases PR for the dashboard). Not a blocker, I just thought I should mention it.

Issue #727, opened 2019-07-30
Should allow transfering aarch64 devices to armv7hf apps


thgreasi:
Placeholder issue.
See: https://www.flowdock.com/app/rulemotion/r-beginners/threads/A4a3CALEDol4WIUXZZIlv-ZD3v0

PR #728, opened 2019-07-30
Allow moving aarch64 devices to armv7hf applications

thgreasi:
Also adds an `os.isArchitectureCompatibleWith()` method so that consumers can rely on the SDK instead of repeating this logic.

Resolves: #727
See: https://www.flowdock.com/app/rulemotion/r-beginners/threads/A4a3CALEDol4WIUXZZIlv-ZD3v0
Change-type: minor

---
##### Contributor checklist

- [x] Includes tests
- [x] Includes typings

### DIFF
--- a/lib/models/device.coffee
+++ b/lib/models/device.coffee
@@ -36,6 +36,7 @@ deviceStatus = require('balena-device-status')
 	getCurrentServiceDetailsPineOptions
 	getOsUpdateHelper: _getOsUpdateHelper
 	generateCurrentServiceDetails
+	deviceTypes: deviceTypesUtil
 	mergePineOptions
 	notFoundResponse
 	noDeviceForKeyResponse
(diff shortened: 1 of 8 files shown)


lib/models/os.coffee
```diff
@@ -58,16 +58,10 @@ getOsModel = (deps, opts) ->
...
+	getValidatedDeviceType = (deviceTypeSlug) ->
 		getDeviceTypes()
 		.then (types) ->
-			!!deviceTypesUtil.findBySlug(types, deviceType)
-
-	validateDeviceType = (deviceType) ->
-		isValidDeviceType(deviceType)
-		.then (isValid) ->
-			if not isValid
-				throw new errors.BalenaInvalidDeviceType('No such device type')
```
thgreasi: Since I had to do something like this in  the `device.coffee` I grabbed the opportunity and unified it in a single function in the device type helper (which is also in TS).

lib/util/device-types.ts
```diff
@@ -9,14 +11,24 @@ const find = <T>(array: T[], predicate: (el: T) => boolean) => {
 	}
 };
 
-const includes = <T>(array: T[], el: T) => array && array.indexOf(el) >= 0;
+const includes = <T>(array: T[] | undefined, el: T) =>
+	array != null && array.indexOf(el) >= 0;
```
thgreasi: I will refactor this at a later point of time, along with splitting all util imports and exposing some of them.

lib/models/device.coffee
```diff
@@ -873,10 +874,9 @@ getDeviceModel = (deps, opts) ->
 			deviceTypes: configModel().getDeviceTypes()
 			application: applicationModel().get(applicationNameOrId, $select: [ 'id', 'device_type' ])
 		.then ({ application, device, deviceTypes }) ->
-			deviceDeviceType = find(deviceTypes, { slug: device.device_type })
-			appDeviceType = find(deviceTypes, { slug: application.device_type })
```
thgreasi: This would throw an undefined exception if a user would try to move a device and either of the device type slugs are non-existing (the validation hook on the API was added at a later point of time).

PR #731, opened 2019-08-21
Add typings & docs for the captcha parameters

thgreasi:
Change-type: minor

---
##### Contributor checklist

- [ ] Includes tests
- [x] Includes typings

### DIFF
--- a/lib/auth.coffee
+++ b/lib/auth.coffee
@@ -340,6 +340,7 @@ getAuth = (deps, opts) ->
 	# @param {Object} [credentials={}] - in the form of username, password and email
 	# @param {String} credentials.email - the email
 	# @param {String} credentials.password - the password
+	# @param {(String|undefined)} [credentials.'g-recaptcha-response'] - the captcha response
 	#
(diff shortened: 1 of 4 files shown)


lib/auth.coffee
```diff
@@ -340,6 +340,7 @@ getAuth = (deps, opts) ->
 	# @param {Object} [credentials={}] - in the form of username, password and email
 	# @param {String} credentials.email - the email
 	# @param {String} credentials.password - the password
+	# @param {(String|undefined)} [credentials.'g-recaptcha-response'] - the captcha response
```
thgreasi: I was originally trying to do `[billingInfo['g-recaptcha-response']]` or `[billingInfo["g-recaptcha-response"]]`, but jsdoc didn't like them and thought it was a separate argument. After some googling it seems that there is an ongoing discussion on this.

sradevski: Interesting, I didn't notice that TBH, I just assumed that is the syntax.

Issue #732, opened 2019-08-28
Should be able to enable/disable device local mode


thgreasi:
See: https://github.com/balena-io/balena-cli/issues/1304

PR #733, opened 2019-08-28
Add local mode related methods

thgreasi:
Resolves: #732
See: https://github.com/balena-io/balena-cli/issues/1304
Change-type: minor

---
##### Contributor checklist

- [x] Includes tests
- [x] Includes typings

### DIFF
--- a/lib/models/device.coffee
+++ b/lib/models/device.coffee
@@ -49,6 +49,12 @@ deviceStatus = require('balena-device-status')
 	getDeviceOsSemverWithVariant
 	normalizeDeviceOsVersion
 } = require('../util/device-os-version')
+{
+	checkLocalModeSupported
+	getLocalModeSupport
+	LOCAL_MODE_ENV_VAR
+	LOCAL_MODE_SUPPORT_PROPERTIES
+} = require('../util/local-mode')
 
 # The min version where /apps API endpoints are implemented is 1.8.0 but we'll
 # be accepting >= 1.8.0-alpha.0 instead. This is a workaround for a published 1.8.0-p1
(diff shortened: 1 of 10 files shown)


lib/models/device.coffee
```diff
@@ -49,6 +49,12 @@ deviceStatus = require('balena-device-status')
 	getDeviceOsSemverWithVariant
 	normalizeDeviceOsVersion
 } = require('../util/device-os-version')
+{
+	checkLocalModeSupported
+	getLocalModeSupport
+	LOCAL_MODE_ENV_VAR
+	LOCAL_MODE_SUPPORT_PROPERTIES
+} = require('../util/local-mode')
```
thgreasi: Chose to implement this in TS and just re-export the methods.

sradevski: I think that's the right approach and we should try to do that whenever possible.

lib/models/device.coffee
```diff
@@ -1853,6 +1859,119 @@ getDeviceModel = (deps, opts) ->
...
+	# 	}
+	# });
+	###
+	exports.isInLocalMode = (uuidOrId, callback) ->
+		exports.configVar.get(uuidOrId, LOCAL_MODE_ENV_VAR)
+		.then (value) -> value == '1'
+		.asCallback(callback)
+
+	###*
+	# @summary Returns whether local mode is supported along with a message
```
sradevski:
```suggestion
	# @summary Returns whether local mode is supported along with a message describing the reason why local mode is not supported.
```
Feel free to modify it, but it would be nice to have a description what the message is.

lib/models/device.coffee
thgreasi: I agree @sradevski . Atm there a big `util/index` which also re-exports all other utils and the models use that, which makes me sad :(
I indeed want to split lots of them and use explicit imports for (almost) all of them, which
* can make the CLI load faster, since it will be able to avoid some dependencies of the utils
* possibly allow us to use `util/index` for exporting some functions to the SDK consumers.

PR #741, opened 2019-09-25
tests/config: Update the mixpanelToken expected value

thgreasi:
Change-type: patch
See: https://github.com/balena-io/resin-api/pull/2108

---
##### Contributor checklist

- [x] Includes tests
- [ ] Includes typings


thgreasi:
To clarify, the failing tests are b/c of an API regression which will be soon be fixed.

roman-mazur:
@thgreasi no really

The test here assumes that exposed token must have 32 characters. This is no longer true, so this change needs to be changed. Empty value is no longer returned by API, but it's also a different token.

thgreasi:
@roman-mazur I meant that the remaining failing test are b/c of an API regression that will soon be fixed.

PR #746, opened 2019-10-01
Convert the device service details utils to TypeScript

thgreasi:
Change-type: patch

Resolves: # 
HQ: <url> 
See: <url> 
Depends-on: <url> 

---
##### Contributor checklist

- [ ] Includes tests
- [ ] Includes typings

### DIFF
--- a/lib/util/device-service-details.ts
+++ b/lib/util/device-service-details.ts
@@ -1,0 +1,135 @@
+import groupBy = require('lodash/groupBy');
+import mapValues = require('lodash/mapValues');
+import omit = require('lodash/omit');
+import {
+	CurrentGatewayDownload,
+	CurrentService,
+	DeviceWithImageInstalls,
+	DeviceWithServiceDetails,
+	GatewayDownload,
+	Image,
+	ImageInstall,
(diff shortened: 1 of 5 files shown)


lib/util/device-service-details.ts
```diff
@@ -0,0 +1,135 @@
...
+	const service = (image.is_a_build_of__service as Service[])[0];
+
+	let releaseInfo: { commit?: string } = {};
+	if (
+		'is_provided_by__release' in rawData &&
+		rawData.is_provided_by__release != null
+	) {
+		const release = (rawData.is_provided_by__release as Release[])[0];
+		releaseInfo = {
+			commit: release != null ? release.commit : void 0,
```
sradevski: I believe using `undefined` is fine as well for all the browsers we support and non-deprecated NodeJS versions, but I am fine with it as it is as well.

thgreasi: Ooops. I missed that part of decaffeinate :sweat_smile: 
I will fix it in the follow-up PR, so save some time.
Thanks for spotting it :+1:

Issue #745, opened 2019-09-30
Should optimize the generateCurrentServiceDetails utility method


thgreasi:
See: https://www.flowdock.com/app/rulemotion/resin-frontend/threads/NWAfTZlqZC5ghWtz1KO2o8svs54

PR #748, opened 2019-10-03
Optimize the app & device methods that fill service details

thgreasi:
- [x] That's on top of #746, so let's get that in first.

Resolves: #745
Change-type: minor

---
##### Contributor checklist

- [ ] Includes tests
- [ ] Includes typings

### DIFF
--- a/lib/util/device-service-details.ts
+++ b/lib/util/device-service-details.ts
@@ -1,9 +1,7 @@
-import groupBy = require('lodash/groupBy');
-import mapValues = require('lodash/mapValues');
-import omit = require('lodash/omit');
 import {
 	CurrentGatewayDownload,
 	CurrentService,
+	Device,
 	DeviceWithImageInstalls,
 	DeviceWithServiceDetails,
 	GatewayDownload,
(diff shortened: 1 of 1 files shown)


lib/util/device-service-details.ts
```diff
@@ -110,26 +119,37 @@ export const generateCurrentServiceDetails = (
...
+
+		// remove the extra property that we added for the grouping
+		delete container.service_name;
+		serviceContainerGroup.push(container);
+
+		return acc;
+	}, {});
+
+	// TODO: change to Object.values() in a later major release
+	Object.keys(currentServicesGroupedByName).forEach(serviceName =>
```
Page-: For performance it's probably better to do
```suggestion
	for (const serviceName in currentServicesGroupedByName) {
```
rather than creating the intermediary array which will later need to be GCed

lib/util/device-service-details.ts
```diff
@@ -110,26 +119,37 @@ export const generateCurrentServiceDetails = (
...
-		},
+	// prefer over omit for performance reasons
+	delete rawDevice.image_install;
+	delete rawDevice.gateway_download;
+
+	const device = (rawDevice as Device) as DeviceWithServiceDetails;
+
+	// Essentially a groupBy(installs, 'service_name')
+	// but try making it a bit faster for the sake of large fleets
+	const currentServicesGroupedByName = installs.reduce<
```
Page-: I'd switch this block to
```ts
	const currentServicesGroupedByName: Record<string, CurrentService[]> = {};
	installs.forEach((container) => {
		let serviceContainerGroup = currentServicesGroupedByName[container.service_name];
		if (!serviceContainerGroup) {
			serviceContainerGroup = currentServicesGroupedByName[container.service_name] = [];
		}

		// remove the extra property that we added for the grouping
		delete container.service_name;
		serviceContainerGroup.push(container);
	});
```
since the ability to pass the accumulator forward is unused except to keep the original object in scope, which is easily done via it being a closure anyway

lib/util/device-service-details.ts
```diff
@@ -82,16 +86,21 @@ function getSingleInstallSummary(
...
 		const release = (rawData.is_provided_by__release as Release[])[0];
 		releaseInfo = {
-			commit: release != null ? release.commit : void 0,
+			commit: release != null ? release.commit : undefined,
 		};
 	}
 
+	// prefer over omit for performance reasons
+	delete rawData.image;
+	if ('is_provided_by__release' in rawData) {
```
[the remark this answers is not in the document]
thgreasi: As discussed in FD, this is so that TS is happy about `is_provided_by__release` possibly being `?` in the  `CurrentService | CurrentGatewayDownload` union.

PR #749, opened 2019-10-04
Add release.createFromUrl method

thgreasi:
Change-type: minor

Resolves: # 
HQ: <url> 
See: <url> 
Depends-on: <url> 

---
##### Contributor checklist

- [x] Includes tests
- [x] Includes typings

### DIFF
--- a/lib/util/builder.ts
+++ b/lib/util/builder.ts
@@ -1,0 +1,56 @@
+import * as errors from 'balena-errors';
+import * as BalenaSdk from '../../typings/balena-sdk';
+import { InjectedDependenciesParam, InjectedOptionsParam } from '../balena';
+
+type BalenaBuilderRequestError = errors.BalenaError & {
+	body: BuilderBuildFromUrlErrorResponse;
+};
+
+interface BuilderBuildFromUrlResponse {
+	started: true;
+	releaseId: number;
+}
+
+interface BuilderBuildFromUrlErrorResponse {
+	started: false;
+	error: string;
+	message: string;
+}
+
+const isBuilderError = (error: any): error is BalenaBuilderRequestError =>
+	error.code === 'BalenaRequestError' &&
+	typeof error.body === 'object' &&
+	!!error.body.error;
+
+export class BuilderHelper {
+	constructor(
+		private deps: InjectedDependenciesParam,
+		private opts: InjectedOptionsParam,
+	) {}
+
+	public buildFromUrl(
+		owner: string,
+		appName: string,
+		urlDeployOptions: BalenaSdk.BuilderUrlDeployOptions,
+	) {
+		return this.deps.request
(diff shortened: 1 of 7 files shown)


lib/util/builder.ts
```diff
@@ -0,0 +1,48 @@
...
+		urlDeployOptions: BalenaSdk.BuilderUrlDeployOptions,
+	) {
+		return this.deps.request
+			.send<BuilderBuildFromUrlResponse>({
+				method: 'POST',
+				url: `/v3/buildFromUrl?headless=true&owner=${owner}&app=${appName}`,
+				baseUrl: this.opts.builderUrl,
+				body: Object.assign(
+					{
+						shouldFlatten: true,
```
sradevski: Tarballs from github, once unpackaged, give you a single folder, with the contents inside that folder. If a user passes a tarball that directly wraps the content, this should be false. I would suggest adding this to the `urlDeployOptions` jsdoc signature, and just set it to `true` by default.

lib/util/builder.ts
```diff
@@ -0,0 +1,48 @@
...
+					},
+					urlDeployOptions,
+				),
+			})
+			.then(resp => {
+				const data = resp.body;
+				if ('error' in data) {
```
sradevski: NIT: in most places we would do `data[error]`

tests/integration/models/release.spec.coffee
```diff
@@ -60,6 +61,55 @@ describe 'Release Model', ->
...
+			TEST_SOURCE_URL = 'https://github.com/balena-io-projects/simple-server-node/archive/v1.0.0.tar.gz'
+
+			it 'should be rejected if the application name does not exist', ->
+				promise = balena.models.release.createFromUrl('HelloWorldApp', { url: TEST_SOURCE_URL })
+				m.chai.expect(promise).to.be.rejectedWith('Application not found: HelloWorldApp')
+
+			it 'should be rejected if the application id does not exist', ->
+				promise = balena.models.release.createFromUrl(999999, { url: TEST_SOURCE_URL })
+				m.chai.expect(promise).to.be.rejectedWith('Application not found: 999999')
+
```
sradevski: Maybe add one test where the url is invalid (we already have tests in the builder for such cases, but just for completeness)

lib/util/builder.ts
thgreasi: That was mostly for TS to be able to reason about the union types, but I will change it to check for truthiness :+1:

tests/integration/models/release.spec.coffee
thgreasi: Great idea :+1:

lib/util/builder.ts
thgreasi: After adding the tests, it seems this only has to be inside a `.catch` since balena-request will reject all promises with non 2xx statusCode.

Issue #751, opened 2019-10-07
Should add support for private device types


thgreasi:
HQ: https://github.com/balena-io/balena/pull/1740

PR #752, opened 2019-10-07
Support retrieving private device types

thgreasi:
Resolves: #751
HQ: balena-io/balena#1740
Change-type: minor

---
##### Contributor checklist

- [x] Includes tests
- [ ] Includes typings

### DIFF
--- a/lib/auth.coffee
+++ b/lib/auth.coffee
@@ -18,7 +18,7 @@ errors = require('balena-errors')
 Promise = require('bluebird')
 
 getAuth = (deps, opts) ->
-	{ auth: authBase, request} = deps
+	{ auth: authBase, pubsub, request } = deps
 	{ apiUrl } = opts
 
 	exports = {}
@@ -31,23 +31,14 @@ getAuth = (deps, opts) ->
 		else
 			return err
 
-	authSubscribers = []
-
-	exports._onChange = (fn) ->
-		authSubscribers.push(fn)
-		# an unsubscribe function
-		return ->
-			indexOfFn = authSubscribers.indexOf(fn)
-			authSubscribers.splice(indexOfFn, 1)
-
-	wrapAuthFn = (fn) ->
+	wrapAuthFn = (eventName, fn) ->
 		return ->
 			fn.apply(authBase, arguments)
-			.finally -> Promise.map(authSubscribers, (fn) -> fn())
+			.finally -> pubsub.publish(eventName)
 
 	auth = Object.assign({}, authBase, {
-		setKey: wrapAuthFn(authBase.setKey)
-		removeKey: wrapAuthFn(authBase.removeKey)
+		setKey: wrapAuthFn('auth.keyChange', authBase.setKey)
+		removeKey: wrapAuthFn('auth.keyChange', authBase.removeKey)
 	})
 
 	twoFactor = require('./2fa')(Object.assign({}, deps, { auth }), opts)
--- a/lib/auth.coffee
+++ b/lib/auth.coffee
@@ -18,9 +18,9 @@ errors = require('balena-errors')
 Promise = require('bluebird')
 
 getAuth = (deps, opts) ->
-	{ auth, request} = deps
+	{ auth: authBase, request} = deps
 	{ apiUrl } = opts
-	twoFactor = require('./2fa')(deps, opts)
+
 	exports = {}
 
 	normalizeAuthError = (err) ->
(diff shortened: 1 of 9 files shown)


lib/models/config.coffee
```diff
@@ -100,7 +100,6 @@ getConfigModel = (deps, opts) ->
 			method: 'GET'
 			url: '/device-types/v1'
```
sradevski: I assume there are no breaking changes in the API for private types, right?

lib/auth.coffee
```diff
@@ -31,6 +31,27 @@ getAuth = (deps, opts) ->
 		else
 			return err
 
+	authSubscribers = []
+
+	exports._onChange = (fn) ->
+		authSubscribers.push(fn)
+		# an ubsubscribe function
```
sradevski:
```suggestion
		# an unsubscribe function
```

lib/auth.coffee
```diff
@@ -31,6 +31,27 @@ getAuth = (deps, opts) ->
 		else
 			return err
 
+	authSubscribers = []
+
+	exports._onChange = (fn) ->
```
sradevski: I don't really understand the logic behind this, what's the purpose of keeping a list of subscribers? Isn't there only one authenticated client at a time?

lib/models/os.coffee
```diff
@@ -42,26 +42,38 @@ BALENAOS_VERSION_REGEX = /v?\d+\.\d+\.\d+(\.rev\d+)?((\-|\+).+)?/
...
 getOsModel = (deps, opts) ->
-	{ request } = deps
+	{ request, sdkInstance: { auth } } = deps
 	{ apiUrl, isBrowser } = opts
 
 	configModel = once -> require('./config')(deps, opts)
 	applicationModel = once -> require('./application')(deps, opts)
 
-	getDeviceTypes = once ->
+	withDeviceTypesEndpointCaching = (fn) ->
```
sradevski: Why do we memoize the function and not the result from it, if it succeeds?

lib/auth.coffee
thgreasi: @sradevski There is indeed only one auth client per SDK instance.
The purpose of this is to provide a way to notify the appropriate caches of the sdk to become invalidated.
This way the auth modules is decoupled from the caches of the various modules that need to be cleared.
I might be missing something, so do you propose something different?

lib/models/config.coffee
thgreasi: No there are no breaking changes.
We were already able to send the auth token but that didn't have any effect so far, so it wasn't sent, so that we are explicit that it isn't required.
Now the auth token is optional, when present it's used to resolve the user permissions on private device types.
Let me add a comment about this.

lib/models/os.coffee
thgreasi: The package used here already caches the result only when the promise gets resolved.
We do follow the same pattern in other top components as well and it saves us from
* manually denouncing calls while there is already one ongoing, in which case it returns the same ongoing promise
* having a manual `If (cached) { return cached; }`

Is there a reason that you think we should be in favor of doing the above two on our own code?

lib/auth.coffee
sradevski: @thgreasi Oh now I get it, you are creating a very simple event emitter :). Basically, any model can subscribe to the auth events with a callback, and when an auth change event occurs the subscribers will be notified about it. 

I guess we could go with a proper event handling library, but it might be an overkill for now. I would only suggest to rename `onChange` to `subscribeToEvent` that takes two arguments: event name, and a callback function. I guess for now we can only have a single `eventName` that will be `change` or `*` and the rest will be the same. What do you think?

lib/auth.coffee
```diff
@@ -31,6 +31,18 @@ getAuth = (deps, opts) ->
 		else
 			return err
 
+	wrapAuthFn = (eventName, fn) ->
+		return ->
+			fn.apply(authBase, arguments)
+			.finally -> pubsub.publish(eventName)
+
+	auth = Object.assign({}, authBase, {
+		setKey: wrapAuthFn('auth.keyChange', authBase.setKey)
```
sradevski: NIT: Extract the event name to a constant and use it when subscribing as well (similar to Redux actions). You can even put it in a single file that exports all events so it is also easy to keep track of what events are supported for pubsub.

thgreasi: Let's keep this for when we add a second supported event, should we?

PR #768, opened 2019-10-31
Move util index file to Typescript

sradevski:
---
##### Contributor checklist

- [ ] Includes tests
- [ ] Includes typings


lib/models/api-key.ts
```diff
@@ -115,7 +115,9 @@ const getApiKeysModel = function(
 	 */
 	exports.getAll = function(
 		options: BalenaSdk.PineOptionsFor<BalenaSdk.ApiKey> = {},
-		callback: (error?: Error, apiKeys?: BalenaSdk.ApiKey[]) => void,
+		callback:
```
Page-: I think this should be
```suggestion
		callback?: (error?: Error, apiKeys?: BalenaSdk.ApiKey[]) => voi
```
so that can omit it as well as explicitly specifying `undefined`

lib/util/index.ts
```diff
@@ -0,0 +1,270 @@
...
+import * as dt from './device-types';
+
+export const getOsUpdateHelper = updateHelper;
+export const deviceTypes = dt;
+
+export const notImplemented = () => {
+	throw new Error('The method is not implemented.');
+};
+
+export const onlyIf = (condition: boolean) => (fn: () => void) => {
```
Page-: You can make this
```suggestion
export const onlyIf = (condition: boolean) => {
	if (condition) {
		return _.identity
	} else {
		return _.constant(notImplemented)
	}
}
```
so that the condition is only checked once

lib/util/index.ts
```diff
@@ -0,0 +1,270 @@
...
+export const noApplicationForKeyResponse = {
+	code: 'BalenaRequestError',
+	statusCode: 500,
+	body: 'No application found to associate with the api key',
+};
+
+export const isUniqueKeyViolationResponse = ({
+	code,
+	body,
+}: {
```
Page-: Probably worth making an interface for these custom error objects and using it here

lib/util/index.ts
thgreasi: That's clever :) But tbh I would prefer to keep the lodash imports to the minimum, so that we can eventually drop it.

Page-: Well the equivalent plain js would be
```suggestion
export const onlyIf = (condition: boolean) => {
	if (condition) {
		return (x) => x
	} else {
		return() => notImplemented
	}
}

PR #769, opened 2019-11-04
Add ESR versions to getSupportedVersions function result

sradevski:
Change-type: minor

---
##### Contributor checklist

- [ ] Includes tests
- [ ] Includes typings

### DIFF
--- a/tests/integration/models/os.spec.coffee
+++ b/tests/integration/models/os.spec.coffee
@@ -206,6 +206,12 @@ describe 'OS model', ->
 
 				return true
 
+			areValidEsrVersions = (osVersions) ->
+				m.chai.expect(osVersions).to.have.property('esr').that.is.an('object')
+				areValidVersions(osVersions.esr.next)
+				areValidVersions(osVersions.esr.current)
+				areValidVersions(osVersions.esr.sunset)
+
 			it 'should eventually return the valid versions object', ->
 				promise = balena.models.os.getSupportedVersions('raspberry-pi')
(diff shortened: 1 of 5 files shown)


tests/integration/models/os.spec.coffee
```diff
@@ -235,6 +241,15 @@ describe 'OS model', ->
 				promise = balena.models.os.getSupportedVersions('foo-bar-baz')
 				m.chai.expect(promise).to.be.rejectedWith('No such device type')
 
+		describe 'given a user that supports ESR', ->
+			# TODO: This should be production plan user
```
sradevski: @thgreasi there is also a paid user envvar, but for ESR we only need production or enterprise plan users. shall we create another instance of a paid user, or shall we make sure that the paid user passed is actually on a production/enterprise plan?

thgreasi: @sradevski The paid user that we currently have is for a $0 test plan. Having a test user on a production/enterprice plan probably needs some discussion about the analytics & billing implications.

We could definitely have such a user on staging and run the respective tests when the appropriate env vars are available.

sradevski: @thgreasi if we can enable ESR for the test plan, then that would work as well. Is that possible?

thgreasi: I haven't seen the API code for the ESR feature, but that should work as well.

sradevski:
Closing this as the implementation details completely changed. Will reopen a new PR once the changes are final.

PR #770, opened 2019-11-05
Add typings tests for the pine options

thgreasi:
Change-type: patch

---
##### Contributor checklist

- [x] Includes tests
- [ ] Includes typings


typing_tests/pine-options.ts
```diff
@@ -0,0 +1,624 @@
...
+				},
+			},
+		},
+	},
+};
+
+export const appOptionsEFValid1: BalenaSdk.PineOptionsFor<
+	BalenaSdk.Application
+> = {
+	$expand: [
```
sradevski: Would it work if we have a single (or 2-3) complete query with all possible combinations inside it, and we just add multiple error assertions there? That should dramatically reduce the amount of code, and still achieve the same result.

thgreasi: No unfortunately not :(
TS either
* reports the error on the outermost parent (you can see the `nestedArray$expandWithManyErrors` tests)
* or stops further reporting errors for the expression as soon as an error is found, which makes the rest expectations to error with `Expected an error on this line, but found none.`

thgreasi: Except if I'm missing/forgetting something, since this needed considerable effort, in which case PRs are of course welcome :)

thgreasi:
@balena-ci retest

Issue #764, opened 2019-10-28
Should prohibit pine requests w/o a $select


thgreasi:
See: https://www.flowdock.com/app/rulemotion/resin-tech/threads/fizOBvTINENLFKlLXTHczcbazGT
See: https://github.com/balena-io/resin-api/pull/2227

thgreasi:
We ended up not making this mandatory from the user's side. Closing.

PR #773, opened 2019-11-13
Make $select required in pine GET requests and $filter or ID required in PATCH & DELETE

thgreasi:
That's on top of https://github.com/balena-io/balena-sdk/pull/772, so let's get that in first.

In v6 $select will be required in all GET requests, so we now have to hard-code $select all fields available in open-balena, to all the GET requests found in the available methods.
If we also added non-open-balena fields, then requests to an open-balena instance would fail with 400.

Resolves: #764
Change-type: major
See: balena-io/balena-api#2226

---
##### Contributor checklist

- [x] Includes tests
- [x] Includes typings

### DIFF
--- a/lib/models/release.coffee
+++ b/lib/models/release.coffee
@@ -22,6 +22,27 @@ Promise = require('bluebird')
 
 { isId, findCallback, mergePineOptions } = require('../util')
 
+ReleaseFields = [
+	'id'
+	'created_at'
+	'commit'
+	'composition'
+	'status'
+	'source'
+	'build_log'
+	'start_timestamp'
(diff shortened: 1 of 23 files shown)


lib/models/release.coffee
```diff
@@ -389,6 +409,12 @@ getReleaseModel = (deps, opts) ->
 			$expand:
 				release_tag:
 					mergePineOptions
+						$select: [
```
sradevski: Shall we extract this in a variable for consistency?

thgreasi: Sure :+1:

PR #778, opened 2019-11-18
Upgrade Typescript to v3.7.2

thgreasi:
Just so that we can start using the new 3.6 & 3.7 syntaxt & recursive typings w/o worrying supporting versions all the way back to v3.4.0.

Change-type: major

---
##### Contributor checklist

- [ ] Includes tests
- [ ] Includes typings

### DIFF
--- a/package.json
+++ b/package.json
@@ -32,7 +32,7 @@
     "test:all": "npm run test:ts-compatibility && npm run test:typings && npm run test:node && npm run test:browser",
     "test:node": "gulp test",
     "test:browser": "karma start",
-    "test:ts-compatibility": "npx typescript@~3.4.0 --noEmit --project ./tsconfig.dist.json",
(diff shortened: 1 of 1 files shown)


thgreasi:
@sradevski actually, since there is no lockfile, we can already use TS 3.7 features in `master`.

On the other hand I'm actually in favor of supporting older TS versions as a major change for the SDK, and that's why we do run `npx typescript@~3.4.0`. This way we be confident that the SDK typings work both with the latest TS 3.x version as well as the one we were committed for the current major (TS 3.4 that is for SDK v12).
This has been discussed before, in the era that TS did a few breaking changes in minor version bumps (eg v2.8) and as a way to avoid breaking the CLI's CI tests all the time (both b/c of just TS, as well as b/c of the typings of our dependencies), we ended up not bumping TS so often.

In short, I'm fine with changing `"typescript": "^3.4.5",` in `master` but keeping it for testing the typings, but on the other hand this will not benefit us at all, since `npm i` is anyway fetching the latest 3.x version.

PR #780, opened 2019-11-18
Drop the imageMakerUrl & apiKey constructor option

thgreasi:
* The `imageMakerUrl` was deprecated and no longer used since v12.12.0, so it's just removing it from the docs.
* For the `apiKey`, event though it was deprecated, it seems that balena-request was still picking this up from the constructor options, in case a use was still providing it. So we are now actively preventing that to be passed down. (The equivalent suggested method is `sdk.auth.loginWithToken()`.)

Change-type: major

Resolves: # 
HQ: <url> 
See: <url> 
Depends-on: <url> 

---
##### Contributor checklist

- [ ] Includes tests
- [ ] Includes typings


thgreasi:
@sradevski yea we can definitely remove the `apiKey` as well.
After further investigation, I think I'm also going to open a separate PR to drop the pre-mc supervisor actions.

Issue #776, opened 2019-11-14
Should add the sdk version to a request header


thgreasi:
Discussed in arch call 2019-11-14.
HQ: https://github.com/balena-io/balena/issues/2036

PR #781, opened 2019-11-21
Add the SDK version as a property on the instance & in the request headers

thgreasi:
I highly suggest reviewing this, commit by commit, at least the first one that's just re-running prettier.
The node tests pass fine even w/o the API changes, but browser need the new header to appear in the CORS OPTIONS request.

Resolves: #776
HQ: https://github.com/balena-io/balena/issues/2036
Depends-on: https://github.com/balena-io/open-balena-api/issues/222 (done)
Depends-on: https://github.com/balena-io/open-balena-api/pull/223
Depends-on: https://github.com/balena-io/balena-api/issues/2269 (done)
Depends-on: https://github.com/balena-io/balena-api/pull/2273
Change-type: minor

---
##### Contributor checklist

- [x] Includes tests
- [x] Includes typings


lib/balena.coffee
```diff
@@ -159,6 +160,23 @@ getSdk = (opts = {}) ->
...
+			url = request.url
+
+			if typeof url != 'string'
+				return request
+
+			if typeof request.baseUrl == 'string'
+				url = request.baseUrl + url
+
+			if url.indexOf(opts.apiUrl) == 0
+				request.headers['X-Balena-Client'] = "balena-sdk v#{version}"
```
thgreasi: I'm not sure whether it would make more sense as `balena-sdk/1.2.3` similarly to how `node-fetch/1.0` does it :thinking:

thgreasi: Yea... I'm changing this to use `/`...

gulpfile.coffee
```diff
@@ -45,15 +46,19 @@ gulp.task 'test', ->
 		}))
 
 gulp.task 'build', (callback) ->
-	runSequence('build-node', 'build-browser', callback)
+	runSequence('build-node', 'pack-browser', callback)
 
+gulp.task 'inject-version', ->
+	gulp.src(["#{OPTIONS.directories.build}util/sdk-version.js"])
+		.pipe(replace("var sdkVersion = '0.0.0';", "var sdkVersion = '#{packageJSON.version}';"))
```
thgreasi: // TODO: use a better build step for the node sources, that supports this kind of stuff.
Eg: rollup

afitzek: You could use some magic value for the version string like `__REPLACE_WITH_SDK_VERSION__` and replace this string instead of the whole line, since I guess it depends on the ts to js compilation?
```
.pipe(replace("__REPLACE_WITH_SDK_VERSION__", "#{packageJSON.version}"))
```

thgreasi:
Sure @afitzek

thgreasi:
@dfunckt just because `User-Agent` for the browser to set it.

PR #782, opened 2019-11-22
Tighten the types of the JWT's `social_service_account` entry

Page-:
Change-type: patch

Resolves: # 
HQ: <url> 
See: <url> 
Depends-on: <url> 

---
##### Contributor checklist

- [ ] Includes tests
- [ ] Includes typings

### DIFF
--- a/typings/balena-sdk.d.ts
+++ b/typings/balena-sdk.d.ts
@@ -171,11 +171,6 @@ declare namespace BalenaSdk {
 	interface SocialServiceAccount {
 		provider: string;
 		display_name: string;
-		created_at: string;
-		id: number;
-		remote_id: string;
-
-		belongs_to__user: NavigationResource<User>;
 	}
 
 	interface User {


typings/balena-sdk.d.ts
```diff
@@ -207,7 +207,7 @@ declare namespace BalenaSdk {
 			ApplicationMember
 		>;
 		// this is what the api route returns
-		social_service_account: ReverseNavigationResource<SocialServiceAccount>;
+		social_service_account: ReverseNavigationResource<Pick<SocialServiceAccount, 'provider' | 'display_name'>>;
```
thgreasi: I think we could just drop all other props from SocialServiceAccount, instead of using a Pick here.

Page-: It's your call @thgreasi

thgreasi: @Page- I just confirmed this locally on the UI. I indeed prefer just having an accurate SocialServiceAccount instead of a Pick. Let me know if you prefer me to pick this up.

Page-: done

thgreasi:
@balena-ci retest

PR #785, opened 2019-11-27
Split the startOsUpdate method to improve testability

sradevski:
Change-type: patch

---
##### Contributor checklist

- [ ] Includes tests
- [ ] Includes typings

### DIFF
--- a/lib/models/device.coffee
+++ b/lib/models/device.coffee
@@ -2448,6 +2448,34 @@ getDeviceModel = (deps, opts) ->
 		hupActionHelper.getHUPActionType(device_type, currentOsVersion, targetOsVersion)
 		return
 
+	# TODO: This is a temporary solution for ESR, as the ESR-supported versions are not part of the SDK yet. It should be removed once the getSupportedVersions is updated to support ESR as well.
+	exports._startOsUpdate = (uuid, targetOsVersion, skipCheck, callback) ->
+		Promise.try ->
+			if not targetOsVersion
+				throw new errors.BalenaInvalidParameterError('targetOsVersion', targetOsVersion)
+
+			exports.get(uuid, $select: [
+				'device_type'
+				'is_online'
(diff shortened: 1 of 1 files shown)


lib/models/device.coffee
```diff
@@ -2448,6 +2448,34 @@ getDeviceModel = (deps, opts) ->
 		hupActionHelper.getHUPActionType(device_type, currentOsVersion, targetOsVersion)
 		return
 
+	# TODO: This is a temporary solution for ESR, as the ESR-supported versions are not part of the SDK yet. It should be removed once the getSupportedVersions is updated to support ESR as well.
+	exports._startOsUpdate = (uuid, targetOsVersion, skipCheck, callback) ->
```
thgreasi: How about defining just
```
	exports._startOsUpdate = (uuid, targetOsVersion) ->
		getOsUpdateHelper()
		.then (osUpdateHelper) ->
			osUpdateHelper.startOsUpdate(uuid, targetOsVersion)
```
and re-using this in the `device.startOsUpdate()` method?
It will for sure result cleaner code and less duplication.
If the reason it that we need to run just a portion of the test, then I would suggest to also use the `_checkOsUpdateTarget` in the UI code or even use the balena-hup-action-utils directly, since we already have them as a dependency.
This way the checks will live in the same place until we have proper ESR support in the SDK.

sradevski: @thgreasi What I was trying to achieve is, still have the check in most cases, but skip it for ESR only, and all of this without touching any existing code. Since I will be moving the ESR related methods to the SDK next week, I think this is the cleanest approach as then the whole _startOsUpdate can be removed.

thgreasi: On the other hand, if we had a `exports._startOsUpdate`, then we could leave it as is.
I will leave this up to you, since this is a private method anyway.

PR #793, opened 2019-12-26
Add missing typings to device type options

sradevski:
Change-type: minor

---
##### Contributor checklist

- [ ] Includes tests
- [ ] Includes typings

### DIFF
--- a/typings/balena-sdk.d.ts
+++ b/typings/balena-sdk.d.ts
@@ -164,8 +164,11 @@ declare namespace BalenaSdk {
 		name: string;
 		type: string;
 		min?: number;
+		max?: number;
+		hidden?: boolean;
+		when?: Dictionary<number | string | boolean>;
 		choices?: string[] | number[];
-		choicesLabels?: { [key: string]: string };
+		choicesLabels?: Dictionary<string>;
 	}
 
 	interface SocialServiceAccount {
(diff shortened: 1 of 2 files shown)


typings/balena-sdk.d.ts
```diff
@@ -164,8 +164,11 @@ declare namespace BalenaSdk {
 		name: string;
 		type: string;
 		min?: number;
+		max?: number;
+		hidden?: boolean;
+		when?: Dictionary<number | string | boolean>;
```
thgreasi: From what I could find in the `device-types/v1/` endpoint, this seems to always hold a string. Here are the values I could find on prod.
```
"coprocessorCore": "16",
"network": "wifi"
"os": "linux"
"os": "osx"
"os": "win32"
"processorCore": "quad"
"processorCore": "single"
"processorType": "Z7010"
"processorType": "Z7020"
```
I'm fine with it though.

sradevski: @thgreasi Things like `coprocessorCore` could easily be a number, and I am not sure how these are specified and validated. If there is no validation, it is safer to assume that the `when` condition can be any of them.

sradevski:
@thgreasi  can you update the appveyor status please? We should actually see if it is possible to remove appveyor and test using balenaCI, and what is stopping us from doing that.

Issue #799, opened 2020-01-20
User reports application.getAllWithDeviceServiceDetails is erroring


thgreasi:
```
TypeError: e.push is not a function
    at https://dashboard.balena-staging.com/vendor~scripts.e624d7aee1253a84c0cf.chunk.js:75:106313
    at Array.forEach (<anonymous>)
    at e.generateCurrentServiceDetails (https://dashboard.balena-staging.com/vendor~scripts.e624d7aee1253a84c0cf.chunk.js:75:106217)
    at Array.map (<anonymous>)
```
See: https://www.flowdock.com/app/rulemotion/resin-frontend/threads/r1bu5ypCACGLqz-nI2oufwXh7CJ

PR #800, opened 2020-01-21
Fix device.getWithServiceDetails to properly handle all service names

thgreasi:
Resolves: #799
Change-type: patch

Resolves: # 
HQ: <url> 
See: <url> 
Depends-on: <url> 

---
##### Contributor checklist

- [ ] Includes tests
- [ ] Includes typings

### DIFF
--- a/lib/util/device-service-details.ts
+++ b/lib/util/device-service-details.ts
@@ -126,24 +126,30 @@ export const generateCurrentServiceDetails = (
 	const device = (rawDevice as Device) as DeviceWithServiceDetails;
 
 	// Essentially a groupBy(installs, 'service_name')
-	// but try making it a bit faster for the sake of large fleets
-	const currentServicesGroupedByName: Record<string, CurrentService[]> = {};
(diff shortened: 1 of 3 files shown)


lib/util/device-service-details.ts
```diff
@@ -126,24 +126,30 @@ export const generateCurrentServiceDetails = (
...
-				container.service_name
-			] = [];
+	// but try making it a bit faster for the sake of large fleets.
+	// Uses Object.create(null) so that there are no inherited properties
+	// which could match service names
+	const currentServicesGroupedByName: Record<
+		string,
+		CurrentService[]
+	> = Object.create(null);
+	for (const container of installs) {
```
Page-: This can be
```suggestion
	for (const { service_name } of installs) {
```

thgreasi: Can't do @Page- , since we also need the whole `container` at the end.
PS: You had me for more than a moment, I almost typed this change before realizing... :)

PR #802, opened 2020-01-21
Update dependencies

Page-:
Change-type: patch

Resolves: # 
HQ: <url> 
See: <url> 
Depends-on: <url> 

---
##### Contributor checklist

- [ ] Includes tests
- [ ] Includes typings


package.json
```diff
@@ -54,69 +54,69 @@
...
     "uglify-es": "^3.3.9",
     "vinyl-source-buffer": "^1.1.1",
     "watch": "^1.0.2"
   },
   "dependencies": {
-    "@types/bluebird": "3.5.21",
-    "@types/common-tags": "^1.4.0",
+    "@types/bluebird": "^3.5.29",
+    "@types/common-tags": "^1.8.0",
+    "@types/lodash": "^4.14.149",
```
thgreasi: I don't think that `@types/lodash` has to be a prod dependency, since we do not have any such import in our typings.
Not a big deal though.

package.json
```diff
@@ -54,69 +54,69 @@
...
     "abortcontroller-polyfill": "^1.4.0",
-    "balena-auth": "^3.0.0",
-    "balena-device-status": "^3.1.0",
-    "balena-errors": "^4.1.0",
-    "balena-hup-action-utils": "~3.0.0",
-    "balena-pine": "^9.0.0",
+    "balena-auth": "^3.0.1",
+    "balena-device-status": "^3.2.0",
+    "balena-errors": "^4.2.1",
+    "balena-hup-action-utils": "^3.0.0",
```
thgreasi: The `~` here was intentional, since atm after each minor release, we wait for the proxy to pick-up the new rules.

package.json
Page-: lodash is a prod dependency so I think the types also should be, for safety if nothing else

package.json
Page-: So a minor update is actually a breaking update?

Page-: I've changed it back so as to not block but it still sounds problematic

Page-:
@balena-ci retest

Issue #795, opened 2020-01-15
Should be able to search an application by slug


thgreasi:
`application.get()` should also support searching by application slug.
This would allow simplifying other clients, like the CLI
See: https://github.com/balena-io/balena-cli/blob/a2ca8e8f7363788bec7e173a5c126bb0b78e753a/lib/utils/helpers.ts#L169-L191
See: https://www.flowdock.com/app/rulemotion/resin-frontend/threads/lgIBVoecxQ5MVpGcFaTQSrF3V6K

PR #804, opened 2020-01-22
Add support for operations by application slug

thgreasi:
This allows users to use the application slug in addition to the application name, as a way to disambiguate shared applications that have the same name in a more
user friendly way than using the application id.
The cli is already doing this in a few places, so this will allow us to remove such code.

The application slug
* on the cloud api equals `` `${owner}/${app_name}`.toLowerCase()``
* on open balena it's `app_name.toLowerCase()`

so we can be sure that the $or will never return more than one items when a slug is provided.

Resolves: #795
See: https://www.flowdock.com/app/rulemotion/resin-frontend/threads/lgIBVoecxQ5MVpGcFaTQSrF3V6K
Change-type: minor

---
##### Contributor checklist

- [x] Includes tests
- [ ] Includes typings

### DIFF
--- a/lib/models/application.coffee
+++ b/lib/models/application.coffee
@@ -52,7 +52,7 @@ getApplicationModel = (deps, opts) ->
 		resourceName: 'application_tag'
 		resourceKeyField: 'tag_key'
 		parentResourceName: 'application',
-		getResourceId: (nameOrId) -> exports.get(nameOrId, $select: 'id').get('id')
+		getResourceId: (nameOrSlugOrId) -> exports.get(nameOrSlugOrId, $select: 'id').get('id')
 		ResourceNotFoundError: errors.BalenaApplicationNotFound
 	}
 
@@ -60,14 +60,14 @@ getApplicationModel = (deps, opts) ->
 		resourceName: 'application_config_variable'
 		resourceKeyField: 'name'
 		parentResourceName: 'application',
-		getResourceId: (nameOrId) -> exports.get(nameOrId, $select: 'id').get('id')
+		getResourceId: (nameOrSlugOrId) -> exports.get(nameOrSlugOrId, $select: 'id').get('id')
 		ResourceNotFoundError: errors.BalenaApplicationNotFound
 	}
 	envVarModel = buildDependentResource { pine }, {
 		resourceName: 'application_environment_variable'
 		resourceKeyField: 'name'
 		parentResourceName: 'application',
-		getResourceId: (nameOrId) -> exports.get(nameOrId, $select: 'id').get('id')
+		getResourceId: (nameOrSlugOrId) -> exports.get(nameOrSlugOrId, $select: 'id').get('id')
 		ResourceNotFoundError: errors.BalenaApplicationNotFound
 	}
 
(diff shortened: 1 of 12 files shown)


lib/models/application.coffee
```diff
@@ -202,35 +202,37 @@ getApplicationModel = (deps, opts) ->
...
 			else
 				pine.get
 					resource: 'application'
 					options:
 						mergePineOptions
 							$filter:
-								app_name: nameOrId
+								$or:
+									app_name: nameOrSlugOrId
+									slug: nameOrSlugOrId
```
dfunckt: I think it's worth converting to lowercase:

```suggestion
									slug: nameOrSlugOrId.toLowerCase()
```

pdcastro: If it was true (is it?!) that a slug always includes a slash character (`'/'`) and an app name never does, then the SDK could decide which query to make instead of asking the database to search for both (i.e., avoiding the need for `$or`).

thgreasi: I'm not sure about that :thinking: 
Imo it feels weird thinking that a user has an app named `MyApp` and doing `username/MyaPP` still gives back results.
Moreover, we already do not do such a conversion for the app name, even though that's also unique per user on its lower case form. So doing it for the slug feels a bit inconsistent.
See: https://github.com/balena-io/balena-api/blob/51f6c5e67553d6fd57824fd720f2d96ac3c609ad/src/models/hooks/resources/application.ts#L196-L218

I think we should leave the casing changing for SDK consumers, eg the CLI.
An SDK consumer might be indeed be interested on getting an app by exact slug casing.
Another, not that related though is that the git url (which is tightly coupled with the slug) is case sensitive, so doing the same here kind of makes sense to me.

thgreasi: @pdcastro the slug always has a `/` on balenaCloud, but on open balena the slug is just the `app_name.toLowerCase()` (since there are no owners). We could still have a check to not filter on the `app_name` if there is a `/`, but that still doesn't allow use to be able to filter only on a single field for all cases and environments.

If we expect that's going to have measurable impact on the DBs load, then it makes sense to add it and avoid the $or iff the arg contains a `/`. On the other hand this will add a differentiation point based on a check that only exists on a hook and isn't part of the model.

I'm partially convinced of adding that though. What do you think?

pdcastro: `application.get` is a very common operation, but I don't know if the impact on DB/API load would be measurable. The fact that openBalena slugs don't include a `'/'` spoils my suggestion, and makes the extra logic harder to justify (thinking of Petro's summit presentation on avoiding `'if's`). It's your call -- I'm happy for the PR to be merged as it is. :-)

dfunckt: I still think converting to lowercase is best. This is because the API implicitly manages the slug for you behind the scenes. The SDK converting to lower case is the flip side of that coin. Ideally the model itself would automatically do case-insensitive comparisons against that field, but having the SDK normalise the input value is the next best thing.

thgreasi: @dfunckt @sradevski It still feels weird to me to be able to find an app by using `THGREASI/TESTSTARTER`, and prefer the strict-over-loose when in doubt, but I see your point.
Pushed the suggested change as a separate commit.
Let's see what @shaunmulligan thinks about this whole PR.

thgreasi:
Product team said, let's start with case sensitive and we can make it case insensitive later if needed.
See: https://www.flowdock.com/app/rulemotion/resin-frontend/messages/2266834

Issue #582, opened 2018-09-28
Should be able to run the tests on balenaCI


thgreasi:
Arch call discussions:
See: https://app.frontapp.com/open/cnv_1ha7dkz use sandboxed accounts for tests
See: https://github.com/balena-io/balena-api/issues/2558
See: https://app.frontapp.com/open/cnv_13stakz signup for a new user on each run of the suite & bypass ratelimiting on balenaCI

myarmolinsky:
Closing as we no longer have balenaCI

PR #809, opened 2020-02-05
Disable balenaCI auto rebase

thgreasi:
Having autorebase enabled could lead to having
more then one instances of the sdk tests running in
parallel, which will result all of them failing, b/c they
will use the same test user, and will also add
unnecessary load to the API. Once we get #582 fixed
we can remove this.

Change-type: patch
See: https://github.com/balena-io/scripts/pull/80

---
##### Contributor checklist

- [ ] Includes tests
- [ ] Includes typings

### DIFF
--- a/repo.yml
+++ b/repo.yml
@@ -1,0 +1,2 @@
+type: 'node'
+autoRebase: false


thgreasi:
@balena-ci retest

thgreasi:
@balena-ci retest

PR #812, opened 2020-02-06
Switch to resin-lint for linting 

Page-:
Change-type: patch

Resolves: # 
HQ: <url> 
See: <url> 
Depends-on: <url> 

---
##### Contributor checklist

- [ ] Includes tests
- [ ] Includes typings


tslint.json
```diff
@@ -1,55 +0,0 @@
-{
```
thgreasi: How about having a tslint like
```
{
    "extends": "./node_modules/resin-lint/config/tslint-prettier.json"
}
```
So that IDE's can easily find the configuration?

Page-: Done

PR #816, opened 2020-02-11
Added `getDashboardUrl(id)` to application model

codewithcheese:
Requirement to make dashboard url visible for cli app commands.

See: https://github.com/balena-io/balena-cli/issues/1440
Change-type: minor
Signed-off-by: Thomas Manning <thomasm@balena.io>

---
##### Contributor checklist

- [x] Includes tests
- [ ] Includes typings

### DIFF
--- a/lib/models/application.coffee
+++ b/lib/models/application.coffee
@@ -14,6 +14,7 @@ See the License for the specific language governing permissions and
 limitations under the License.
 ###
 
+url = require('url')
 Promise = require('bluebird')
 once = require('lodash/once')
 assign = require('lodash/assign')
@@ -40,7 +41,7 @@ errors = require('balena-errors')
 
 getApplicationModel = (deps, opts) ->
 	{ request, pine } = deps
-	{ apiUrl } = opts
+	{ apiUrl, dashboardUrl } = opts
 
 	deviceModel = once -> require('./device')(deps, opts)
 	releaseModel = once -> require('./release')(deps, opts)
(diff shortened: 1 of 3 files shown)


thgreasi:
@sradevski I also thought about this, but thought
* it would be convenient for the app list in the CLI to use a sync call that doesn't do an extra request to the API for each item
* it would be consistent with the current implementation of the `device.getDashboardUrl()`

Let me actually ask @codewithcheese to extend the example to also include how to fetch the application id.

lib/models/application.coffee
```diff
@@ -89,6 +94,25 @@ getApplicationModel = (deps, opts) ->
...
+	# @function getDashboardUrl
+	# @memberof balena.models.application
+	#
+	# @param {Number} id - Application id
+	#
+	# @returns {String} - Dashboard URL for the specific application
+	# @throws Exception if the id is not a finite number
+	#
+	# @example
+	# dashboardApplicationUrl = balena.models.application.getDashboardUrl(1)
```
thgreasi: @codewithcheese lets change the example to something like:
```suggestion
	# balena.models.application.get('MyApp').then(function(application) {
	# 	const dashboardApplicationUrl = balena.models.application.getDashboardUrl(application.id);
	# 	console.log(dashboardApplicationUrl);
	# });
```

codewithcheese:
Accepted @thgreasi example to fetch application id using application name.

@sradevski OK for this function to accept only application `id` for the reasons @thgreasi listed? I see I should have noted my similar reasoning in the PR description.

Check's seem to be broken in various ways not related to this PR any tips on how to fix?

PR #818, opened 2020-02-11
appveyor: Set to test against node 8

thgreasi:
The latest resin-lint doesn't work on node 6, which is
EOL. In the next major release we will also change
the engines field of the package.json and the node
typings.

Change-type: patch

Resolves: # 
HQ: <url> 
See: <url> 
Depends-on: <url> 

---
##### Contributor checklist

- [ ] Includes tests
- [ ] Includes typings


thgreasi:
Merging in order to unblock #819, since the tests pass individually, but fail on appveyor b/c of needed more than 1 hour after the recent API deployment.
#819 will run the node & browser tests in parallel so that the tests finish in less than 1 hour.

PR #822, opened 2020-02-17
Add device status enum and typings

sradevski:
Adds the typings for the overall_status calculated term
and exports an enum with the possible status values

Anyone using https://github.com/balena-io-modules/balena-device-status should start using `overall_status` and the enum exported by the SDK instead.

Change-type: minor

---
##### Contributor checklist

- [ ] Includes tests
- [ ] Includes typings

### DIFF
--- a/lib/models/device.coffee
+++ b/lib/models/device.coffee
@@ -55,6 +55,9 @@ deviceStatus = require('balena-device-status')
 	LOCAL_MODE_ENV_VAR
 	LOCAL_MODE_SUPPORT_PROPERTIES
 } = require('../util/local-mode')
+{
+	OverallStatus
+} = require('./device-ts')
 
 # The min version where /apps API endpoints are implemented is 1.8.0 but we'll
 # be accepting >= 1.8.0-alpha.0 instead. This is a workaround for a published 1.8.0-p1
@@ -106,20 +109,8 @@ getDeviceModel = (deps, opts) ->
 		ResourceNotFoundError: errors.BalenaDeviceNotFound
 	}
 
-	overallStatus = {
-		CONFIGURING: 'configuring',
-		IDLE: 'idle',
-		OFFLINE: 'offline',
-		INACTIVE: 'inactive',
-		POST_PROVISIONING: 'post-provisioning',
-		UPDATING: 'updating',
-		ORDERED: 'ordered',
-		PREPARING: 'preparing',
-		SHIPPED: 'shipped',
-	}
-
 	exports = {
-		overallStatus
+		OverallStatus
 	}
 
 	# Infer dashboardUrl from apiUrl if former is undefined
--- a/lib/models/device.coffee
+++ b/lib/models/device.coffee
@@ -106,7 +106,21 @@ getDeviceModel = (deps, opts) ->
 		ResourceNotFoundError: errors.BalenaDeviceNotFound
 	}
 
-	exports = {}
+	overallStatus = {
+		CONFIGURING: 'configuring',
+		IDLE: 'idle',
+		OFFLINE: 'offline',
+		INACTIVE: 'inactive',
+		POST_PROVISIONING: 'post-provisioning',
+		UPDATING: 'updating',
+		ORDERED: 'ordered',
+		PREPARING: 'preparing',
+		SHIPPED: 'shipped',
+	}
+
+	exports = {
+		overallStatus
+	}
 
 	# Infer dashboardUrl from apiUrl if former is undefined
 	if not dashboardUrl?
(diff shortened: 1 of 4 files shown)


DOCUMENTATION.md
```diff
@@ -11,7 +11,7 @@ If you feel something is missing, not clear or could be improved, please don't h
 **Kind**: global namespace  
 
 * [balena](#balena) : <code>object</code>
-    * [.interceptors](#balena.interceptors) : <code>Array.&lt;Interceptor&gt;</code>
+    * [.interceptors](#balena.interceptors) : <code>[ &#x27;Array&#x27; ].&lt;Interceptor&gt;</code>
```
sradevski: I am not sure why this changed, maybe some of the tooling is environment-specific?

typings/balena-sdk.d.ts
```diff
@@ -426,6 +426,20 @@ declare namespace BalenaSdk {
 		state: 'pending' | 'paid' | 'failed' | 'past_due';
 	}
 
+	export interface DeviceOverallStatusMap {
```
Page-: Any reason this isn't an enum?

DOCUMENTATION.md
Page-: I think it's due to an updated dependency

lib/models/device.coffee
```diff
@@ -106,7 +106,21 @@ getDeviceModel = (deps, opts) ->
 		ResourceNotFoundError: errors.BalenaDeviceNotFound
 	}
 
-	exports = {}
+	overallStatus = {
```
thgreasi: How about moving this in a TS file and use an `enum` (not a `const enum`) and then import it both in here and in the `typings/balena-sdk.d.ts`?

DOCUMENTATION.md
thgreasi: That's already updated in master, so we can probably drop this after rebasing.

typings/balena-sdk.d.ts
thgreasi: Jinx @Page- :)

lib/models/device.coffee
sradevski: @thgreasi as far as I am aware, we cannot get something like `valueof` on an enum, so we will have duplication in the `overall_status` typings. I can still go for `enum` though, but we will have a duplication.

Page-: Why not a `const enum` @thgreasi? And @sradevski the "valueof" type of an enum is just the enum itself

sradevski: @Page- I meant the union of all possible values for the enum, what `overalProgress[keyof typeof overallProgress]` achieves.

sradevski: Nevermind, I was looking at the code and I see what you mean, that would definitely make more sense, I was just thinking in `balena-request-status` terms.

typings/device-overall-status.d.ts
```diff
@@ -0,0 +1,19 @@
+/* tslint:disable:no-namespace */
+declare namespace DeviceOverallStatus {
```
thgreasi: Do we really need to use a namespace? We also do use plain type exports in other typings that get re-exported by the main balena-sdk typings.

typings/balena-sdk.d.ts
```diff
@@ -463,6 +466,7 @@ declare namespace BalenaSdk {
 		vpn_address: string | null;
 		should_be_managed_by__supervisor_release: number;
 		api_heartbeat_state: 'online' | 'offline' | 'timeout' | 'unknown';
+		overall_status: DeviceOverallStatus.OverallStatus;
```
thgreasi: I would love to have a `/** */` comment that this is computed, or maybe even try declaring it as `readonly`!

sradevski: I don't think the `readonly` is necessary, otherwise we need to set it to all of them (since ideally we never modify the object)

thgreasi: It isn't necessary indeed, I was just thinking of a way to represent that this is a computed term in a way the TS can understand it (and in the future error if you try to patch it).

typings/device-overall-status.d.ts
```diff
@@ -0,0 +1,14 @@
+/**
+ * Extracted from the enum defined in `device-ts.ts`.
+ */
+export enum DeviceOverallStatus {
```
Page-: Why is there a separate .d.ts file for this? The type is already available from the .ts file so will have an auto generated .d.ts for it

sradevski: @Page- we chatted with Thodoris about it, and you cannot import a `.ts` file in `d.ts` since the `.ts` file gets compiled away and the import will fail.

Page-: Ok, I see why you need it now, I'd say it makes sense to rename device-ts.ts to device-overall-status.ts then to make it a little clearer that this is just the .d.ts for it (which we need to commit for the sake of the device.d.ts

PR #827, opened 2020-02-19
Convert service model to typescript

sradevski:
Change-type: patch

---
##### Contributor checklist

- [ ] Includes tests
- [ ] Includes typings


lib/models/service.ts
```diff
@@ -0,0 +1,291 @@
...
+			getResourceId: (id: number) => get(id, { $select: 'id' }).get('id'),
+			ResourceNotFoundError: errors.BalenaServiceNotFound,
+		},
+	);
+
+	exports = {};
+
+	// Not exported for now, but we could document & export in future
+	// if there are external use cases for this.
+	const get = (id: number, options: PineOptionsFor<Service>) => {
```
thgreasi: How about:
```suggestion
	const get = (id: number, options: PineOptionsFor<Service> = {}) => {
```
Applies in the rest of the file as well.

lib/models/service.ts
```diff
@@ -0,0 +1,291 @@
...
+	 * @memberof balena.models.service
+	 *
+	 * @param {String|Number} nameOrSlugOrId - application name (string), slug (string) or id (number)
+	 * @param {Object} [options={}] - extra pine options to use
+	 * @fulfil {Object[]} - services
+	 * @returns {Promise}
+	 *
+	 * @example
+	 * balena.models.service.getAllByApplication('MyApp').then(function(services) {
+	 * console.log(services);
```
thgreasi: It seems that the indentation broke here, making the README examples to have zero indentation.
Let's re-indent them plz.
Applies to the two other examples right below.

lib/models/service.ts
```diff
@@ -0,0 +1,291 @@
...
+	 * balena.models.service.getAllByApplication('MyApp', function(error, services) {
+	 * if (error) throw error;
+	 * console.log(services);
+	 * });
+	 */
+	exports.getAllByApplication = function(
+		nameOrSlugOrId: string | number,
+		options: PineOptionsFor<Service>,
+		callback?: (error?: Error, result?: Service) => void,
+	): Promise<Service> {
```
thgreasi:
```suggestion
	): Promise<Service[]> {
```

lib/models/service.ts
```diff
@@ -0,0 +1,291 @@
...
+	 * @example
+	 * balena.models.service.getAllByApplication('MyApp', function(error, services) {
+	 * if (error) throw error;
+	 * console.log(services);
+	 * });
+	 */
+	exports.getAllByApplication = function(
+		nameOrSlugOrId: string | number,
+		options: PineOptionsFor<Service>,
+		callback?: (error?: Error, result?: Service) => void,
```
thgreasi:
```suggestion
		callback?: (error?: Error, result?: Service[]) => void,
```

lib/models/service.ts
```diff
@@ -0,0 +1,291 @@
...
+		 * @example
+		 * balena.models.service.var.getAllByApplication('MyApp', function(error, vars) {
+		 * 	if (error) throw error;
+		 * 	console.log(vars)
+		 * });
+		 */
+		getAllByApplication(
+			nameOrSlugOrId: string | number,
+			options: PineOptionsFor<ServiceEnvironmentVariable>,
+			callback?: (error?: Error, result?: ServiceEnvironmentVariable) => void,
```
thgreasi:
```suggestion
			callback?: (error?: Error, result?: ServiceEnvironmentVariable[]) => void,
```

thgreasi:
Also let's rename the commit to something like
> Convert the service model to typescript

since the CHANGELOG is the way that we communicate changes to our users :)

lib/models/service.ts
sradevski: I just used decaffeinate and didn't really think of doing this

lib/models/service.ts
sradevski: the linter was complaining about the original formatting, but I managed to get it fixed now.

PR #831, opened 2020-02-26
Bump TypeScript version to ^3.8.2, so that's used in tests & we can benefit from its better error detection

thgreasi:
Change-type: patch

---
##### Contributor checklist

- [ ] Includes tests
- [ ] Includes typings

### DIFF
--- a/typing_tests/pine-options.ts
+++ b/typing_tests/pine-options.ts
@@ -88,7 +88,7 @@ export const unkown$selectPropInsideNested$expandWith$select2Fixed: BalenaSdk.Pi
 };
 
 export const unkown$expandPropInArray$expands: BalenaSdk.PineOptionsFor<BalenaSdk.Application> = {
-	$expand: [ // $ExpectError
+	$expand: [
 		{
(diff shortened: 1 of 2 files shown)


typing_tests/pine-options.ts
```diff
@@ -342,9 +334,7 @@ export const unknown$any$filterPropInsideArray$expandPlural: BalenaSdk.PineOptio
 							$expr: {
 								dev: {
 									name: 'name',
-									// this should be failing, but it isn't :(
-									// See: https://github.com/microsoft/TypeScript/issues/32000
-									asdf: 'asdf', // $ExpectError-Skip
```
thgreasi: At last this is detected :tada:

thgreasi:
@sradevski the issue was the the new TS version is emitting the errors nearer the exact line with the issue, while before in some cases the error was emitted on the root level of each object.
Moreover, it now properly detects invalid props in union types w/ dictionaries :tada: 
Will bump the ts version in pakcage.json :+1:

PR #835, opened 2020-02-28
Convert to ts os

sradevski:
Convert OS model to typescript

Change-type: patch

---
##### Contributor checklist

- [ ] Includes tests
- [ ] Includes typings


lib/models/os.ts
```diff
@@ -0,0 +1,674 @@
...
+	/**
+	 * @summary Get device types with caching
+	 * @description Utility method exported for testability.
+	 * @name _getDeviceTypes
+	 * @private
+	 * @function
+	 * @memberof balena.models.os
+	 */
+	const _getDeviceTypes = withDeviceTypesEndpointCaching<
+		() => Promise<DeviceType[]>
```
thgreasi: Do we have to explicitly specify this one?

lib/models/os.ts
```diff
@@ -0,0 +1,674 @@
...
+	 * @private
+	 * @function
+	 * @memberof balena.models.os
+	 */
+	const _getOsVersions = withDeviceTypesEndpointCaching((deviceType: string) =>
+		request
+			.send({
+				method: 'GET',
+				url: `/device-types/v1/${deviceType}/images`,
+				baseUrl: apiUrl,
```
thgreasi: Let's add back the comment that explains why this is not omitting the token, so that we don't forget about it.
> optionally authenticated, so we send the token in all cases

lib/models/os.ts
```diff
@@ -0,0 +1,674 @@
...
+	 * @description Utility method exported for testability.
+	 * @name _clearDeviceTypesEndpointCaches
+	 * @private
+	 * @function
+	 * @memberof balena.models.os
+	 */
+	const _clearDeviceTypesEndpointCaches = () => {
+		_getDeviceTypes.clear();
+		_getDownloadSize.clear();
+		return _getOsVersions.clear();
```
thgreasi:
```suggestion
		_getOsVersions.clear();
```

lib/models/os.ts
```diff
@@ -0,0 +1,674 @@
...
+	/**
+	 * @summary Get the max OS version satisfying the given range.
+	 * @description Utility method exported for testability.
+	 * @name _getMaxSatisfyingVersion
+	 * @private
+	 * @function
+	 * @memberof balena.models.os
+	 */
+	const _getMaxSatisfyingVersion = function(
+		versionOrRange: keyof OsVersions,
```
thgreasi: I think that this should be just `string` since it can hold a semver range as well.
Doing `| 'default'` would add some clarity for devs, but the resulting typings will just be widen to `string`.

lib/models/os.ts
```diff
@@ -0,0 +1,674 @@
...
+	): Promise<Date> {
+		callback = findCallback(arguments);
+
+		return getValidatedDeviceType(deviceType)
+			.then(() => normalizeVersion(version))
+			.then(ver =>
+				request.send({
+					method: 'HEAD',
+					url: deviceImageUrl(deviceType, ver),
+					baseUrl: apiUrl,
```
thgreasi: Let's add back the comment that explains why this is not omitting the token, so that we don't forget about it.
> optionally authenticated, so we send the token in all cases

lib/models/os.ts
```diff
@@ -0,0 +1,674 @@
...
+					return _getOsVersions(deviceType).get('latest');
+				}
+
+				return normalizeVersion(version);
+			})
+			.then(ver =>
+				request.stream({
+					method: 'GET',
+					url: deviceImageUrl(deviceType, ver),
+					baseUrl: apiUrl,
```
thgreasi: Let's add back the comment that explains why this is not omitting the token, so that we don't forget about it.
> optionally authenticated, so we send the token in all cases

typings/balena-sdk.d.ts
```diff
@@ -584,7 +584,7 @@ declare namespace BalenaSdk {
 		ip?: string;
 		gateway?: string;
 		netmask?: string;
-		version: string;
+		version?: string;
```
thgreasi: I think that this always holds a `version`.
See: https://github.com/balena-io/open-balena-api/blob/test-suite-improvements/src/lib/device-config.ts#L93

lib/models/os.ts
```diff
@@ -0,0 +1,674 @@
...
+					allVersions,
+					v => bSemver.compare(v, currentVersion) === 0,
+				);
+
+				const versions = allVersions.filter((
+					v, // avoid the extra call to getValidatedDeviceType,
+				) =>
+					// since getSupportedVersions is already does that
+					hupActionHelper.isSupportedOsUpdate(deviceType, currentVersion, v),
+				);
```
thgreasi: It looks like the `,` in the comment, confused decaffeinate
```suggestion
				const versions = allVersions.filter(v =>
					// avoid the extra call to getValidatedDeviceType,
					// since getSupportedVersions is already does that
					hupActionHelper.isSupportedOsUpdate(deviceType, currentVersion, v),
				);
```

lib/models/os.ts
sradevski: Yes, because the config model is in coffeescript, so getDeviceTypes returns `any`.

lib/models/os.ts
sradevski: Hmm decaffeinate seems to have removed it, nice catch.

lib/models/os.ts
```diff
@@ -0,0 +1,676 @@
...
+			.then(({ versions: allVersions }) => {
+				// use bSemver.compare to find the current version in the OS list
+				// to benefit from the baked-in normalization
+				const current = find(
+					allVersions,
+					v => bSemver.compare(v, currentVersion) === 0,
+				);
+
+				const versions = allVersions.filter(v =>
+					// avoid the extra call to getValidatedDeviceType, since getSupportedVersions is already does that
```
thgreasi: It seems that I had a typo here :sweat_smile: 
```suggestion
					// avoid the extra call to getValidatedDeviceType, since getSupportedVersions already does that
```

PR #837, opened 2020-02-28
Lazy load sdk and models props

Page-:
Resolves: # 
HQ: <url> 
See: <url> 
Depends-on: <url> 

---
##### Contributor checklist

- [ ] Includes tests
- [ ] Includes typings


lib/models/index.ts
```diff
@@ -0,0 +1,116 @@
...
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+const modelsTemplate = [
+	/**
```
Page-: I converted to typescript as I could not put jsdoc comments within an array in coffeescript

lib/balena.coffee
```diff
@@ -81,6 +75,7 @@ getSdk = (opts = {}) ->
 		apiVersion: 'v5'
 
 	if opts.isBrowser
+		{ notImplemented } = require('./util')
```
Page-: This is beneficial because with the new lazy loading it means there's a lot we can actually do without ever triggering utils to load (and it's fairly costly because of memoizee and cloneDeep requires)

lib/balena.coffee
```diff
@@ -104,8 +99,22 @@ getSdk = (opts = {}) ->
...
+		'auth',
+		'models'
+		'logs'
+		'settings'
+	].forEach (moduleName) ->
+		Object.defineProperty(sdk, moduleName, {
+			enumerable: true,
+			configurable: true,
+			get: ->
+				moduleFactory = require("./#{moduleName}")
```
thgreasi: It seems that browserify doesn't really like this way of imports.
How about doing
```
sdkTemplate =
	auth: -> require('./auth')
	models: -> require('./models')
	logs: -> require('./logs')
	settings: -> require('./settings'
```
and then in here just call
```
moduleFactory = sdkTemplate["./#{moduleName}"]()
```
?

Page-:
Tested in the CLI and the CLI worked without changes

lib/balena.coffee
Page-: Ok, I was going to ask about browserify, I'll switch it over to that style - I was already considering using that style to remove the distinction between `.default` and non `.default` handling and if it keeps browserify happy then it's a no-brainer

lib/models/index.ts
```diff
@@ -0,0 +1,116 @@
...
+			configurable: true,
+			get() {
+				const moduleFactory = require(`./${moduleName}`).default;
+				// We need the delete first as the current property is read-only
+				// and the delete removes that restriction
+				delete this[moduleName];
+				return (this[moduleName] = moduleFactory(deps, opts));
+			},
+		});
+	});
```
thgreasi: How about doing something like:
```
modelsTemplate =

	/**
	 * @namespace application
	 * @memberof balena.models
	 */
	application: () => require('./application')

	/**
	 * @namespace image
	 * @memberof balena.models
	 */
	image: () => require('./image').default
```
This way we don't need the second `tsModelsTemplate` code section and can even preserve the order (to avoid the unnecessary changes in the docs). Most importantly, it should work on browserify as well :)

Page-:
All still looks good in the CLI with the changes

lib/models/index.ts
Page-: With the new object form I could technically keep it in coffeescript but I don't think it's worth reverting :P  (although we don't actually get typing benefits atm, it does at least make it easier to add proper typings in future)

PR #839, opened 2020-03-03
Update dependencies

Page-:
Change-type: patch

Resolves: # 
HQ: <url> 
See: <url> 
Depends-on: <url> 

---
##### Contributor checklist

- [ ] Includes tests
- [ ] Includes typings


package.json
```diff
@@ -93,27 +91,25 @@
...
     "balena-hup-action-utils": "~3.0.0",
     "balena-pine": "^10.1.1",
     "balena-register-device": "^6.0.1",
-    "balena-request": "^10.0.4",
+    "balena-request": "^10.0.8",
     "balena-semver": "^2.1.0",
-    "balena-settings-client": "^4.0.2",
+    "balena-settings-client": "^4.0.4",
     "bluebird": "^3.7.2",
-    "common-tags": "^1.8.0",
```
Page-: This was unused so I removed it

package.json
```diff
@@ -69,19 +69,17 @@
...
     "js-polyfills": "^0.1.42",
-    "jsdoc-to-markdown": "^3.0.4",
+    "jsdoc-to-markdown": "^5.0.3",
     "karma": "^3.1.4",
     "karma-env-preprocessor": "^0.1.1",
-    "lint-staged": "^10.0.1",
+    "lint-staged": "^10.0.8",
     "mocha": "^3.5.3",
     "mochainon": "^2.0.0",
-    "prettier": "^1.19.1",
```
Page-: This is used via resin-lint now so I removed it here

package.json
```diff
@@ -93,27 +91,25 @@
...
-    "balena-settings-client": "^4.0.2",
+    "balena-settings-client": "^4.0.4",
     "bluebird": "^3.7.2",
-    "common-tags": "^1.8.0",
     "lodash": "^4.17.15",
     "memoizee": "^0.4.14",
     "moment": "^2.24.0",
     "ndjson": "^1.5.0",
-    "semver": "^5.7.1"
+    "semver": "^7.1.3"
```
Page-: Fixes a bug with intersects (which we don't use anyway) and drops old node versions we don't support: https://github.com/npm/node-semver/blob/master/CHANGELOG.md

PR #840, opened 2020-03-05
balena-action-hup-utils: bump to 4.0.0 to allow dev upgrades

xginn8:
HQ: https://github.com/balena-io/balena/issues/1815
HQ: https://github.com/balena-io/balena/issues/1322

Change-type: minor

### DIFF
--- a/tests/integration/models/device.spec.coffee
+++ b/tests/integration/models/device.spec.coffee
@@ -2087,25 +2087,43 @@ describe 'Device Model', ->
 						target_os_version
 					)).to.throw('The device is offline')
 
-			it 'should throw when the device is running a dev version', ->
+			it 'should throw for upgrades from prod -> dev', ->
+				[
+					['Resin OS 2.0.0+rev3 (prod)', 'prod']
+					['Resin OS 2.0.0+rev3 (prod)', '']
+					['Resin OS 2.0.4+rev1', 'prod']
+					['Resin OS 2.0.5', 'prod']
+					['Resin OS 2.12.1+rev1', 'prod']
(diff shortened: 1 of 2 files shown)


tests/integration/models/device.spec.coffee
```diff
@@ -2087,7 +2087,43 @@ describe 'Device Model', ->
...
+					['Resin OS 2.0.0+rev3 (prod)', 'prod']
+					['Resin OS 2.0.0+rev3 (prod)', '']
+					['Resin OS 2.0.0+rev5 (prod)', '']
+					['Resin OS 2.0.2+rev2', 'prod']
+					['Resin OS 2.0.3+rev1', 'prod']
+					['Resin OS 2.0.4+rev1', 'prod']
+					['Resin OS 2.0.5', 'prod']
+					['Resin OS 2.0.5+rev1', 'prod']
+					['Resin OS 2.0.6+rev1', 'prod']
+					['Resin OS 2.0.6+rev2', 'prod']
```
sradevski: We can drop some of these as there is a lot of redundancy.

xginn8: Removed some similar cases.

xginn8:
@thgreasi looks like we're hitting the same error as in https://github.com/balena-io-modules/balena-hup-action-utils/pull/19:

```
Error: could not get uid/gid

[ 'nobody', 0 ]
```

thgreasi:
@xginn8 we don't run the sdk tests on balenaCI yet :(
So you should only need appveyor to pass, which it seems that had errors while running the node tests but worked for the browser ones. So, I just restarted them since it seems like an API hickup.

PR #846, opened 2020-03-09
Change the error filters to functions

thgreasi:
This was needed so that the Bluebird typings stop
complaining.

Change-type: patch

Resolves: # 
HQ: <url> 
See: <url> 
Depends-on: <url> 

---
##### Contributor checklist

- [ ] Includes tests
- [ ] Includes typings

### DIFF
--- a/lib/util/index.ts
+++ b/lib/util/index.ts
@@ -66,27 +66,25 @@ export const findCallback = (args: IArguments) => {
 	return;
 };
 
-export const unauthorizedError: ErrorResponse = {
-	code: 'BalenaRequestError',
-	statusCode: 401,
-};
+const isBalenaRequestErrorResponseWithCode = (
+	error: Partial<errors.BalenaRequestError>,
+	statusCode: number,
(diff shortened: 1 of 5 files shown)


lib/util/index.ts
```diff
@@ -66,27 +66,25 @@ export const findCallback = (args: IArguments) => {
...
-	statusCode: 500,
-	body: 'No application found to associate with the api key',
-};
+export const isNoDeviceForKeyResponse = (err: Error) =>
+	isBalenaRequestErrorResponseWithCode(err, 500) &&
+	err.body === 'No device found to associate with the api key';
+
+export const isNoApplicationForKeyResponse = (err: Error) =>
+	isBalenaRequestErrorResponseWithCode(err, 500) &&
+	err.body === 'No application found to associate with the api key';
```
thgreasi: The initial PR actually had a bug where I was using 404 instead of 500 for these :disappointed: 
Thankfully the tests caught it.

PR #848, opened 2020-03-10
Add tests for arbitary pine requests

thgreasi:
This makes sure that the typings that we have are indeed working.
Once the orgs PR finalizes we could move the respective requests under their tests cases and clean them up, but for now it's on a separate file to ease rebasing.

The commented lines are alt options for the same expands that do work.
I chose the non-commented ones based on what made more sense based on the SBVR and when in doubt I tried to use the same terms as the ones that we use in the API.

Change-type: patch

---
##### Contributor checklist

- [x] Includes tests
- [ ] Includes typings

### DIFF
--- a/tests/integration/arbitrary-pine-requests.spec.ts
+++ b/tests/integration/arbitrary-pine-requests.spec.ts
@@ -1,0 +1,114 @@
+import * as m from 'mochainon';
+import { describeExpandAssertions } from '../util_ts';
+import {
+	balena,
+	BalenaSdk,
+	givenAnApplication,
+	givenLoggedInUser,
+} from './setup';
+const { expect } = m.chai;
+
+describe('arbitrary pine requests', function() {
+	givenLoggedInUser(before);
+	givenAnApplication(before);
+
+	describeExpandAssertions<BalenaSdk.Application>({
+		resource: 'application',
+		options: {
+			$expand: {
+				organization: {},
+				// includes__user: {},
+				user__is_member_of__application: {},
+				// is_accessible_by__team: {},
+				team_application_access: {},
+			},
+		},
+	});
+
+	describeExpandAssertions<BalenaSdk.ApplicationMembership>({
+		resource: 'user__is_member_of__application',
+		options: {
+			$expand: {
+				user: {},
+				// application: {},
+				is_member_of__application: {},
+				application_membership_role: {},
+			},
+		},
+	});
+
+	describeExpandAssertions<BalenaSdk.User>({
+		resource: 'user',
+		options: {
+			$expand: {
+				// includes__organization_membership: {},
+				organization_membership: {},
+				// is_member_of__application: {},
+				// user_application_membership: {},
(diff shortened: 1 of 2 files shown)


dfunckt:
@thgreasi which OData versions did you try the alternative relations with and found out they work? I think for some of these this is only due to translations and they'll stop working when the next version is released.

thgreasi:
@dfunckt back when I wrote the typings I had tested it against `v6` as well, but today while writing these tests, I only used `/resin`. I will checkout the `v6` branch again to give them a try.

dfunckt:
Hmm, I wouldn't expect `/resin` to work for eg. `user__is_member_of__application.application` (it should be `user__is_member_of__application.is_member_of__application`) but this being an `$expand` I could be wrong and mixing up things with `$filter`.

thgreasi:
The tests seem to pass on your old v6 API branch @dfunckt (even when uncommenting all other props), so I guess we are fine.

tests/integration/arbitrary-pine-requests.spec.ts
```diff
@@ -0,0 +1,114 @@
...
+import { describeExpandAssertions } from '../util_ts';
+import {
+	balena,
+	BalenaSdk,
+	givenAnApplication,
+	givenLoggedInUser,
+} from './setup';
+const { expect } = m.chai;
+
+describe('arbitrary pine requests on organizations model', function() {
```
sradevski: I really don't like the description, can't we find several representative cases and test those instead? Or we can rename it to "the organization model object shape" or something along those lines.

thgreasi: @sradevski We actually already have `should support arbitrary pinejs options` in various model tests and that's why I followed a similar naming scheme here.
For first, let me drop the `on organizations model` suffix, since that's too specific.
Secondly, `describeExpandAssertions` is actually creating more sensible test cases and the result that it prints looks something like:
```
  arbitrary pine requests
    expanding from application
      to organization
        ✓ should succeed (535ms)
        ✓ should include the expanded property
      to user__is_member_of__application
        ✓ should succeed (536ms)
        ✓ should include the expanded property
      to team_application_access
        ✓ should succeed (685ms)
        ✓ should include the expanded property
    expanding from user__is_member_of__application
      to user
        ✓ should succeed (523ms)
        - should include the expanded property
      to is_member_of__application
        ✓ should succeed (518ms)
        - should include the expanded property
      to application_membership_role
        ✓ should succeed (527ms)
        - should include the expanded property
```

The whole point of this PR was to make sure that our typings are still valid after the (hopefully last) upcoming model changes.
If you feel that it's better to not merge it, I can definitely just keep rebasing it for testing purposes and just close it once orgs get merged.

> can't we find several representative cases and test those instead?

Can I get an example?

sradevski: @thgreasi ah I see what's your goal here. As long as we cleanup afterwards, I am fine with it as it is.

PR #849, opened 2020-03-13
Make the unique application name test more explicit

thgreasi:
Change-type: patch

---
##### Contributor checklist

- [ ] Includes tests
- [ ] Includes typings

### DIFF
--- a/tests/integration/models/application.spec.coffee
+++ b/tests/integration/models/application.spec.coffee
@@ -149,7 +149,13 @@ describe 'Application Model', ->
 						name: 'FooBar'
 						applicationType: 'microservices-starter'
 						deviceType: 'beaglebone-black'
-					m.chai.expect(promise).to.be.rejectedWith('Application name must be unique')
+					m.chai.expect(promise).to.be.rejected
+					.then (error) ->
+						m.chai.expect(error).to.have.property('code', 'BalenaRequestError')
+						m.chai.expect(error).to.have.property('statusCode', 404)
+						m.chai.expect(error).to.have.property('message').that.matches(/\bunique\b/i)
+						# TODO: re-enable once the API regression gets fixed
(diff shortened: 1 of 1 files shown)


tests/integration/models/application.spec.coffee
```diff
@@ -142,14 +142,18 @@ describe 'Application Model', ->
 					promise = balena.models.application.hasAny()
 					m.chai.expect(promise).to.eventually.be.true
 
-			describe 'balena.models.application.create()', ->
+			describe.only 'balena.models.application.create()', ->
```
sradevski: I guess you forgot this one :)

tests/integration/models/application.spec.coffee
```diff
@@ -142,14 +142,18 @@ describe 'Application Model', ->
...
 				it 'should reject if trying to create an app with the same name', ->
 					promise = balena.models.application.create
 						name: 'FooBar'
 						applicationType: 'microservices-starter'
 						deviceType: 'beaglebone-black'
-					m.chai.expect(promise).to.be.rejectedWith('Application name must be unique')
+					m.chai.expect(promise).to.be.rejected
+					.then (error) ->
+						m.chai.expect(error).to.have.property('code', 'BalenaRequestError')
+						m.chai.expect(error).to.have.property('statusCode', 404)
```
sradevski: Why is it 404?

tests/integration/models/application.spec.coffee
thgreasi: No, I added this on purpose on a separate commit with a `Change-type: dropme`, so that Akis can give it a try, since there seems to be a regression.

tests/integration/models/application.spec.coffee
thgreasi: This goes way back on how pine was originally statusing db & rule errors and we have kept if for compatibility even in v5. We might change this to a 409 in v6 though, since pine now property uses that status code.

PR #850, opened 2020-03-23
Update to the new orgs model where the is no personal org concept

thgreasi:
Change-type: major
Depends-on: https://github.com/balena-io/balena-api/pull/2465

---
##### Contributor checklist

- [ ] Includes tests
- [ ] Includes typings

### DIFF
--- a/tests/integration/utils.ts
+++ b/tests/integration/utils.ts
@@ -1,0 +1,15 @@
+import { balena, BalenaSdk } from './setup';
+
+export const getInitialOrganization = async () => {
+	const [org] = await balena.pine.get<BalenaSdk.Organization>({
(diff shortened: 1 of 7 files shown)


tests/integration/utils.ts
```diff
@@ -0,0 +1,15 @@
+import { balena, BalenaSdk } from './setup';
+
+export const getInitialOrganization = async () => {
```
sradevski: NIT: would calling this `default` be more appropriate?

thgreasi: Didn't do so, since I expect more helpers to be added here & also didn't have any issue while consuming this in the test.

PR #853, opened 2020-03-24
device.getStatus: Suggest selecting the overall_status field directly

thgreasi:
Did not deprecate it since the plan is to change it in
the next major to just be a convenience method for a
get() with the overall_status field, so that users can still 
find it easily.

Change-type: patch
See: https://www.flowdock.com/app/rulemotion/resin-frontend/threads/KlyXZlJ86S-nvOuJbuOdpbmo54i

Change-type: patch
See: https://www.flowdock.com/app/rulemotion/resin-frontend/threads/KlyXZlJ86S-nvOuJbuOdpbmo54i

---
##### Contributor checklist

- [ ] Includes tests
- [ ] Includes typings

### DIFF
--- a/DOCUMENTATION.md
+++ b/DOCUMENTATION.md
@@ -3942,6 +3942,8 @@ balena.models.device.ping('7cf02a6', function(error) {
 
 ##### device.getStatus(device) ⇒ <code>Promise</code>
 Computes the status of an already retrieved device object.
+It's recommended to use `balena.models.device.get(deviceUuid, { $select: ['overall_status'] })` instead,
+in case that you need to retrieve more device fields than just the status.
 
 **Kind**: static method of [<code>device</code>](#balena.models.device)  
 **Summary**: Get the status of a device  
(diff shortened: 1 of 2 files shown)


DOCUMENTATION.md
```diff
@@ -3941,18 +3941,24 @@ balena.models.device.ping('7cf02a6', function(error) {
 <a name="balena.models.device.getStatus"></a>
 
 ##### device.getStatus(device) ⇒ <code>Promise</code>
+Computes the status of an already retrieved device object.
+It's suggested to use `balena.models.device.get(deviceUuid, { $select: ['overall_status'] })` instead.
```
sradevski:
```suggestion
We recommend using`balena.models.device.get(deviceUuid, { $select: ['overall_status'] })` instead, as this function will get deprecated in the future.
```

thgreasi: @sradevski We decided that we will not deprecate this.
See: https://www.flowdock.com/app/rulemotion/resin-frontend/threads/KlyXZlJ86S-nvOuJbuOdpbmo54i

We are still going to keep it around as a convenience method (which will just return that field) so that the way of retrieving the status is still discoverable by new users. It will still though have the recommendation of using the `device.get()` directly in case they need to retrieve more fields at once.

sradevski: Ah ok then 👍

PR #863, opened 2020-04-03
Switch to using native versions of lodash functions

Page-:
This leaves just `cloneDeep`, `omit`, `once`, `pick`, `throttle`, `union`, `without`


lib/balena.coffee
```diff
@@ -69,7 +67,7 @@ getSdk = (opts = {}) ->
 		logs: -> require('./logs')
 		settings: -> require('./settings')
 
-	defaults opts,
+	opts = Object.assign({
```
Page-, line 70: This is technically different in that `opts = { apiUrl: undefined }` will no longer be defaulted, `opts = {}` however would. If you feel it's necessary I'm happy to emulate the previous behavior

Page-:
None of these changes should impact the minimum supported browser versions - for most (all?) we already used a mix of lodash and native versions

lib/util/index.ts
```diff
@@ -228,15 +225,22 @@ const convertExpandToObject = <T extends {}>(
...
-	} else if (isString(expandOption)) {
+	}
+	if (typeof expandOption === 'string') {
 		return {
 			[expandOption]: {},
 		} as Pine.ResourceExpandFor<T>;
-	} else if (Array.isArray(expandOption)) {
+	}
+	// Cast to the correct type based upon the checks above as the `typeof string` check misses some string cases?
+	expandOption = expandOption as Required<Exclude<typeof expandOption, string>>;
```
Page-, line 235: This was a weird one, but it fixes the typings and lets us drop a require

PR #864, opened 2020-04-03
Convert lib/util/dependent-resource.coffee to javascript

Page-:
This allows some type-checking both of this file and of others that import it, it will also allow using modern js features that will then be transpiled by typescript the same as .ts files

Change-type: patch


lib/util/dependent-resource.js
```diff
@@ -0,0 +1,162 @@
...
+import { getUpsertHelper } from '../util/upsert';
+
+export function buildDependentResource(
+	{ pine },
+	{
+		resourceName, // e.g. device_tag
+		resourceKeyField, // e.g. tag_key
+		parentResourceName, // e.g. device
+		getResourceId, // e.g. getId(uuidOrId)
+	},
```
Page-, line 39: This used to include `ResourceNotFoundError` but typescript told me it wasn't used anywhere and made me remove it :P

lib/util/dependent-resource.js
```diff
@@ -0,0 +1,162 @@
...
+) {
+	const upsert = getUpsertHelper({ pine });
+
+	const exports = {
+		getAll(options, callback) {
```
sradevski, line 44:
```suggestion
		getAll(options = {}, callback) {
```

lib/util/dependent-resource.js
```diff
@@ -0,0 +1,162 @@
...
+					resource: resourceName,
+					options: mergePineOptions(
+						{ $orderby: `${resourceKeyField} asc` },
+						options,
+					),
+				})
+				.asCallback(callback);
+		},
+
+		getAllByParent(parentParam, options, callback) {
```
sradevski, line 61:
```suggestion
		getAllByParent(parentParam, options = {}, callback) {
```

lib/util/dependent-resource.js
```diff
@@ -0,0 +1,162 @@
...
+							resource: resourceName,
+							body: {
+								[parentResourceName]: parentId,
+								[resourceKeyField]: key,
+								value,
+							},
+						},
+						[parentResourceName, resourceKeyField],
+					).tapCatch(isUnauthorizedResponse, function() {
+						// On Pine 7, when the post throws a 401
```
sradevski, line 131: Is the comment still valid?

lib/util/dependent-resource.js
Page-: These would be breaking/major changes for calls like `.getAll(null)` which is why I didn't do them - they're probably fairly safe to bundle in the orgs bump, cc @thgreasi, but I'm playing it as safe as possible

lib/util/dependent-resource.js
Page-: I think so, if you try to link to a non-existent resource then I'd expect it to always be a 401 - you don't have access to that resource and we don't want to distinguish between it existing but you don't have access or it just not existing, as that would leak info that we don't want to leak

sradevski: @Page I would actually expect a 404, but I guess if we always return a 401 regardless if it exists or you don't have access, then it's fine.

PR #865, opened 2020-04-03
Coffeescript to js conversions

Page-:
Resolves: # 
HQ: <url> 
See: <url> 
Depends-on: <url> 

---
##### Contributor checklist

- [ ] Includes tests
- [ ] Includes typings


lib/logs.js
```diff
@@ -0,0 +1,249 @@
...
+import { EventEmitter } from 'events';
+import * as ndjson from 'ndjson';
+import { AbortController as AbortControllerPonyfill } from 'abortcontroller-polyfill/dist/cjs-ponyfill';
+import { findCallback } from './util';
+import { globalEnv } from './util/global-env';
+
+const AbortController =
+	globalEnv != null && 'AbortController' in globalEnv
+		? globalEnv.AbortController
+		: AbortControllerPonyfill;
```
Page-, line 28: This changed slightly to make typescript happy but the effect should be equivalent, previous was
```coffee
AbortController = globalEnv?.AbortController || AbortControllerPonyfill

File: lib/settings.js [hunk 2]
```diff
@@ -0,0 +1,76 @@
...
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+import * as Promise from 'bluebird';
+
+const getSettings = function(deps) {
```
Page-, line 19: Removed an unused `opts` parameter which should be safe to do

lib/logs.js
```diff
@@ -0,0 +1,249 @@
...
+				baseUrl: opts.apiUrl,
+			})
+			.get('body');
+	};
+
+	const subscribeToApiLogs = function(device, options) {
+		if (options == null) {
+			options = {};
+		}
+		/** @type {LogSubscription & { unsubscribe?: () => void } } */
```
thgreasi, line 58: I think that this broke the docs builder.

Page-: Turns out just turning it into a normal comment does the trick

Page-:
@balena-ci retest

PR #868, opened 2020-04-04
Convert application.coffee to javascript

Page-:
This required very few changes so it's easy to repeat

Depends on #865


lib/models/application.js
```diff
@@ -0,0 +1,1618 @@
...
+	const tagsModel = buildDependentResource(
+		{ pine },
+		{
+			resourceName: 'application_tag',
+			resourceKeyField: 'tag_key',
+			parentResourceName: 'application',
+			getResourceId(nameOrSlugOrId) {
+				return exports.get(nameOrSlugOrId, { $select: 'id' }).get('id');
+			},
+		},
```
Page-, line 57: I had to remove unused params from these, see https://github.com/balena-io/balena-sdk/pull/867

lib/models/application.js
```diff
@@ -0,0 +1,1618 @@
...
+					return Promise.props({
+						applicationId: getId(nameOrSlugOrId),
+						/** @type {?string} */
+						commit: null,
+					});
+				}
+
+				const app = latestRelease.belongs_to__application;
+				return {
+					applicationId: '__id' in app ? app.__id : app[0].id,
```
Page-, line 1082: I had to add this `in` check to satisfy the typings even though I know it's unnecessary, maybe you know how to avoid it @thgreasi ?

lib/models/application.js
```diff
@@ -0,0 +1,1618 @@
...
+	exports.trackLatestRelease = (nameOrSlugOrId, callback) =>
+		releaseModel()
+			.getLatestByApplication(nameOrSlugOrId, {
+				$select: ['commit', 'belongs_to__application'],
+			})
+			.then(function(latestRelease) {
+				if (!latestRelease) {
+					return Promise.props({
+						applicationId: getId(nameOrSlugOrId),
+						/** @type {?string} */
```
Page-, line 1075: I had to add this as typescript with js will just default to using the type of the first return for every return and `null` was incompatible with the `string` below

lib/models/application.js
thgreasi: @Page- casting to `BalenaSdk.PineDeferred` is required atm, until we do that Hack Friday to get pine typing return exact types.
For now the following worked for me:
```
/** @type {object} */
const app = latestRelease.belongs_to__application;
```

lib/models/application.js
thgreasi: Not sure why this caused the docs to change, but changing to the following stopped it:
```suggestion
						/** @private @type {?string} */
```

lib/models/application.js
Page-: Done

lib/models/application.js
Page-: Done

PR #871, opened 2020-04-04
Convert lib/balena.coffee to javascript

Page-:
Depends on #870


lib/util/global-env.ts
```diff
@@ -1,12 +1,14 @@
...
+let globalEnv: typeof window | typeof self | typeof global;
 // Use window (web)/self (web worker)/global (node) as appropriate
 if (typeof window !== 'undefined') {
 	globalEnv = window;
 } else if (typeof self !== 'undefined') {
 	globalEnv = self;
 } else if (typeof global !== 'undefined') {
 	globalEnv = global;
+} else {
+	throw new Error('Could not get global env');
```
Page-, line 11: This now matches the comment at the top and I think is safe as `setSharedOptions`/`fromSharedOptions` would both have blown up if we reached the previous `undefined` path

lib/balena.js
```diff
@@ -0,0 +1,350 @@
...
+ */
+getSdk.fromSharedOptions = function() {
+	const sharedOpts = globalEnv[BALENA_SDK_SHARED_OPTIONS];
+
+	globalEnv[BALENA_SDK_HAS_USED_SHARED_OPTIONS] = true;
+
+	return getSdk(sharedOpts);
+};
+
+modules.exports = getSdk;
```
thgreasi, line 350: Shouldn't this be
```suggestion
module.exports = getSdk;
```
like it was before?
It seems that the tests are complaining as well.

lib/balena.js
```diff
@@ -0,0 +1,350 @@
...
+	 * @example
+	 * balena.interceptors.push({
+	 * 	responseError: function (error) {
+	 * 		console.log(error);
+	 * 		throw error;
+	 * 	})
+	 * });
+	 */
+	Object.defineProperty(sdk, 'interceptors', {
+		/** @returns Interceptor[] */
```
thgreasi, line 189: Even though it might not feel 100% correct, we might need to add `@private` here to not cause changes in the docs.

PR #875, opened 2020-04-08
Abstract callback support out of the method implementations

thgreasi:
Regardless of whether we later decide to completely drop callback support or not, this already goes through all modules and removes the special callback handling. If we indeed later decide to drop support for callbacks we will just need to:
* delete util/callbacks.ts and it's references
* remove the callback related doc examples

Technically this changes a bit how callbacks work in method that don't accept optional callbacks, since we will now be using the `findCallback` helper even on those. But as far as a user is following the doc examples for the way they call the methods, this shouldn't cause any issues.

Also drops the node build size from 354kb to 345kb
and the minified browser build from 843kb to 840kb.

Change-type: patch
See: https://www.flowdock.com/app/rulemotion/resin-frontend/threads/XvE6YtBD5G4MqyGcVaVtBsnTilB

---
##### Contributor checklist

- [x] Includes tests
- [ ] Includes typings

### DIFF
--- a/lib/util/callbacks.ts
+++ b/lib/util/callbacks.ts
@@ -1,0 +1,61 @@
+import * as Bluebird from 'bluebird';
+import { Dictionary } from '../../typings/utils';
+
+// Use with: `findCallback(arguments)`.
+const findCallback = (args: IArguments): ((...args: any[]) => any) | void => {
+	const lastArg = args[args.length - 1];
+	if (typeof lastArg === 'function') {
+		return lastArg;
+	}
+
+	return;
+};
+
+const isThenable = (a: any): a is PromiseLike<any> => {
+	return typeof a === 'object' && typeof a.then === 'function';
+};
+
+const addCallbackSupport = <T extends (...args: any[]) => any>(fn: T): T => {
+	return function() {
+		const callback = findCallback(arguments);
+		const result = fn.apply(this, arguments);
+		if (!callback || !isThenable(result)) {
+			return result;
+		}
+
+		const bluebirdPromise =
+			result instanceof Bluebird
+				? (result as Bluebird<any>)
+				: Bluebird.resolve(result);
+
+		return bluebirdPromise.asCallback(callback);
+	} as T;
+};
+
+export const addCallbackSupportToModule = <T extends Dictionary<any>>(
+	sdkModule: T,
+): T => {
+	const result = {} as T;
+	for (const key of Object.keys(sdkModule) as Array<keyof typeof sdkModule>) {
+		const moduleProp = sdkModule[key];
+		const shouldAddCallback =
+			typeof key === 'string' &&
+			!key.startsWith('_') &&
+			typeof moduleProp === 'function';
+
+		result[key] = shouldAddCallback
+			? addCallbackSupport(moduleProp)
+			: moduleProp;
+	}
+	return result;
+};
+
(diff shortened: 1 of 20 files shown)


tests/integration/models/application.spec.coffee
```diff
@@ -23,16 +23,31 @@ describe 'Application Model', ->
...
+			it 'should eventually become an empty array [Promise]', ->
 				promise = balena.models.application.getAll()
 				m.chai.expect(promise).to.become([])
 
+			it 'should eventually become an empty array [callback]', (done) ->
+				balena.models.application.getAll (err, applications) ->
+					m.chai.expect(err).to.be.null
+					m.chai.expect(applications).to.deep.equal([])
+					done()
+				return
```
thgreasi, line 35: That's so that we are 100% sure that it was the `done` call that finished the test.

lib/util/callbacks.ts
```diff
@@ -0,0 +1,51 @@
...
+				: Bluebird.resolve(result);
+
+		return bluebirdPromise.asCallback(callback);
+	} as T;
+};
+
+export const addCallbackSupportToModule = <T extends Dictionary<any>>(
+	sdkModule: T,
+): T => {
+	const result = {} as T;
```
Page-, line 38:
```suggestion
	const result: T = {};
```

lib/util/callbacks.ts
```diff
@@ -0,0 +1,51 @@
...
+const addCallbackSupport = <T extends (...args: any[]) => any>(fn: T): T => {
+	return function() {
+		const callback = findCallback(arguments);
+		const result = fn.apply(this, arguments);
+		if (!callback || !isThenable(result)) {
+			return result;
+		}
+
+		const bluebirdPromise =
+			'asCallback' in result
```
Page-, line 27: You should also be able to use `result instanceof Bluebird`

lib/util/callbacks.ts
thgreasi: TS would complain since the assigned `{}` is missing atm all the props of T and that's why I needed the looser `as T`.
Otherwise we could do something:
```ts
const result: Partial<T> = {};
// for...
return result as T;
```
but I decided to stick with the approach of this PR since it had less type related code.

thgreasi:
@Page-  WRT lazyloading, I don't expect it to be making things worse, since:
* The callbacks related file is quite small and only depends on Bluebird, which is already a top level import in most files.
* We only require the callbacks code inside each model's getter, right before its construction.
* This should actually make the logs.js load faster, since it no longer needs to import the whole `util/index.js`.

lib/util/callbacks.ts
Page-: ok

Page-:
I was actually thinking if it triggers the loading eagerly and undoes the lazy element - it's not insurmountable but it would be good to be sure of.

It would also be nice imo if we could major change it to be included via `require('balena-sdk/callbacks')` so we can avoid the cost for anyone who uses promises but that's a breaking change and more of a stretch

thgreasi:
@Page- can you take an extra look on this?
I think that not having to deal with callbacks on each method separately makes both the code cleaner as well as saves us from having to type the callback in the TS files (which was a boring pain for my TS conversions so far).

thgreasi:
@balena-ci retest

PR #888, opened 2020-04-20
Lodash to native

Page-:
Resolves: # 
HQ: <url> 
See: <url> 
Depends-on: <url> 

---
##### Contributor checklist

- [ ] Includes tests
- [ ] Includes typings

### DIFF
--- a/lib/util/upsert.ts
+++ b/lib/util/upsert.ts
@@ -1,7 +1,5 @@
 import * as errors from 'balena-errors';
 import * as Promise from 'bluebird';
-import omit = require('lodash/omit');
-import pick = require('lodash/pick');
 import * as BalenaPine from '../../typings/balena-pine';
 import * as PineClient from '../../typings/pinejs-client-core';
 import { isUniqueKeyViolationResponse } from './index';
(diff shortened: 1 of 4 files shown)


lib/util/upsert.ts
```diff
@@ -43,13 +41,18 @@ export const getUpsertHelper = ({ pine }: { pine: BalenaPine.Pine }) => {
 			}
 
 			return pine.post<T>(params).catch(isUniqueKeyViolationResponse, () => {
+				const filter: { [key in keyof T]?: any } = {};
+				for (const naturalKeyProp of naturalKeyProps) {
```
sradevski, line 45: I am fine like this as well, but maybe for the few functions that are still helpful we can just create our own implementation in a `utils` file and use that instead? I'll leave it up to you.

Page-: Honestly for the ones that are useful and not easily replaceable with native modern js I'd carry on using them as they're well tested and typed - the ones that really stand out as being hard to replace to me are `cloneDeep`/`throttle`, `once` is a special case in that's it's fairly easy to replace but also absolutely tiny to keep (so I'd only replace if it was the last thing left). For this particular case though I used this style because it meant it was possible to do the pick/omit in one :)

Issue #887, opened 2020-04-20
Should have fully typed results


thgreasi:
## Expected Behavior
When we $select or $expand just a few props, then the result object should only contain those props & the included AssociatedResources should be typed only with the correct type variant.
($expanded AssociatedResources should be arrays, $selected NavigationResources should be typed as a PineDeferred)

## Actual Behavior
Currently we just return the complete interfaces of the resource with all props.

thgreasi:
Resolved by #889

PR #889, opened 2020-04-21
Fully type the pine.get results

thgreasi:
I decided to only initially release this for `pine.get` calls, since
* in that case it's not a major, since the previous typed method overloads are still working
* we can give this a good try on the UI and once we are fine with how it works we can do a major and release is for all calls (dropping the old "whole resource" typings)

Connects-to: #887
See: https://www.flowdock.com/app/rulemotion/resin-frontend/threads/FOa1dMbGF9pAqoaNd6vkWggxYIH
See: https://www.flowdock.com/app/rulemotion/resin-frontend/threads/5woCd-4nD7q-4p_YF9OO8KHzG3F
Change-type: minor

---
##### Contributor checklist

- [x] Includes tests
- [x] Includes typings

### DIFF
--- a/typing_tests/pine-params.ts
+++ b/typing_tests/pine-params.ts
@@ -1,0 +1,300 @@
+/// <reference types="node" />
+import * as BalenaSdk from '../typings/balena-sdk';
+import { AnyObject } from '../typings/utils';
+import { Compute, Equals, EqualsTrue } from './utils';
+
+const sdk: BalenaSdk.BalenaSDK = {} as any;
+
+let aAny: any;
+let aNumber: number;
+let aNumberOrUndefined: number | undefined;
+let aString: string;
+
+// This file is in .prettierignore, since otherwise
+// the @ts-expect-error comments would move to the wrong place
+
+// AnyObject pine queries
+
+(async () => {
+	const result = await sdk.pine.get<AnyObject>({
+		resource: 'device',
+		options: {
+			$select: ['device_name', 'uuid'],
+			$expand: {
+				belongs_to__application: {},
+				device_tag: {},
+			},
+		},
+	});
+	const test: Equals<typeof result, AnyObject[]> = EqualsTrue;
+})();
+
+(async () => {
+	const result = await sdk.pine.get<AnyObject>({
+		resource: 'device',
+		id: 1,
+		options: {
+			$select: ['device_name', 'uuid'],
+			$expand: {
+				belongs_to__application: {},
+				device_tag: {},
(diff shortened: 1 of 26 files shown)


typing_tests/pine-params.ts
```diff
@@ -0,0 +1,301 @@
+/// <reference types="node" />
+import * as BalenaSdk from '../typings/balena-sdk';
+import { AnyObject } from '../typings/utils';
+import * as PineClient from '../typings/pinejs-client-core';
```
JSReds, line 4: I can't find where this is used

typing_tests/utils.ts
```diff
@@ -1,4 +1,5 @@
-import { Any, Boolean } from 'ts-toolbelt';
+import { Any, Boolean, } from 'ts-toolbelt';
```
JSReds, line 1: To keep it consistent we can remove the `,` as long as it's on the same line

JSReds (commented):
Two small comments, other than that LGTM 👍

typing_tests/utils.ts
thgreasi: That's weird and probably means that we do not run linting as a test on ci...

typings/pinejs-client-core.d.ts
```diff
@@ -362,10 +444,34 @@ export interface SubscribeParamsWithId<T> extends ParamsObjWithId<T> {
 
 export interface Pine {
 	delete<T>(params: ParamsObjWithId<T> | ParamsObjWithFilter<T>): Promise<'OK'>;
-	get<T>(params: ParamsObjWithCount<T>): Promise<number>;
-	get<T>(params: ParamsObjWithId<T>): Promise<T | undefined>;
-	get<T>(params: ParamsObj<T>): Promise<T[]>;
-	get<T, Result>(params: ParamsObj<T>): Promise<Result>;
+	// Fully typed result overloads
+	get<
+		R extends keyof ResouceTypeMap,
```
JSReds, line 449:
```suggestion
		R extends keyof ResourceTypeMap,
```

thgreasi: Nice spot :+1:

sradevski (approved):
LGTM as far as I could tell, but there are plenty of changes so I might have missed something :) As long as the tests pass, I'm happy though, and we can fix if something was missed afterwards

JSReds (approved):
LGTM 👍

thgreasi:
@balena-ci retest

PR #890, opened 2020-04-21
Update dependencies

Page-:
Change-type: patch

Resolves: # 
HQ: <url> 
See: <url> 
Depends-on: <url> 

---
##### Contributor checklist

- [ ] Includes tests
- [ ] Includes typings


package.json
```diff
@@ -89,16 +89,16 @@
...
+    "@types/lodash": "^4.14.150",
     "@types/memoizee": "^0.4.3",
-    "@types/node": "^10.17.19",
+    "@types/node": "^10.17.20",
     "abortcontroller-polyfill": "^1.4.0",
     "balena-auth": "^3.0.1",
     "balena-errors": "^4.3.2",
-    "balena-hup-action-utils": "~4.0.0",
-    "balena-pine": "^11.0.0",
+    "balena-hup-action-utils": "^4.0.1",
```
thgreasi, line 98: This should still be a `~`, so that we only bump minors after the Proxy is also updated to the new version and released to prod.
It's unfortunate that this is still a thing.

Page-: done

Issue #767, opened 2019-10-30
Should have a deprecation policy


thgreasi:
As discussed during the Summit, the SDK should have a clear deprecation/end-of-life policy.

PR #892, opened 2020-04-21
Add a deprecation policy

thgreasi:
Resolves: #767
Change-type: minor
See: https://app.frontapp.com/open/cnv_7d92qx9
See: https://github.com/balena-io/balena/issues/2032
See: https://www.flowdock.com/app/rulemotion/r-product/threads/A6-bJBldfUFnhG9vixTz-DLz9iB

---
##### Contributor checklist

- [ ] Includes tests
- [ ] Includes typings

### DIFF
--- a/README.md
+++ b/README.md
@@ -71,7 +71,22 @@ See the JSDoc markdown documentation for the returned `balena` object in [DOCUME
 Support
 -------
 
-If you're having any problem, please [raise an issue](https://github.com/balena-io/balena-sdk/issues/new) on GitHub and the balena team will be happy to help.
+If you face any issues, please [raise an issue](https://github.com/balena-io/balena-sdk/issues/new) on GitHub and the balena team will be happy to help.
+
+Deprecation policy
+------------------
+
+The balena SDK uses [semver versioning](https://semver.org/), with the
+concepts of major, minor and patch version releases.
+
+The latest release of the previous major version of the balena SDK will be
+supported for one year from the date when the next major version is released.
+For example, the balena SDK v12.33.4 (the latest v12 release) will be supported for
+one year from the date when v13.0.0 is released.
+
+At the end of this support period, the older major version is considered
+deprecated and you are highly encouraged to upgrade to the to latest major
+version.
 
 Tests
 -----
--- a/README.md
+++ b/README.md
@@ -8,13 +8,11 @@ Balena SDK
 [![Build Status](https://travis-ci.org/balena-io/balena-sdk.svg?branch=master)](https://travis-ci.org/balena-io/balena-sdk)
 [![Build status](https://ci.appveyor.com/api/projects/status/qbsivehgnq0vyrrb/branch/master?svg=true)](https://ci.appveyor.com/project/balena-io/balena-sdk/branch/master)
 
-Role
-----
+## Role
 
 The intention of this module is to provide developers a nice API to integrate their JavaScript applications with balena.
 
-Installation
-------------
+## Installation
 
 Install the balena SDK by running:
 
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -71,7 +71,15 @@ See the JSDoc markdown documentation for the returned `balena` object in [DOCUME
...
-If you're having any problem, please [raise an issue](https://github.com/balena-io/balena-sdk/issues/new) on GitHub and the balena team will be happy to help.
+If you face any issues, please [raise an issue](https://github.com/balena-io/balena-sdk/issues/new) on GitHub and the balena team will be happy to help.
+
+Deprecation policy
+------------------
+
+The latest (minor,patch) release of each major balena-sdk version will be supported for a year after the next major version gets released.
+After that point older SDK versions will be considered as deprecated and you are highly encouraged to upgrade to the new major.
+
+For example, we will continue to support balena-sdk v12.33.0, which is the latest version of the v12 release, for a year after balena-sdk v13.0.0 got released.
```
iamsolankiamit, line 82: current latest is 13 right?

thgreasi: Indeed

pdcastro: Would it be a bit clearer to phrase it in terms of major releases, instead of (minor,patch)? For example:
```
`balena-sdk` uses [semver versioning](https://semver.org/), with the concepts of
major, minor and patch version releases.

The previous major version of balena-sdk will be supported for one year from the
date when the next major version is released. For example, balena-sdk v12.x.x
will be supported for one year from the date when v13.0.0 is released.

At the end of this support period, the older major version is considered deprecated
and you are highly encouraged to upgrade to the to latest major version.
```
Is there a meaningful difference between the following two statements? -
* Version `12.33.4` will be supported for a year from the date when v13 is released.
* Version `12.x.x` will be supported for a year from the date when v13 is released.

I have the impression that users would understand any of these two statements.

sradevski: I agree with Paulo here 👍

thgreasi: Loved your suggestion @pdcastro :+1: 

> Is there a meaningful difference between the following two statements? -

There is some difference indeed, that being that we will only support the latest release of that previous major. For example v12 was initially using the image maker endpoints which we want to deprecate, but changing to the respective API endpoints was only a minor release.

So will do some slight modification on your proposal to also include the above concept (as discussed in the arch call). Thanks for this, it was really a big improvement.

README.md
```diff
@@ -71,7 +71,22 @@ See the JSDoc markdown documentation for the returned `balena` object in [DOCUME
...
+Deprecation policy
+------------------
+
+The `balena-sdk` uses [semver versioning](https://semver.org/), with the
+concepts of major, minor and patch version releases.
+
+The latest release of the previous major version of the balena-sdk will be
+supported for one year from the date when the next major version is released.
+For example, balena-sdk v12.33.4 (the latest v12 release) will be supported for
+one year from the date when v13.0.0 is released.
```
pdcastro, line 85: NIT in this paragraph and the previous one:
* Some occurrences of `balena-sdk` are enclosed in backticks, others are not. If using a dash, I suggest always enclosing in backticks: `balena-sdk` _vs._ "balena SDK".
* Use of "the" before "balena-sdk". The way I read it:
  * _"The `balena-sdk` uses semver..."_ (bad)
  * _"The balena SDK uses semver..."_ (good)
  * _"`balena-sdk` uses semver"_ (good)

When written with a dash, and even more so if enclosed in backticks, `balena-sdk` sounds to me like the name of the package. Replace it with "Microsoft" to see how it sounds:

  * _"The Microsoft uses semver..."_ (bad)
  * _"The Microsoft SDK uses semver..."_ (good)
  * _"Microsoft uses semver"_ (good)

thgreasi: Makes sense @pdcastro :+1:  even though I don't feel that
> The `balena-sdk`

is that bad.

I will go with
> the balena SDK

approach since that's more consistent with the rest of the README.

PR #894, opened 2020-04-22
Add SDK methods to invite application members

iamsolankiamit:
Connects-to: https://github.com/balena-io/balena-ui/pull/3625
Depends-on: https://github.com/balena-io/balena-api/pull/2526
HQ: https://github.com/balena-io/balena/issues/2121
Change-type: minor
Signed-off-by: Amit Solanki <amit@balena.io>

---
##### Contributor checklist

- [x] Includes tests
- [x] Includes typings

### DIFF
--- a/lib/models/application-invite.ts
+++ b/lib/models/application-invite.ts
@@ -1,0 +1,263 @@
+/*
+Copyright 2020 Balena
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+     http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+import * as errors from 'balena-errors';
+import * as Promise from 'bluebird';
+import {
+	Application,
+	ApplicationInvite,
+	ApplicationMembershipRole,
+	PineOptionsFor,
+	ApplicationInviteOptions,
+	BalenaSDK,
+	PineSubmitBody,
+} from '../../typings/balena-sdk';
+import { InjectedDependenciesParam, InjectedOptionsParam } from '../balena';
+import { mergePineOptions } from '../util';
+
+const RESOURCE = 'invitee__is_invited_to__application';
+
+const getApplicationInviteModel = function (
+	deps: InjectedDependenciesParam,
+	opts: InjectedOptionsParam,
+	getApplication: BalenaSDK['models']['application']['get'],
+) {
+	const { request, pine } = deps;
+	const { apiUrl } = opts;
+
+	const exports = {
+		/**
+		 * @summary Get all invites
+		 * @name getAll
+		 * @public
+		 * @function
+		 * @memberof balena.models.application.invite
+		 *
+		 * @description
+		 * This method returns all invites.
+		 *
+		 * @param {Object} [options={}] - extra pine options to use
+		 * @fulfil {Object[]} - invites
+		 * @returns {Promise}
+		 *
+		 * @example
+		 * balena.models.application.invite.getAll().then(function(invites) {
+		 * 	console.log(invites);
+		 * });
+		 *
+		 * @example
+		 * balena.models.application.invite.getAll(function(error, invites) {
+		 * 	console.log(invites);
+		 * });
+		 */
+		getAll(
+			options: PineOptionsFor<ApplicationInvite> = {},
+		): Promise<ApplicationInvite[]> {
+			return pine.get({
+				resource: RESOURCE,
+				options,
+			});
+		},
+
+		/**
+		 * @summary Get all invites by application
+		 * @name getAllByApplication
+		 * @public
+		 * @function
+		 * @memberof balena.models.application.invite
+		 *
+		 * @description
+		 * This method returns all invites for a specific application.
+		 *
+		 * @param {String|Number} nameOrSlugOrId - application name (string), slug (string) or id (number)
+		 * @param {Object} [options={}] - extra pine options to use
+		 * @fulfil {Object[]} - invites
+		 * @returns {Promise}
+		 *
+		 * @example
+		 * balena.models.application.invite.getAllByApplication('MyApp').then(function(invites) {
+		 * 	console.log(invites);
+		 * });
+		 *
+		 * @example
+		 * balena.models.application.invite.getAllByApplication(123).then(function(invites) {
+		 * 	console.log(invites);
+		 * });
+		 *
+		 * @example
+		 * balena.models.application.invite.getAllByApplication(123, function(error, invites) {
+		 * 	console.log(invites);
+		 * });
+		 */
+		getAllByApplication(
+			nameOrSlugOrId: number | string,
+			options: PineOptionsFor<ApplicationInvite> = {},
+		): Promise<ApplicationInvite[]> {
+			return getApplication(nameOrSlugOrId, {
+				$select: 'id',
+			}).then(({ id }: Application) =>
+				exports.getAll(
+					mergePineOptions(
+						{ $filter: { is_invited_to__application: id } },
+						options,
+					),
+				),
+			);
+		},
+
+		/**
+		 * @summary Creates a new invite for an application
+		 * @name create
+		 * @public
+		 * @function
+		 * @memberof balena.models.application.invite
+		 *
+		 * @description This method invites a user by their email/username to an application.
+		 *
+		 * @param {String|Number} nameOrSlugOrId - application name (string), slug (string) or id (number)
+		 * @param {Object} options - invite creation parameters
+		 * @param {String} options.invitee - the email/balena_username of the invitee
+		 * @param {String} [options.roleName="developer"] - the role name to be granted to the invitee
+		 * @param {String} [message=null] - the message to send along with the invite
+		 *
+		 * @fulfil {String} - application invite
+		 * @returns {Promise}
+		 *
+		 * @example
+		 * balena.models.application.invite.create('MyApp', { collabortor: "invitee@example.org", roleName: "developer", message: "join my app" }).then(function(invite) {
+		 * 	console.log(invite);
+		 * });
+		 *
+		 * @example
+		 * balena.models.application.invite.create(53, { invitee: "invitee_username" }).then(function(invite) {
+		 * 	console.log(invite);
+		 * });
+		 *
+		 * @example
+		 * balena.models.application.invite.create(53, { invitee: "invitee_username" }, function(error, invite) {
+		 * 	console.log(invite);
+		 * });
+		 */
+		create(
+			nameOrSlugOrId: string | number,
+			{ invitee, roleName, message }: ApplicationInviteOptions,
+		): Promise<Partial<ApplicationInvite>> {
+			return Promise.all([
+				getApplication(nameOrSlugOrId, { $select: 'id' }).get('id'),
+				roleName
+					? pine.get<ApplicationMembershipRole>({
+							resource: 'application_membership_role',
+							options: {
+								$top: 1,
+								$select: ['id'],
+								$filter: {
+									name: roleName,
+								},
+							},
+					  })
+					: undefined,
+			]).then(([id, roles]) => {
+				type ApplicationInviteBase = Omit<ApplicationInvite, 'invitee'>;
+				type ApplicationInvitePostBody = ApplicationInviteBase & {
+					invitee: string;
+				};
+				const body: PineSubmitBody<ApplicationInvitePostBody> = {
+					invitee,
+					is_invited_to__application: id,
+					message,
+				};
+				if (roles) {
+					const [{ id: roleId }] = roles;
+					// Throw if the user provided a roleName, but we didn't find that role
+					if (!roleId && roleName) {
+						throw new errors.BalenaApplicationMembershipRoleNotFound(roleName);
+					}
+					body.application_membership_role = roleId;
+				}
+				return pine.post<ApplicationInviteBase>({
+					resource: RESOURCE,
+					body,
+				});
+			});
+		},
+
+		/**
+		 * @summary Revoke an invite
+		 * @name revoke
+		 * @public
+		 * @function
+		 * @memberof balena.models.application.invite
+		 *
+		 * @param {Number} id - application invite id
+		 * @returns {Promise}
+		 *
+		 * @example
+		 * balena.models.application.invite.revoke(123);
+		 *
+		 * @example
+		 * balena.models.application.invite.revoke(123,function(error) {
+		 * 	if (error) throw error;
+		 * 	...
+		 * });
+		 */
+		revoke(id: number): Promise<void> {
+			return pine.delete({ resource: RESOURCE, id }).return();
+		},
+
+		/**
+		 * @summary Accepts an invite
+		 * @name accept
+		 * @public
+		 * @function
+		 * @memberof balena.models.application.invite
+		 *
+		 * @description This method adds the calling user to the application.
+		 *
+		 * @param {String} invitationToken - invite token
+		 * @returns {Promise}
+		 *
+		 * @example
+		 * balena.models.application.invite.accept("qwerty-invitation-token");
+		 *
+		 * @example
+		 * balena.models.application.invite.accept("qwerty-invitation-token", function(error) {
+		 * 	if (error) throw error;
+		 * 	...
+		 * });
+		 */
+		accept(invitationToken: string) {
+			return Promise.try(() => {
+				return request
+					.send({
+						method: 'POST',
+						url: `/user/v1/invitation/${invitationToken}`,
+						baseUrl: apiUrl,
+					})
+					.get('body')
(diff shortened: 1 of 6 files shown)


lib/models/application-invite.ts
```diff
@@ -0,0 +1,280 @@
...
+	 * @fulfil {Object} - applicationInvite
+	 * @returns {Promise}
+	 *
+	 * @example
+	 * balena.models.applicationInvite.get(51).then(function(applicationInvite) {
+	 * 	console.log(applicationInvite);
+	 * });
+	 *
+	 * @example
+	 * balena.models.applicationInvite.get(51, function(error, applicationInvite) {
```
sradevski, line 57: I'd say no need to put a callback example, it's just noise nowadays.

lib/models/application-invite.ts
```diff
@@ -0,0 +1,280 @@
...
+	 * 	if (error) throw error;
+	 * 	console.log(applicationInvite);
+	 * });
+	 */
+	exports.get = function(
+		id: number,
+		options: PineOptionsFor<ApplicationInvite> = {},
+		callback?: (error?: Error, result?: ApplicationInvite) => void,
```
sradevski, line 65: I believe Thodoris abstracted the callback parameter, so once you rebase you'll need to remove them.

lib/models/application-invite.ts
```diff
@@ -0,0 +1,280 @@
...
+						],
+					},
+					options,
+				),
+			})
+			.asCallback(callback);
+	};
+
+	/**
+	 * @summary Get all Invites by application
```
sradevski, line 89: Let's make the caps consistent, make `Invites` lowercase here. Check how the rest of the docs are and match that.

lib/models/application-invite.ts
```diff
@@ -0,0 +1,280 @@
...
+	 * @name getAll
+	 * @public
+	 * @function
+	 * @memberof balena.models.applicationInvite
+	 *
+	 * @description
+	 * This method returns all invites of a specific application.
```
sradevski, line 96:
```suggestion
	 * This method returns all invites for a specific application.
```

lib/models/application-invite.ts
```diff
@@ -0,0 +1,280 @@
...
+	 *
+	 * @param {number} applicationId - Application Id for w
```
sradevski, line 98: what is the `w` supposed to be? I think we can rephrase this message to something like `Id of the application` or whatever else we use in the other resources.

lib/models/application-invite.ts
```diff
@@ -0,0 +1,280 @@
...
+	 * @param {Object} [options={}] - extra pine options to use
```
sradevski, line 99: Let's match the above docs message with this one.

lib/models/application-invite.ts
```diff
@@ -0,0 +1,280 @@
...
+	 * @fulfil {Object[]} - applicationInvites
+	 * @returns {Promise}
+	 *
+	 * @example
+	 * balena.models.applicationInvite.getAllByApplication(applicationId).then(function(applicationInvites) {
```
sradevski, line 104: Above we put a random ID number, so let's do the same here.

lib/models/application-invite.ts
```diff
@@ -0,0 +1,280 @@
...
+	 * 	console.log(applicationInvites);
+	 * });
+	 *
+	 * @example
+	 * balena.models.applicationInvite.getAllByApplication(applicationId, function(error, applicationInvites) {
```
sradevski, line 109: Again, I'll leave it to Thodoris to say what he prefers, but I'd say we don't have to add callback examples.

lib/models/application-invite.ts
```diff
@@ -0,0 +1,280 @@
...
+						},
+						$expand: {
+							invitee: {
+								$select: ['id', 'email'],
+							},
+							application_membership_role: {
+								$select: ['id', 'name'],
+							},
+						},
+						$select: ['id', 'message', 'created_at'],
```
sradevski, line 137: NIT: Put the `$select` on top, it's a bit easier to read

lib/models/application-invite.ts
```diff
@@ -0,0 +1,280 @@
...
+	};
+
+	/**
+	 * @summary Creates a new invite for an application
+	 * @name create
+	 * @public
+	 * @function
+	 * @memberof balena.models.applicationInvite
+	 *
+	 * @description This method invites a user by thier email/username to an application.
```
sradevski, line 152:
```suggestion
	 * @description This method invites a user by their email/username to an application.
```

lib/models/application-invite.ts
```diff
@@ -0,0 +1,280 @@
...
+	 *
+	 * @param {number} applicationId - the application id
+	 * @param {String} collaborator - the email/username of invitee
```
sradevski, line 155:
```suggestion
	 * @param {String} collaborator - the email/username of the invitee
```

lib/models/application-invite.ts
```diff
@@ -0,0 +1,280 @@
...
+	 * @param {String} roleTitle - the role title to be granted to the invitee
+	 * @param {String} [message=null] - the message to send along with the invite
+	 *
+	 * @fulfil {String} - application invite id
+	 * @returns {Promise}
+	 *
+	 * @example
+	 * balena.models.applicationInvite.create(applicationId,"invitee@example.org", "developer","join my app").then(function(applicationInvite) {
```
sradevski, line 163: Let's put a random appId number here as well, just so it matches the rest of the example.

lib/models/application-invite.ts
```diff
@@ -0,0 +1,280 @@
...
+	 * });
+	 */
+	exports.create = function(
+		applicationId: Application['id'],
+		collaborator: string,
+		roleTitle: ApplicationMembershipRole['name'],
+		message?: ApplicationInvite['message'],
+	) {
+		return pine
+			.get<ApplicationMembershipRole>({
```
sradevski, line 185: We can probably memoize this request, as it doesn't change often

lib/models/application-invite.ts
```diff
@@ -0,0 +1,280 @@
...
+					$filter: {
+						name: roleTitle,
+					},
+				},
+			})
+			.get(0)
+			.then((role: ApplicationMembershipRole) => {
+				return role.id;
+			})
+			.then(roleId => {
```
sradevski, line 198: We can combine this `.then` with the above one.

lib/models/application-invite.ts
```diff
@@ -0,0 +1,280 @@
...
+						is_invited_to__application: applicationId,
+						message,
+						application_membership_role: roleId,
+					},
+				});
+			});
+	};
+
+	/**
+	 * @summary Revoke an Application Invite
```
sradevski, line 212: Again the casing is inconsistent with the rest of the messages, let's stick with one format.

tests/integration/models/applicationInvite.spec.ts
```diff
@@ -0,0 +1,130 @@
...
+	describe('given no application invite', function() {
+		describe('balena.models.applicationInvite.getAllByApplication()', function() {
+			it('shoud return an empty Array', function() {
+				const promise = balena.models.applicationInvite.getAllByApplication(1);
+				return expect(promise).to.become([]);
+			});
+
+			it('should support a callback with no options', function(done) {
+				// tslint:disable-next-line:ban-types
+				(balena.models.applicationInvite.getAllByApplication as Function)(
```
thgreasi, line 28: Optional-NIT: Doing the following should allow you to remove the tslint hint.
```suggestion
				(balena.models.applicationInvite.getAllByApplication as (...args: any[]) => any))(
```

tests/integration/models/applicationInvite.spec.ts
```diff
@@ -0,0 +1,130 @@
...
+	givenLoggedInUser,
+} from '../setup';
+const { expect } = m.chai;
+
+const TEST_EMAIL = 'user.test@example.org';
+const TEST_MESSAGE = 'Hey!, Join my app on balenaCloud';
+const TEST_ROLE = 'developer';
+
+describe('Application Invite Model', function() {
+	givenLoggedInUser(beforeEach);
```
thgreasi, line 16:
```suggestion
	givenLoggedInUser(before);
```

tests/integration/models/applicationInvite.spec.ts
```diff
@@ -0,0 +1,130 @@
...
+	givenAnApplication(beforeEach);
```
thgreasi, line 17: Since doing this in a beforeEach is a slow operation for the API, which ends up slowing the SDK tests a lot, please consider removing this from here and adding a `givenAnApplication(before);` inside the
`describe('given no application invite'` & `describe('given a single application invite'`.

tests/integration/models/applicationInvite.spec.ts
```diff
@@ -0,0 +1,130 @@
...
+							TEST_ROLE,
+						);
+						expect(invites[0].message).to.equal(TEST_MESSAGE);
+					});
+			});
+		});
+	});
+
+	describe('given a single application invite', function() {
+		beforeEach(function() {
```
thgreasi, line 74: Let's also make this a `before` and add a `[contained scenario]` note right above
> describe('given a single application invite [contained scenario]',

lib/models/application-invite.ts
iamsolankiamit: Not sure how to do that

sradevski: @iamsolankiamit see `os.ts`, we memoize the device types call there

lib/models/application.js
```diff
@@ -1562,6 +1565,141 @@ const getApplicationModel = function (deps, opts) {
...
+			 *
+			 * @description This method adds the calling user to the application.
+			 *
+			 * @param {String} invitationToken - invite token
+			 * @returns {Promise}
+			 *
+			 * @example
+			 * balena.models.application.invite.acceptInvite("qwerty-invitation-token");
+			 */
+			acceptInvite: inviteModel.acceptInvite,
```
sradevski, line 1701:
```suggestion
			accept: inviteModel.accept,
```

lib/models/application.js
```diff
@@ -1562,6 +1565,141 @@ const getApplicationModel = function (deps, opts) {
 			 */
 			remove: envVarModel.remove,
 		},
+
+		/**
+		 * @namespace balena.models.application.invite
+		 * @memberof balena.models.application
+		 */
+		invite: {
```
thgreasi, line 1573: @iamsolankiamit You probably need to check https://github.com/balena-io/balena-sdk/pull/901 which adds back callback support for nested namespaces (essentially just wrap with an `addCallbackSupportToModule` call).

lib/models/application.js
```diff
@@ -1562,6 +1565,141 @@ const getApplicationModel = function (deps, opts) {
...
+			 * @function
+			 * @memberof balena.models.application.invite
+			 *
+			 * @param {Number} id - invite id
+			 * @param {Object} [options={}] - extra pine options to use
+			 * @fulfil {Object} - invite
+			 * @returns {Promise}
+			 *
+			 * @example
+			 * balena.models.application.invite.get(51).then(function(invite) {
```
thgreasi, line 1587: I remember that you also had callback examples didn't you?
Just to be clear I never meant to remove the callback examples, just the implementation part.

lib/models/application.js
```diff
@@ -1562,6 +1565,141 @@ const getApplicationModel = function (deps, opts) {
...
+			 * @param {Object} [options={}] - extra pine options to use
+			 * @fulfil {Object[]} - invites
+			 * @returns {Promise}
+			 *
+			 * @example
+			 * balena.models.application.invite.getAll().then(function(invites) {
+			 * 	console.log(invites);
+			 * });
+			 */
+			getAll: inviteModel.getAll,
```
thgreasi, line 1612: How about all these jsdoc comments to the `application-invite.ts` and just import and assign it here in the `invite` prop (instead of assigning each prop separately)?

lib/models/application.js
iamsolankiamit: @sradevski mentioned that they are no longer relevant and are noise so removed them.

lib/models/application.js
```diff
@@ -1564,6 +1567,12 @@ const getApplicationModel = function (deps, opts) {
 			 */
 			remove: envVarModel.remove,
 		}),
+
+		/**
+		 * @namespace balena.models.application.invite
+		 * @memberof balena.models.application
+		 */
+		invite: addCallbackSupportToModule(inviteModel),
```
thgreasi, line 1575: Are you sure that this doesn't need to be 
```suggestion
		invite: addCallbackSupportToModule(inviteModel()),
```

?? We will not have to do this change though if we drop the `once`.

lib/models/application.js
```diff
@@ -41,6 +41,9 @@ const getApplicationModel = function (deps, opts) {
 
 	const deviceModel = once(() => require('./device').default(deps, opts));
 	const releaseModel = once(() => require('./release').default(deps, opts));
+	const inviteModel = once(() =>
```
thgreasi, line 44: I think that the `once` here doesn't offer any benefit (just overhead), since are using the `inviteModel` in the same code block (while in other cases we use it inside method).

lib/models/application.js
```diff
@@ -41,6 +41,9 @@ const getApplicationModel = function (deps, opts) {
...
+		require('./application-invite').default(deps, opts, exports.get),
```
thgreasi, line 45: I think that this needs to be:
```suggestion
		require('./application-invite').default(deps, opts, (nameOrSlugOrId) => exports.get(nameOrSlugOrId)),
```
since `exports.get` isn't yet defined (and I don't expect the `once` to have any effect, since we are calling this right away).

tests/integration/models/applicationInvite.spec.ts
```diff
@@ -0,0 +1,132 @@
...
+			});
+		});
+
+		describe('balena.models.application.invite.create()', function () {
+			it('should create and return an application invite', async function () {
+				const response = await balena.models.application.invite.create(
+					this.application.id,
+					{
+						collaborator: TEST_EMAIL,
+						roleName: TEST_ROLE,
```
thgreasi, line 45: Reminder to add:
* a test that the user isn't logged in
* a test w/ an app param for an app that the user doesn't have access to (eg: 999999)
* a test w/ an invalid role
* a test w/o a specified role

tests/integration/models/applicationInvite.spec.ts
```diff
@@ -0,0 +1,132 @@
...
+			});
+
+			it('should support arbitrary pinejs options', async function () {
+				const applicationInvites = await balena.models.application.invite.getAllByApplication(
+					this.application.id,
+					{
+						$expand: { invitee: { $select: ['email'] } },
+					},
+				);
+				expect(applicationInvites[0].invitee[0].email).to.equal(TEST_EMAIL);
```
thgreasi, line 102: You should be able to do:
```suggestion
				assertDeepMatchAndLength(applicationInvites, [applicationInvites: [{email: TEST_EMAIL}]]
```

You will need to add the following import:
```
const { assertDeepMatchAndLength } = require('../../util');
```

tests/integration/models/applicationInvite.spec.ts
```diff
@@ -0,0 +1,132 @@
...
+			it('should be able to get an invite', async function () {
+				const applicationInvite = await balena.models.application.invite.get(
+					this.application.invite.id,
+				);
+				expect(applicationInvite.message).to.equal(TEST_MESSAGE);
+			});
+
+			it('should be rejected if the applicationInvite id is invalid', function () {
+				const promise = balena.models.application.invite.get(99999999999);
+				return expect(promise).to.be.rejectedWith('Request error');
```
thgreasi, line 116: It would be great to also check the status code w/ something like
```
				return expect(promise).to.be.rejected.then(function (error) {
					expect(error).to.have.property('code', 'BalenaRequestError');
					expect(error).to.have.property('statusCode', 404); // For example. I don't know what this really returns.
					expect(error)
						.to.have.property('message')
						.that.contains('Request error');
				});
```

tests/integration/models/applicationInvite.spec.ts
thgreasi: Actually while trying to help avoiding the `any` import here, I've opened https://github.com/balena-io/balena-sdk/pull/904.
After that you should be able to use a normal
```
import { assertDeepMatchAndLength } from '../../util';
```

lib/models/application-invite.ts
```diff
@@ -0,0 +1,296 @@
...
+		 *
+		 * @example
+		 * balena.models.application.invite.get(51, {}, function(invite) {
+		 * 	console.log(invite);
+		 * });
+		 */
+		get(
+			id: number,
+			options: PineOptionsFor<ApplicationInvite> = {},
+		): Promise<ApplicationInvite> {
```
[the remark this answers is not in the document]
iamsolankiamit: Actually there is no need of a `.get` method now, so let's drop it. If needed in future we might add it later

lib/models/application-invite.ts
```diff
@@ -0,0 +1,263 @@
...
+							options: {
+								$top: 1,
+								$select: ['id'],
+								$filter: {
+									name: roleName,
+								},
+							},
+					  })
+					: undefined,
+			]).then(([id, roles]: any[]) => {
```
thgreasi, line 172:
```suggestion
			]).then(([id, roles]) => {
```

iamsolankiamit:
@balena-ci rebase

PR #902, opened 2020-04-28
Split util function imports and lazy load them

thgreasi:
This allows us to avoid loading:
* lodash/throttle
* memoizee
* moment
* balena-hup-action-utils

in all files that import from the util.ts
and also adds lazy loading when possible.

Change-type: patch

Resolves: # 
HQ: <url> 
See: <url> 
Depends-on: <url> 

---
##### Contributor checklist

- [ ] Includes tests
- [ ] Includes typings

### DIFF
--- a/lib/models/os.ts
+++ b/lib/models/os.ts
@@ -21,15 +21,9 @@ import once = require('lodash/once');
 import * as memoizee from 'memoizee';
 import * as semver from 'semver';
 
-import {
-	deviceTypes as deviceTypesUtil,
-	isDevelopmentVersion,
-	isNotFoundResponse,
-	onlyIf,
-	treatAsMissingApplication,
-} from '../util';
-
-import { hupActionHelper } from '../util/device-actions/os-update/utils';
+import { isNotFoundResponse, onlyIf, treatAsMissingApplication } from '../util';
+import * as deviceTypesUtils from '../util/device-types';
+import type { hupActionHelper as hupActionHelperType } from '../util/device-actions/os-update/utils';
 
 import { BalenaRequestStreamResult } from '../../typings/balena-request';
 import {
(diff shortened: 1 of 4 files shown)


lib/util/index.ts
```diff
@@ -96,9 +70,6 @@ export const treatAsMissingDevice = (uuidOrId: string | number) => (
 	throw replacementErr;
 };
 
-export const isDevelopmentVersion = (version: string) =>
```
thgreasi, line 99: It was only used in os.ts, so I just moved it there, since it might also make sense to export this method at some point.

lib/util/index.ts
```diff
@@ -252,25 +223,3 @@ const convertExpandToObject = <T extends {}>(
 
 	return cloneDeep(expandOption);
 };
-
-// In order not to introduce a breaking change, we export each element independently and all together as a default export.
-export default {
```
thgreasi, line 257: We no longer re-export everything in a default export :tada: 
Now that nothing is in coffee, it was clear that we can still use the individual exports with `require()` calls.

lib/util/index.ts
```diff
@@ -1,21 +1,13 @@
...
-import * as dt from './device-types';
 
 export interface ErrorResponse {
 	code: string;
 	statusCode?: number;
 	body?: string;
 }
 
-export const getOsUpdateHelper = updateHelper;
-export const deviceTypes = dt;
```
thgreasi, line 17: These re-exports were convenient, but also only used for very small % of the files/methods.

lib/models/os.ts
```diff
@@ -21,15 +21,9 @@ import once = require('lodash/once');
...
-	deviceTypes as deviceTypesUtil,
-	isDevelopmentVersion,
-	isNotFoundResponse,
-	onlyIf,
-	treatAsMissingApplication,
-} from '../util';
-
-import { hupActionHelper } from '../util/device-actions/os-update/utils';
+import { isNotFoundResponse, onlyIf, treatAsMissingApplication } from '../util';
+import * as deviceTypesUtils from '../util/device-types';
```
thgreasi, line 25: B/c `isOsArchitectureCompatibleWith` is re-exported by this module there is no point of lazy loading this.

PR #903, opened 2020-04-29
Lazy load balena-register-device & the device type utils

thgreasi:
Change-type: patch

---
##### Contributor checklist

- [ ] Includes tests
- [ ] Includes typings

### DIFF
--- a/lib/models/device.js
+++ b/lib/models/device.js
@@ -1867,7 +1867,10 @@ const getDeviceModel = function (deps, opts) {
 		 * // randomKey is a randomly generated key that can be used as either a uuid or an api key
 		 * console.log(randomKey);
 		 */
-		generateUniqueKey: registerDevice().generateUniqueKey,
+		generateUniqueKey: function (...args) {
+			// Wrap with a function so that we can lazy load registerDevice
+			return registerDevice().generateUniqueKey(...args);
+		},
 
 		/**
 		 * @summary Register a new device with a Balena application.
--- a/lib/models/os.ts
+++ b/lib/models/os.ts
@@ -22,7 +22,7 @@ import * as memoizee from 'memoizee';
 import * as semver from 'semver';
 
 import { isNotFoundResponse, onlyIf, treatAsMissingApplication } from '../util';
-import * as deviceTypesUtils from '../util/device-types';
+import type * as deviceTypesUtilsType from '../util/device-types';
 import type { hupActionHelper as hupActionHelperType } from '../util/device-actions/os-update/utils';
 
 import { BalenaRequestStreamResult } from '../../typings/balena-request';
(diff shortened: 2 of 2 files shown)


lib/models/device.js
```diff
@@ -1867,7 +1867,10 @@ const getDeviceModel = function (deps, opts) {
 		 * // randomKey is a randomly generated key that can be used as either a uuid or an api key
 		 * console.log(randomKey);
 		 */
-		generateUniqueKey: registerDevice().generateUniqueKey,
+		generateUniqueKey: function () {
+			// Wrap with a function so that we can lazy load registerDevice
+			return registerDevice().generateUniqueKey.apply(this, arguments);
```
Page-, line 1872: I'd prefer ...arguments as it keeps the 'this' binding the same as device registry will expect

lib/models/os.ts
```diff
@@ -49,6 +49,11 @@ const getOsModel = function (
 	const applicationModel = once(() =>
 		require('./application').default(deps, opts),
 	);
+	const deviceTypesUtils = once(
```
Page-, line 52: If you check the cli lazy.ts pattern I have all the lazy load imports set up in one file so it can easily be shared by everything and only get typed once, you could also combo it with the import blacklist to enforce that only the lazy file can actually import them

lib/models/device.js
thgreasi: Actually, since we previously assigned the function to the exported method, the `this` w/ this approach is exactly as before.
But since the `this` isn't relevant for the `generateUniqueKey()` changing will also communicate that to future readers of this code :)

lib/models/os.ts
thgreasi: Oh nice :+1: 
Is it fine if I leave that for a follow-up?

lib/models/device.js
thgreasi: Tried `...arguments` but TS required an extra `Array.from` in order to allow `...`.
So I ended up doing the equivalent with `...args`.

lib/models/os.ts
```diff
@@ -621,8 +626,13 @@ const getOsModel = function (
 	 * const result2 = balena.models.os.isArchitectureCompatibleWith('armv7hf', 'amd64');
 	 * console.log(result2);
 	 */
-	const isArchitectureCompatibleWith =
-		deviceTypesUtils.isOsArchitectureCompatibleWith;
+	const isArchitectureCompatibleWith = function () {
+		// Wrap with a function so that we can lazy load the deviceTypesUtils
+		return deviceTypesUtils().isOsArchitectureCompatibleWith.apply(
```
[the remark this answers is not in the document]
thgreasi: Oh well...
Shame on me for forgetting that...
I guess it's the repo PR context- switch, sorry :sweat:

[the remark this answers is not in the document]
thgreasi: Also switched to
```
const isArchitectureCompatibleWith: ReturnType<typeof deviceTypesUtils>['isOsArchitectureCompatibleWith'] =
```
which gives stricter type checking and allows us to not repeat the typing for the arguments in the function declaration :)

thgreasi:
@balena-ci retest

thgreasi:
Woohoo! My first auto-merged SDK PR :tada:

Issue #905, opened 2020-04-30
Should offer a non-cached way to check with the api whether the user is logged in/whoami


thgreasi:
## Expected Behavior

The auth module should offer a way to recheck with the API whether the session is still valid.
We should consider the impact on on the CLI & UI.
Changing the `auth.isLoggedIn()` atm seems to have less impact than changing `auth.whoami()`.

## Actual Behavior

Atm most `sdk.auth` method use a cached response from the `/user/v1/whoami` endpoint,
So logging out from other sessions wouldn't be surfaced by those methods if they already have a cached reponse.

## Steps to Reproduce the Problem

1. login in two sessions
1. use an auth method (eg `auth.isLoggedId` or `auth.whoami`) in session A
1. use the "loggout from other sessions" in session B
1. call again that same authe method in session A

The issue also happens in case an api token is used and it then gets revoked,

## Specifications

- SDK version:
- Browser: 
- Node version: 
- Npm version: 
- OS:

PR #449, opened 2018-01-03
V8

MoranF:
---- Autogenerated Waffleboard Connection: Connects to #364

PR #906, opened 2020-04-30
auth.isLoggedIn: Change to always confirm with the API

thgreasi:
We do need a way to check with the API's `/user/v1/whoami` endpoint for the cases like when the "logout other sessions" was used.

Based on the history it looks like the caching was adding so that the CLI doesn't do tons of separate requests for each action. B/c of the approach atm of caching only the result, the CLI atm does 2 requests even just for the `balena whoami` action, one of which is for event reporting and the other to confirm that the user is logged in.

In an effort to not increase the amount of requests that the CLI does, I didn't completely drop the caching but decided to only drop it in the `auth.isLoggedIn()` method. An alt choice was the `auth.whoami()`, but given the docs for the two methods:
> isLoggedIn: Check if you're logged in
> whoami: Return current logged in username

going with the `auth.isLoggedIn()` felt a bit more like a bugfix rather than a "big" change, which also doesn't affect the number of requests that the CLI is doing atm.

As an extra note, this is actually how the `auth.isLoggedIn()` worked on v7 (it always did an API request).
See: https://github.com/balena-io/balena-sdk/pull/449/commits/f1369f0ca3353d044c429efe0d021488480298d7?file-filters%5B%5D=.coffee&file-filters%5B%5D=.json&file-filters%5B%5D=.lock&file-filters%5B%5D=.ts#diff-f1ecd54135ef0d1ecf8300a80f5ac38aL191-L195

Resolves: #905
Change-type: minor

---
##### Contributor checklist

- [ ] Includes tests
- [ ] Includes typings

### DIFF
--- a/lib/auth.ts
+++ b/lib/auth.ts
@@ -16,6 +16,7 @@ limitations under the License.
 
 import * as errors from 'balena-errors';
 import * as Promise from 'bluebird';
+import * as memoizee from 'memoizee';
 // TODO: change to type-only import when we bump TS to 3.8
 import _get2faModel from './2fa';
 import { InjectedDependenciesParam, InjectedOptionsParam } from './balena';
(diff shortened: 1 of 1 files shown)


thgreasi:
My research on where isLogged & whoami are use in the CLI:

# auth.isLoggedIn

auth/utils.ts: loginIfTokenValid (right after a loginWithToken() call)
used in:
* auth/server: awaitForToken()

utils/patterns.ts: checkLoggedIn
used in:
* balena env(s) <command>
* balena os configure, inside validateOptions()
* balena build, when an app arg is used
* balena push
* balena ssh, on non-local devices
* utils/patterns: exitIfNotLoggedIn
  * lib/app-capitano, when the user permission is required
  * balena logs, on non-local devices

# auth.whoami

balena-sync (the module)

balena whoami
balena tunnel, in utils/tunnel tunnelConnectionToDevice()
utils/promote, createApplication()
 * getOrSelectApplication()
   * balena join
events, trackCommand()
  * hooks/prerun/track
    * app-oclif, run()
balena ssh, after checkLoggedIn()
balena deploy, deployProject(), but also has `permission: 'user'`
balena login, after loginWithToken()
balena logout, but also has `permission: 'user'`

PR #908, opened 2020-05-01
Avoid moment v2.25.0 that's missing published files

thgreasi:
Moment v2.25.1 isn't yet released, even though a fix has been implemented, so I decided to use `||` to unblock the SDK tests right away and not limit consumers from getting the bugfixes that get released after 2.25.0.

Change-type: patch
See: https://www.flowdock.com/app/rulemotion/d-build-pipelines/threads/dumJIBFDwtOJzB9_FYYmZO1u6zA

---
##### Contributor checklist

- [ ] Includes tests
- [ ] Includes typings


sradevski:
Maybe you can open an issue so we remember to clean this up a bit? It's not a big deal, but `||` stuff are not very pretty in the dependency versions :)

thgreasi:
I plan to revert as soon as the fix goes out, let's say on Monday.

PR #909, opened 2020-05-04
Avoid unnecessary `cloneDeep` usage

Page-:
Change-type: patch

Resolves: # 
HQ: <url> 
See: <url> 
Depends-on: <url> 

---
##### Contributor checklist

- [ ] Includes tests
- [ ] Includes typings

### DIFF
--- a/lib/util/index.ts
+++ b/lib/util/index.ts
@@ -1,5 +1,4 @@
 import * as errors from 'balena-errors';
-import cloneDeep = require('lodash/cloneDeep');
 import * as Pine from '../../typings/pinejs-client-core';
 
 export interface ErrorResponse {
@@ -87,7 +86,7 @@ export const mergePineOptions = <R extends {}>(
 		return defaults;
 	}
 
-	const result = cloneDeep(defaults);
+	const result = { ...defaults };
 
 	for (const option of Object.keys(extras) as Array<keyof typeof extras>) {
 		switch (option) {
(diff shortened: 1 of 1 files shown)


lib/util/index.ts
```diff
@@ -87,7 +86,7 @@ export const mergePineOptions = <R extends {}>(
 		return defaults;
 	}
 
-	const result = cloneDeep(defaults);
+	const result = { ...defaults };
```
Page-, line 89: We clone the top level object as we mutate it heavily, for the lower levels only $expand actually gets mutated and we handle that separately

lib/util/index.ts
```diff
@@ -145,7 +144,8 @@ const mergeExpandOptions = <T>(
 		return extraExpand;
 	}
 
-	defaultExpand = convertExpandToObject(defaultExpand);
+	// We only need to clone the defaultExpand as it's the only one we mutate
+	defaultExpand = convertExpandToObject(defaultExpand, true);
```
Page-, line 148: As the comment says ^

lib/util/index.ts
```diff
@@ -221,5 +222,9 @@ const convertExpandToObject = <T extends {}>(
 		}
 	}
 
-	return cloneDeep(expandOption);
+	if (cloneIfNeeded) {
+		return { ...(expandOption as Pine.ResourceExpandFor<T>) };
+	}
```
Page-, line 227: We only need to clone here since the other steps return a object

thgreasi:
Really nice :+1:

PR #911, opened 2020-06-02
Generate optional build for es2018 as well as the default es2015

Page-:
Change-type: minor

Resolves: # 
HQ: <url> 
See: <url> 
Depends-on: <url> 

---
##### Contributor checklist

- [ ] Includes tests
- [ ] Includes typings


package.json
```diff
@@ -36,8 +40,10 @@
     "test:ts-js": "npx tsc --noEmit --project ./tsconfig.js.json",
     "test:ts-compatibility": "npx typescript@~3.8.2 --noEmit --project ./tsconfig.dist.json",
     "test:typings": "dtslint --localTs node_modules/typescript/lib --expectOnly typing_tests",
-    "build:ts": "tsc && gulp inject-version",
-    "build": "npm run lint && npm run clean && npm run build:ts && gulp build && npm run docs",
+    "build-es5": "rimraf es5 && tsc --target es5 --outDir es5 && gulp inject-version --buildDir es5 && gulp pack-browser --buildDir es5",
+    "build-es2015": "rimraf es2015 && tsc --target es2015 --outDir es2015 && gulp inject-version --buildDir es2015 && gulp pack-browser --buildDir es2015",
+    "build-es2018": "rimraf es2018 && tsc --target es2018 --outDir es2018 && gulp inject-version --buildDir es2018 && gulp pack-browser --buildDir es2018",
```
thgreasi, line 45: How about having pack-browser only in one of the builds?

Page-: The idea of having it in each is that downstream can use `balena-sdk/es2015/browser` to get a packed build for a specific es version

thgreasi: More specifically I think that we should have:
```
"build-es5:ts": "rimraf es5 && tsc --target es5 --outDir es5 && gulp inject-version --buildDir es5 && gulp pack-browser --buildDir es5",
"build-es5": "build-es5:ts && gulp pack-browser --buildDir es5",
```
and then have `test:fast` use `build-es5:ts` so that we don't suffer the browserify build time while developing.

package.json
```diff
@@ -36,8 +40,10 @@
     "test:ts-js": "npx tsc --noEmit --project ./tsconfig.js.json",
     "test:ts-compatibility": "npx typescript@~3.8.2 --noEmit --project ./tsconfig.dist.json",
     "test:typings": "dtslint --localTs node_modules/typescript/lib --expectOnly typing_tests",
-    "build:ts": "tsc && gulp inject-version",
-    "build": "npm run lint && npm run clean && npm run build:ts && gulp build && npm run docs",
+    "build-es5": "rimraf es5 && tsc --target es5 --outDir es5 && gulp inject-version --buildDir es5 && gulp pack-browser --buildDir es5",
```
thgreasi, line 43: I'm not sure whether there are ppl using `unpkg` or similar to get the browserify build.
I do vaguely remember that we once pointed someone on support to use that.
Not a strong opinion on this, but if we care we should use the old build directory for the `gulp pack-browser` output.

package.json
thgreasi: Oh I see, but I would still like a lighter variant for `test:fast` if you agree.

Page-: Done

package.json
Page-: I think if they're using a floating `unpkg` path then they've got bigger problems than the path changing, but I've updated the mentions to it

thgreasi:
@balena-ci rebase

PR #921, opened 2020-06-26
model: Add build_environment_variable

richbayliss:
Change-type: patch
Depends-on: https://github.com/balena-io/balena-api/pull/2652
Signed-off-by: Rich Bayliss <rich@balena.io>

### DIFF
--- a/lib/models/application.js
+++ b/lib/models/application.js
@@ -85,6 +85,17 @@ const getApplicationModel = function (deps, opts) {
 			},
 		},
 	);
+	const buildVarModel = buildDependentResource(
+		{ pine },
+		{
+			resourceName: 'build_environment_variable',
+			resourceKeyField: 'name',
+			parentResourceName: 'application',
+			getResourceId(nameOrSlugOrId) {
+				return exports.get(nameOrSlugOrId, { $select: 'id' }).get('id');
+			},
+		},
+	);
 
 	// Infer dashboardUrl from apiUrl if former is undefined
 	if (dashboardUrl == null) {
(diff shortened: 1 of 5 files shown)


typings/balena-sdk.d.ts
```diff
@@ -317,6 +317,7 @@ declare namespace BalenaSdk {
 		application_environment_variable: ReverseNavigationResource<
 			ApplicationVariable
 		>;
+		build_environment_variable: ReverseNavigationResource<ApplicationVariable>;
```
thgreasi, line 320: Even thought those two are the same, let's create a separate `BuildVariable` type.

lib/models/application.js
```diff
@@ -1650,6 +1661,131 @@ const getApplicationModel = function (deps, opts) {
 			remove: envVarModel.remove,
 		}),
 
+		/**
+		 * @namespace balena.models.application.buildEnvVar
```
thgreasi, line 1665: Not 100% sure about this, but for consistency I think the module's namespace should be `buildVar`.
What do you think?

lib/models/application.js
```diff
@@ -1650,6 +1661,131 @@ const getApplicationModel = function (deps, opts) {
...
+		 * @memberof balena.models.application
+		 */
+		buildEnvVar: addCallbackSupportToModule({
```
thgreasi, line 1668: Don't forget to add typings in the `balena-sdk.d.ts` for the new model.

typings/balena-sdk.d.ts
richbayliss: No probs

lib/models/application.js
richbayliss: I thought to keep to the API model name *build*_*env*ironment_*var*iable. If you'd rather go with `buildVar` then I am not going to fight it 👍

thgreasi: I'm fine with either :

PR #935, opened 2020-07-06
Minimize bluebird sugar usage

Page-:
Change-type: patch

Resolves: # 
HQ: <url> 
See: <url> 
Depends-on: <url> 

---
##### Contributor checklist

- [ ] Includes tests
- [ ] Includes typings

### DIFF
--- a/lib/models/device.js
+++ b/lib/models/device.js
@@ -97,7 +97,7 @@ const getDeviceModel = function (deps, opts) {
 			resourceKeyField: 'tag_key',
 			parentResourceName: 'device',
 			getResourceId(uuidOrId) {
-				return exports.get(uuidOrId, { $select: 'id' }).get('id');
+				return exports.get(uuidOrId, { $select: 'id' }).then(({ id }) => id);
 			},
 		},
 	);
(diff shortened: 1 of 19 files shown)


lib/models/device.js
```diff
@@ -702,7 +708,12 @@ const getDeviceModel = function (deps, opts) {
 			exports
 				.get(uuidOrId, { $select: ['id'] })
 				.return(true)
```
thgreasi, line 710: How about also this?

lib/models/device.js
```diff
@@ -893,7 +906,7 @@ const getDeviceModel = function (deps, opts) {
 						},
 					}),
 				)
-				.return(undefined),
+				.return(),
```
thgreasi, line 909: How about converting this one as well?

lib/models/device.js
Page-: I did, fixed

lib/models/device.js
Page-: I specifically left this because it's an easy pattern to find/remove in an async/await world whereas in a promise world it's quite a pain, so leaving it as is makes life easier overall imo

PR #938, opened 2020-07-07
Add type checking for tests

Page-:
Change-type: patch

Resolves: # 
HQ: <url> 
See: <url> 
Depends-on: <url> 

---
##### Contributor checklist

- [ ] Includes tests
- [ ] Includes typings


tests/util.ts
```diff
@@ -12,6 +12,6 @@ export const assertDeepMatchAndLength = (a: any, b: any) => {
 		expect(a).to.deep.equal(b);
 	}
 
-	expect(a).to.deep.match(b);
+	expect(a).to.deep.equals(b);
```
Page-, line 15: `match` is intended for regexps, based on the usage of this `deep.equals` looked like the correct/valid replacement

thgreasi: @Page- `deep.match` is actually added by the `chai-samsam` module.
it's essentially a partial checking function and is needed here.

tests/integration/models/config.spec.ts
```diff
@@ -4,7 +4,9 @@ import * as m from 'mochainon';
 import { balena, BalenaSdk } from '../setup';
 const { expect } = m.chai;
 
-const expectDeviceTypeArray = function (deviceTypes: BalenaSdk.DeviceType[]) {
+const expectDeviceTypeArray = function (
+	deviceTypes: BalenaSdk.DeviceTypeJson.DeviceType[],
```
thgreasi, line 8: :sweat:

PR #939, opened 2020-07-07
Convert all remaining tests away from cofffeescript

Page-:
Depends on #938


tests/integration/setup.js
```diff
@@ -0,0 +1,520 @@
...
+				}
+
+				return balena.pine.patch({
+					resource: 'device',
+					body: {
+						is_running__release: this.currentRelease.id,
+					},
+					options: {
+						$filter: {
+							uuid: deviceInfo.uuid,
```
Page-, line 264: @thgreasi this fails to type check because of the magic around body props which means you can only filter on props that are also included in the body - how are we meant to get around those typings with force casts (which are very hard to do in javascript+prettier) and ideally without just `@ts-ignore`ing it

thgreasi: :disappointed: I should have set `AnyObject` as the default typing...
The only reason I didn't do so, is that consumers will have to explicitly provide the type, in order to have this type checked, while if I indeed had `= AnyObject`, we could have untyped pine calls all over the place, just b/c we forgot to provide the type.
In the next major the SDK will have a list of known resources and pick the type auto-magically :)

PR #941, opened 2020-07-07
typings: Export pine variant w/ a mandatory $select on get requests

thgreasi:
This allows at least balena modules to opt-in type checking that enforces `$select`s, while not hurting the development & discoverability of properties for other consumers.

Change-type: minor
See: https://github.com/balena-io/balena-ui/pull/3829/files#r450310702

---
##### Contributor checklist

- [ ] Includes tests
- [ ] Includes typings

### DIFF
--- a/typings/balena-pine.d.ts
+++ b/typings/balena-pine.d.ts
@@ -14,3 +14,14 @@ export interface Pine {
 	): Bluebird<'OK'>;
 	upsert<T>(params: PineClient.UpsertParams<T>): Bluebird<T | 'OK'>;
 }
+
+export type PineWithSelectOnGet = Omit<Pine, 'get'> & {
+	get<T>(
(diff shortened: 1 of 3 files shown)


typings/balena-pine.d.ts
```diff
@@ -14,3 +14,14 @@ export interface Pine {
 	): Bluebird<'OK'>;
 	upsert<T>(params: PineClient.UpsertParams<T>): Bluebird<T | 'OK'>;
 }
+
+export type PineWithSelectOnGet = Omit<Pine, 'get'> & {
```
thgreasi, line 18: I initially tried having just a `WithSelectOnGet<T>` type helper using mapped types, but currently mapped typed in TypeScript only pick the last (or maybe most generic) overload of the method.

PR #943, opened 2020-07-08
Improve test typings

Page-:
Resolves: # 
HQ: <url> 
See: <url> 
Depends-on: <url> 

---
##### Contributor checklist

- [ ] Includes tests
- [ ] Includes typings

### DIFF
--- a/tests/integration/models/os.spec.ts
+++ b/tests/integration/models/os.spec.ts
@@ -5,12 +5,12 @@ import * as _ from 'lodash';
 import * as m from 'mochainon';
 import {
 	balena,
-	BalenaSdk,
 	credentials,
 	givenAnApplication,
 	givenLoggedInUser,
 	IS_BROWSER,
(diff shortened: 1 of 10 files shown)


tests/integration/models/os.spec.ts
```diff
@@ -5,23 +5,33 @@ import * as _ from 'lodash';
...
 const eventuallyExpectProperty = <T>(promise: Promise<T>, prop: string) =>
 	expect(promise).to.eventually.have.property(prop);
 
+const {
+	_getDeviceTypes,
+	_getOsVersions,
+	_getDownloadSize,
+	_getMaxSatisfyingVersion,
+	_clearDeviceTypesEndpointCaches,
+} = balena.models.os as ReturnType<
```
thgreasi, line 25: I would prefer `const osModel = balena.models.os as ... ` to be clear where these come from, but I'm fine with it.

PR #947, opened 2020-07-13
Add an organization parameter to all billing methods

thgreasi:
Change-type: minor

---
##### Contributor checklist

- [ ] Includes tests
- [ ] Includes typings

### DIFF
--- a/DOCUMENTATION.md
+++ b/DOCUMENTATION.md
@@ -296,12 +296,12 @@ const sdk = getSdk.fromSharedOptions();
             * [.get(id, [options])](#balena.models.image.get) ⇒ <code>Bluebird</code>
             * [.getLogs(id)](#balena.models.image.getLogs) ⇒ <code>Promise</code>
         * [.billing](#balena.models.billing) : <code>object</code>
-            * [.getAccount()](#balena.models.billing.getAccount) ⇒ <code>Bluebird</code>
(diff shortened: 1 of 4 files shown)


thgreasi:
Since w/o the `if`s the UI would break when upgrading the SDK, I would like to stick with semver semantics.
I will drop all the `if`s right after and push a commit in the `v15` which sounds like it's going to happen quite soon.

DOCUMENTATION.md
```diff
@@ -6045,11 +6050,16 @@ balena.models.billing.getAccount(function(error, billingAccount) {
 ```
 <a name="balena.models.billing.getPlan"></a>
 
-##### billing.getPlan() ⇒ <code>Bluebird</code>
+##### billing.getPlan([organization]) ⇒ <code>Bluebird</code>
```
sradevski, line 6053: Why is it in `[]`?

thgreasi: Indeed that's how jsdoc marks optional arguments.
This is also autogenerated and we only specify `@param {(String|Number)} [organization]` in the comments.
PS: You can see the same pattern being used in `application.getAll([options])`.

sradevski: I did a quick search and I didn't find one, but I guess my searching skills failed me :)

thgreasi:
@balena-ci rebase

Issue #923, opened 2020-06-28
Should merge DeviceWithImageInstalls with the base Device interface


thgreasi:

thgreasi:
Resolved by #951

PR #951, opened 2020-07-14
Merge the DeviceWithImageInstalls with the Device type

thgreasi:
In order to properly match the return type of the `...WithServiceDetails` method (which are meant to provide an easier to consume result) we were so far using a separate type.
This preserves all core model properties on the Device and only augments it with with the easier to consume ones.
This way we can stop having some explicit casting, when trying to assign the result of a `...WithServiceDetails` method to a plain Device variable.

Connects-to: #923
Change-type: minor

---
##### Contributor checklist

- [x] Includes tests
- [x] Includes typings

### DIFF
--- a/lib/util/device-service-details.ts
+++ b/lib/util/device-service-details.ts
@@ -2,7 +2,6 @@ import type {
 	CurrentGatewayDownload,
 	CurrentService,
 	Device,
-	DeviceWithImageInstalls,
 	DeviceWithServiceDetails,
 	GatewayDownload,
 	Image,
@@ -14,7 +13,7 @@ import type {
 
 // Pine expand options necessary for getting raw service data for a device
 export const getCurrentServiceDetailsPineExpand = (expandRelease: boolean) => {
-	const pineExpand: PineExpand<DeviceWithImageInstalls> = {
+	const pineExpand: PineExpand<Device> = {
 		image_install: {
 			$select: ['id', 'download_progress', 'status', 'install_date'],
 			$filter: {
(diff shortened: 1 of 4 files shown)


typings/balena-sdk/models.d.ts
```diff
@@ -307,6 +307,8 @@ declare module './index' {
 		device_tag: ReverseNavigationResource<DeviceTag>;
 		manages__device: ReverseNavigationResource<Device>;
 		service_install: ReverseNavigationResource<ServiceInstall>;
+		image_install: ReverseNavigationResource<ImageInstall>;
+		gateway_download: ReverseNavigationResource<GatewayDownload>;
```
Page-, line 311: I think these should be
```suggestion
		image_install?: ReverseNavigationResource<ImageInstall>;
		gateway_download?: ReverseNavigationResource<GatewayDownload>;
```
to be compatible with the previous definition, and then in the places that explicit need these props you can use the `AtLeast` helper to type that they're required

lib/util/device-service-details.ts
```diff
@@ -117,11 +119,7 @@ export const generateCurrentServiceDetails = (
 		getSingleInstallSummary(gd),
 	);
 
-	// prefer over omit for performance reasons
-	delete rawDevice.image_install;
-	delete rawDevice.gateway_download;
-
-	const device = (rawDevice as Device) as DeviceWithServiceDetails;
+	const device = rawDevice as DeviceWithServiceDetails;
```
Page-, line 122: I'd move the assignment of this var to the bottom and have
```ts
	const device = rawDevice as DeviceWithServiceDetails;
	device.current_services = currentServicesGroupedByName;
	device.current_gateway_downloads = downloads;
	return device;
```
in order to reduce the amount of code it exists for and make it clearer that it's just a reassignment in order to add a couple of properties

thgreasi: :+1:

PR #955, opened 2020-07-21
Fix code snippet for initializing balena-sdk

vipulgupta2048:
Resolves: https://github.com/balena-io/docs/issues/1475
Change-type: patch

@Ereski :eyes:

### DIFF
--- a/DOCUMENTATION.md
+++ b/DOCUMENTATION.md
@@ -37,7 +37,11 @@ The module exports a single factory function.
 **Summary**: Creates a new SDK instance using the default or the provided options.  
 **Example**  
 ```js
+// with es6 imports
 import { getSdk } from 'balena-sdk';
+// or with node require
+const { getSdk } = require('balena-sdk');
+
 const balena = getSdk({
 	apiUrl: "https://api.balena-cloud.com/",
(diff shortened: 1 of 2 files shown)


DOCUMENTATION.md
```diff
@@ -37,10 +37,10 @@ The module exports a single factory function.
 **Summary**: Creates a new SDK instance using the default or the provided options.  
 **Example**  
 ```js
-import { getSdk } from 'balena-sdk';
+const {getSdk} = require('balena-sdk');
```
thgreasi, line 40: The sdk since v15 is now an es module and should be imported as one.
Maybe we can keep the current style as well as the old require style to make it easier for node consumers.
```suggestion
// with es6 imports
import { getSdk } from 'balena-sdk';
// or with node require
const { getSdk } = require('balena-sdk');
```

vipulgupta2048: Keeping backward compatibility is a great idea @thgreasi 
Also, have we documented the part regarding building the docs. Do you mind if we pen down a CONTRIBUTING of some sort for documentation changes? Thanks!

vipulgupta2048:
@thgreasi Good to go :rocket:

PR #958, opened 2020-07-27
device: add method to update target supervisor release

xginn8:
Connects-to: https://github.com/balena-io/balena-io/pull/2177
Change-type: patch

Resolves: # 
HQ: <url> 
See: <url> 
Depends-on: <url> 

---
##### Contributor checklist

- [ ] Includes tests
- [ ] Includes typings

### DIFF
--- a/lib/models/device.js
+++ b/lib/models/device.js
@@ -2712,6 +2712,82 @@ const getDeviceModel = function (deps, opts) {
 		},
 
 		/**
+		 * @summary Set a specific device to run a particular supervisor release
+		 * @name setSupervisorRelease
+		 * @public
+		 * @function
+		 * @memberof balena.models.device
+		 *
+		 * @description Configures the device to run a particular supervisor release.
+		 *
+		 * @param {String|Number} uuidOrId - device uuid (string) or id (number)
+		 * @param {String|Number} supervisorVersionOrId - the version of a released supervisor (string) or id (number)
+		 * @returns {Promise}
+		 *
+		 * @example
+		 * balena.models.device.setSupervisorRelease('7cf02a6', 'v10.8.0').then(function() {
+		 * 	...
+		 * });
+		 *
+		 * @example
+		 * balena.models.device.setSupervisorRelease(123, 'v11.4.14').then(function() {
+		 * 	...
+		 * });
+		 *
+		 * @example
+		 * balena.models.device.setSupervisorRelease('7cf02a6', 123, function(error) {
+		 * 	if (error) throw error;
+		 * 	...
+		 * });
+		 */
+		setSupervisorRelease: async (uuidOrId, supervisorVersionOrId) => {
+			let deviceId;
+			let releaseId;
+			if (isId(uuidOrId) && isId(supervisorVersionOrId)) {
+				deviceId = uuidOrId;
+				releaseId = supervisorVersionOrId;
+			} else {
+				const releaseFilterProperty = isId(supervisorVersionOrId)
+					? 'id'
+					: 'supervisor_version';
+				const device = await exports.get(uuidOrId, {
+					$select: 'id',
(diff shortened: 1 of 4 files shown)


xginn8:
@balena-ci rebase

lib/models/device.js
```diff
@@ -2711,6 +2711,79 @@ const getDeviceModel = function (deps, opts) {
...
+			let releaseId;
+			if (isId(uuidOrId) && isId(supervisorVersionOrId)) {
+				deviceId = uuidOrId;
+				releaseId = supervisorVersionOrId;
+			} else {
+				const releaseFilterProperty = isId(supervisorVersionOrId)
+					? 'id'
+					: 'supervisor_version';
+				const device = await exports.get(uuidOrId, {
+					$expand: { is_of__device_type: { $select: 'slug' } },
```
thgreasi, line 2754:
```suggestion
					$select: 'id',
					$expand: { is_of__device_type: { $select: 'slug' } },
```

lib/models/device.js
```diff
@@ -2711,6 +2711,79 @@ const getDeviceModel = function (deps, opts) {
...
+				});
+				const supervisorRelease = await pine.get({
```
thgreasi, line 2756: NIT: How about doing:
```suggestion
				const [supervisorRelease] = await pine.get({
```
so that we can just check for `== null` instead of `.length` ?

lib/models/device.js
```diff
@@ -2711,6 +2711,79 @@ const getDeviceModel = function (deps, opts) {
...
+					resource: 'supervisor_release',
+					options: {
```
thgreasi, line 2758:
```suggestion
					options: {
						$top: 1,
						$select: 'id',
```

lib/models/device.js
xginn8: Good idea, updated.

lib/models/device.js
xginn8: Good catch

lib/models/device.js
xginn8: Yep, that sounds better. Updated.

lib/models/device.js
```diff
@@ -2711,6 +2711,79 @@ const getDeviceModel = function (deps, opts) {
...
+		 * @summary Set a specific device to run a particular supervisor release
+		 * @name setSupervisorRelease
+		 * @public
+		 * @function
+		 * @memberof balena.models.device
+		 *
+		 * @description Configures the device to run a particular supervisor release.
+		 *
+		 * @param {String|Number} uuidOrId - device uuid (string) or id (number)
+		 * @param {String|Number} supervisorVersionOrId - the version of a released supervisor (string) or id (number)
```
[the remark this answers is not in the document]
xginn8: This is an interesting case. `null` is kind of special due to how legacy devices provision. I _think_ that's the only time it makes sense to allow setting to `null`, otherwise we will wind up allowing implicit downgrades like so: `version B` -> `null` -> `version A` such that B > A (see https://github.com/balena-io/open-balena-api/commit/68120045cf898cb068d361256ad65a30372658a6). Given that, I think the answer is no, we don't want to allow that, but can't tighten up the rule at the moment.

cc @CameronDiver

tests/integration/models/device.spec.js
```diff
@@ -2604,6 +2604,57 @@ describe('Device Model', function () {
 						});
 				});
 			});
+
+			describe.only('balena.models.device.setSupervisorRelease()', function () {
```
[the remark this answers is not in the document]
xginn8: removed!

lib/models/device.js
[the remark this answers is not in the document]
xginn8: Actually, I wonder if we should specifically disable that codepath at the API level as well...

xginn8:
@balena-ci retest

PR #961, opened 2020-08-03
Javascript to typescript conversions

Page-:
Resolves: # 
HQ: <url> 
See: <url> 
Depends-on: <url> 

---
##### Contributor checklist

- [ ] Includes tests
- [ ] Includes typings


typings/balena-sdk/models.d.ts
```diff
@@ -258,7 +258,7 @@ export interface Device {
 	download_progress?: number;
 	ip_address: string | null;
 	mac_address: string | null;
-	is_accessible_by_support_until__date: string;
+	is_accessible_by_support_until__date: string | number | null;
```
Page-, line 261: This is changed because the sdk send both `number` and `null` at different points - you can check where by reverting this change and running tsc

Page-: I changed how this is done in the end

lib/util/global-env.ts
```diff
@@ -1,5 +1,11 @@
+import { SdkOptions } from '../../typings/balena-sdk';
+
 // If we can't guarantee global state, don't fake it: fail instead.
-let globalEnv: typeof window | typeof self | typeof global;
+let globalEnv: (typeof window | typeof self | typeof global) & {
+	BALENA_SDK_SHARED_OPTIONS?: SdkOptions;
+	BALENA_SDK_HAS_USED_SHARED_OPTIONS?: boolean;
+	BALENA_SDK_HAS_SET_SHARED_OPTIONS?: boolean;
```
thgreasi, line 7: TIL that this is the way it works.

typings/balena-sdk/index.d.ts
```diff
@@ -664,26 +665,26 @@ export interface BalenaSDK {
...
+				options: { force?: boolean },
 			): Promise<void>;
 			getSupervisorState(uuidOrId: string | number): Promise<SupervisorStatus>;
 			getSupervisorTargetState(
 				uuidOrId: string | number,
 			): Promise<DeviceState.DeviceState>;
-			getDisplayName(deviceTypeName: string): Promise<string>;
-			getDeviceSlug(deviceTypeName: string): Promise<string>;
+			getDisplayName(deviceTypeName: string): Promise<string | undefined>;
+			getDeviceSlug(deviceTypeName: string): Promise<string | undefined>;
```
thgreasi, line 684:
```suggestion
			getDisplayName(deviceTypeName: string): Promise<string>;
			getDeviceSlug(deviceTypeName: string): Promise<string>;
```

lib/models/device.ts
```diff
@@ -1695,9 +1801,11 @@ const getDeviceModel = function (deps, opts) {
 		 * 	// Raspberry Pi
 		 * });
 		 */
-		getDisplayName: async (deviceTypeSlug) => {
+		getDisplayName: async (
+			deviceTypeName: string,
+		): Promise<string | undefined> => {
```
thgreasi, line 1806: :grey_exclamation: `getManifestBySlug` throws instead of returning `undefined.
```suggestion
		): Promise<string> => {
```

lib/models/device.ts
```diff
@@ -1734,7 +1842,9 @@ const getDeviceModel = function (deps, opts) {
 		 * 	// raspberry-pi
 		 * });
 		 */
-		getDeviceSlug: async (deviceTypeName) => {
+		getDeviceSlug: async (
+			deviceTypeName: string,
+		): Promise<string | undefined> => {
```
thgreasi, line 1847: :grey_exclamation: `getManifestBySlug` throws instead of returning `undefined.
```suggestion
		): Promise<string> => {
```

Page-: @thgreasi in the catch:
```ts
				if (error instanceof errors.BalenaInvalidDeviceType) {
					return;
				}
¬``

lib/models/device.ts
Page-: @thgreasi in the catch:
```ts
				if (error instanceof errors.BalenaInvalidDeviceType) {
					return;
				}

lib/models/device.ts
thgreasi: Indeed you are right, my bad.

PR #962, opened 2020-08-07
Add methods to enable, verify & disable 2fa

thgreasi:
Ideally we should also have a virtual authenticator and have test cases for that, but didn't take the time to do that now.
There is also the edge case where the tests stop right after 2fa is enabled, after which point we would probably not be able to re-run the test with the same user.

Change-type: minor

Resolves: # 
HQ: <url> 
See: <url> 
Depends-on: <url> 

---
##### Contributor checklist

- [ ] Includes tests
- [ ] Includes typings


thgreasi:
@balena-ci I self-certify!

PR #969, opened 2020-08-12
Update balena-pine and add custom typings for prepare/subscribe

Page-:
Update balena-pine from 12.2.0 to 12.3.0

Change-type: minor

Resolves: # 
HQ: <url> 
See: <url> 
Depends-on: <url> 

---
##### Contributor checklist

- [ ] Includes tests
- [ ] Includes typings


typings/pinejs-client-core.d.ts
```diff
@@ -297,3 +309,182 @@ export interface UpsertParams<T>
 	resource: string;
 	body: SubmitBody<T>;
 }
+
+export interface Dictionary<T> {
```
thgreasi, line 313: You can also import this from typings/utils.d.ts

PR #971, opened 2020-08-20
Update TypeScript to 4.0

thgreasi:
Change-type: patch

---
##### Contributor checklist

- [ ] Includes tests
- [ ] Includes typings


lib/models/release.ts
```diff
@@ -224,7 +224,6 @@ const getReleaseModel = function (
 			.sort((a, b) => a.service_name.localeCompare(b.service_name));
 
 		release.user = (release.is_created_by__user as BalenaSdk.User[])[0];
-		delete release.is_created_by__user;
```
thgreasi, line 227: This actually made the typings wrong (since it should at least hold a pine deferred), especially in case the user passes some options with an `$expand` for the `is_created_by__user`

lib/util/device-service-details.ts
```diff
@@ -87,7 +87,15 @@ function getSingleInstallSummary(
...
-	const result = {
+	const result: (
+		| (CurrentService &
+				Partial<
+					Pick<ImageInstall, 'installs__image' | 'is_provided_by__release'>
+				>)
+		| CurrentGatewayDownload
+	) &
+		Partial<Pick<ImageInstall, 'image'>> &
+		WithServiceName = {
```
thgreasi, line 98: All this is b/c now all props that we `delete` must be typed optional.
I don't like it and though of other approaches, but this was implemented this way to optimize this part of code for big fleets.

Issue #973, opened 2020-08-25
Cannot create an executable with pkg


catchtechnologies:
## Expected Behavior
Create an executable from a project using balena sdk with https://github.com/vercel/pkg.

## Actual Behavior
pkg throws the following error because it needs an absolute path to find ./es2015.  

```
Error: Cannot find module './es2015'
1) If you want to compile the package/file into executable, please pay attention to compilation warnings and specify a literal in 'require' call. 2) If you don't want to compile the package/file into executable and want to 'require' it from filesystem (likely plugin), specify an absolute path in 'require' call using process.cwd() or process.execPath.
``` 

## Steps to Reproduce the Problem

1. Add pkg to a project using balena sdk with `npm i pkg`
2. Create an executable with `pkg .`  
3. Run the executable with `./PROJECT_NAME`

## Specifications

- SDK version:
- Node version: 10.16.0
- Npm version: 6.13.7
- OS: OSX 10.15.6

## References
https://github.com/vercel/pkg#detecting-assets-in-source-code

thgreasi:
Hi @catchtechnologies and thanks for filling this detailed issue report.

This is a limitation in how `pkg` processes, where relative constructed path `require` calls do not result the sources being packed properly.
We do use such relative imports along with `@balena/es-version` as a way to enable various consumers to import the best matching build output for their runtime environment:
https://github.com/balena-io-modules/balena-es-version

This pkg limitation makes it a harder to fix this issue in the SDK side without breaking it for other consumers.

A way around it would be to try importing the desired version directly, with something like `require('balena-sdk/es2015')`.

The suggested way to handle this is to configure pkg so that it can properly bundle the requires sources for such imports. This is actually how the balena-cli managing to bundle the balena-sdk with pkg. Let me point you to the respective configuration:
https://github.com/balena-io/balena-cli/blob/fb879d3020e6e8a4bcd648b3b611780b0322c36e/package.json#L26-L47

You should be able to get this working with a pkg configuration like:
```
  "pkg": {
    "scripts": [
      "node_modules/balena-sdk/**/index.js"
    ],
    "assets": [
      "node_modules/balena-pine/**/*",
      "node_modules/pinejs-client-core/**/*"
    ]
  },
```

Let us know whether this works for you.

Kind regards,
Thodoris

catchtechnologies:
> A way around it would be to try importing the desired version directly, with something like `require('balena-sdk/es2015')`.
[...]

PR #976, opened 2020-08-26
Add support for bundling with pkg

thgreasi:
Connects-to: #973
Change-type: minor

---
##### Contributor checklist

- [ ] Includes tests
- [ ] Includes typings


thgreasi:
This will not work actually in this form, since pkg requires the second argument to be a string literal, which would require a lot of `if-else`, which will not be scalable.
See: https://github.com/vercel/pkg#detecting-assets-in-source-code

PR #979, opened 2020-08-26
Add dependabot configuration

thgreasi:
Change-type: patch
See: https://www.flowdock.com/app/rulemotion/d-build-pipelines/threads/mqmgXw4ASi4gKIs1hQdfJkH0OTy

---
##### Contributor checklist

- [ ] Includes tests
- [ ] Includes typings

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -1,0 +1,19 @@
+version: 2
+updates:
+- package-ecosystem: npm
+  directory: "/"
+  schedule:
+    interval: "daily"
+    time: "12:00"
+  open-pull-requests-limit: 1
+  ignore:
+  # Ignore entries in repo.yml until nested changelogs no longer require
+  # us to write the magic "update from to" commit message.
+  - dependency-name: "balena-auth"
+  - dependency-name: "balena-hup-action-utils"
+  - dependency-name: "balena-register-device"
+  - dependency-name: "balena-request"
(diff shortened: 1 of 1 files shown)


.github/dependabot.yml
```diff
@@ -0,0 +1,23 @@
...
+# https://help.github.com/github/administering-a-repository/configuration-options-for-dependency-updates
+
+# The ignored dependencies are so that we get open PRs that result proper nested changelogs
+
+version: 2
+updates:
+  - package-ecosystem: npm
+    directory: "/"
+    schedule:
+      interval: "daily"
```
sradevski, line 13: Isn't daily too frequent? We might get too much noise if it is daily, weekly might be better. Also, is dependabot something we decided to use company-wide or just on the sdk?

thgreasi: We are using it in JF and Lucian pointed me to their config, which I then used as basis for this. We have also extended VB to detect the change type from the title prefix, which this config instructs dependabot to add. So I was just thinking whether we could have this a try in the SDK. Still haven't decided though.

sradevski: @thgreasi yeah we can give it a go I guess, but I still think daily might be too much :) I'll leave it up to you to decide, and feel free to self-approve.

thgreasi:
@balena-ci I self-certify!

PR #995, opened 2020-09-10
Login: add new error handling, update balena-errors

JSReds:
Update balena-errors version to v4.5.1 and add new errors during
authntication

See: https://github.com/balena-io/balena-ui/pull/4032#discussion_r484693120
Change-type: patch

---
##### Contributor checklist

- [ ] Includes tests
- [ ] Includes typings

### DIFF
--- a/lib/auth.ts
+++ b/lib/auth.ts
@@ -175,17 +175,27 @@ const getAuth = function (
 		email: string;
 		password: string;
 	}): Promise<string> {
-		const { body } = await request.send<string>({
-			method: 'POST',
-			baseUrl: apiUrl,
-			url: '/login_',
-			body: {
-				username: credentials.email,
-				password: String(credentials.password),
-			},
-			sendToken: false,
-		});
-		return body;
+		try {
+			const { body } = await request.send<string>({
+				method: 'POST',
+				baseUrl: apiUrl,
+				url: '/login_',
+				body: {
+					username: credentials.email,
+					password: String(credentials.password),
+				},
+				sendToken: false,
+			});
+			return body;
+		} catch (err) {
+			if (err.statusCode === 401) {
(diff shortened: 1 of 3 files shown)


lib/auth.ts
```diff
@@ -175,17 +175,24 @@ const getAuth = function (
...
+					password: String(credentials.password),
+				},
+				sendToken: false,
+			});
+			return body;
+		} catch (err) {
+			if (err.statusCode === 429) {
+				throw new errors.BalenaTooManyRequests();
+			}
+			throw new errors.BalenaInvalidLoginCredentials();
```
thgreasi, line 194: I think that this should be checking for a 401, to not silence different errros.

thgreasi: Can we also get a test case for these plz?

sradevski (approved):

lib/auth.ts
```diff
@@ -175,17 +175,27 @@ const getAuth = function (
...
+			});
+			return body;
+		} catch (err) {
+			if (err.statusCode === 401) {
+				throw new errors.BalenaInvalidLoginCredentials();
+			}
+			if (err.statusCode === 429) {
+				throw new errors.BalenaTooManyRequests();
+			}
+			throw new errors.BalenaRequestError('Unauthorized', err.statusCode, {});
```
thgreasi, line 197: Imo we should re-throw the `err` since
* the server might have returned anything interesting, in which case we don't need to handle/conceal it
* what if `credentials` is undefined and it's just the js runtime that throws

PS: I think that when the `err` is from balena-request, it already a BalenaRequestError

thgreasi (changes requested):

thgreasi (approved):

PR #1014, opened 2020-10-22
Modify the os update to check against hostapp release

sradevski:
Change-type: minor

---
##### Contributor checklist

- [ ] Includes tests
- [ ] Includes typings

### DIFF
--- a/lib/models/device.ts
+++ b/lib/models/device.ts
@@ -122,8 +122,8 @@ const getDeviceModel = function (
 			opts,
 		),
 	);
-	const osModel = once(() =>
-		(require('./os') as typeof import('./os')).default(deps, opts),
+	const hostappModel = once(() =>
+		(require('./hostapp') as typeof import('./hostapp')).default(deps),
 	);
 
 	const {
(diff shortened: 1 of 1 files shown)


lib/models/device.ts
```diff
@@ -3234,11 +3188,17 @@ const getDeviceModel = function (
 			// this will throw an error if the action isn't available
 			exports._checkOsUpdateTarget(device, targetOsVersion);
 
-			const { versions: allVersions } = await osModel().getSupportedVersions(
-				device.is_of__device_type[0].slug,
-			);
-
-			if (!allVersions.some((v) => bSemver.compare(v, targetOsVersion) === 0)) {
+			const osVersions = (
+				await hostappModel().getAllOsVersions([
```
thgreasi, line 3192: Reminder that we also need the new hostAppModel to be used/offer something like `os.getSupportedOsUpdateVersions` (the CLI uses this).

thgreasi (approved):
Would love to see an ESR hup test case as well but I prefer to not block this.

PR #1024, opened 2020-12-02
device: Add deactivate method

mbalamat:
Add device deactivate method

### DIFF
--- a/tests/integration/models/device.spec.js
+++ b/tests/integration/models/device.spec.js
@@ -2079,6 +2079,21 @@ describe('Device Model', function () {
 					});
 				});
 			});
+
+			describe('balena.models.device.deactivate()', function () {
+				it('should be rejected if the device is online with statusCode 400', function () {
+					const promise = balena.models.device.deactivate(this.device.uuid);
+					return expect(promise).to.be.rejected.then(function (error) {
+						expect(error).to.have.property('statusCode', 400);
+						return m.chai
+							.expect(error)
+							.to.have.property(
+								'message',
+								'Request error: Devices must be offline in order to be deactivated.',
+							);
+					});
+				});
+			});
 		});
 
 		describe('given a device id whose shorter uuid is only numbers', function () {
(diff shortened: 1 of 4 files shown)


tests/integration/models/device.spec.js
```diff
@@ -1095,6 +1095,22 @@ describe('Device Model', function () {
 				});
 			});
 
+			describe('balena.models.device.deactivate()', function () {
```
sradevski, line 1098: Can you add a test case where it does succeed, just for completeness?

sradevski (commented):
You can also PR this in the dashboard after it's merged 😄

tests/integration/models/device.spec.js
mbalamat: @sradevski I thought the same thing though this involves billing so the user that runs the device spec tests should have billing info attached to them, which it doesn't. I asked [here](https://www.flowdock.com/app/rulemotion/r-beginners/threads/Dco-wiN7GyqHKk1_ZTvXBIpq_Ti).

sradevski: I see, I'll let Thodoris approve it in case he has something to add, otherwise LGTM. If it's urgent, feel free to self-certify and we can tackle this later.

thgreasi: @mbalamat let's move this last under the `given an application > given a single offline device > [contained scenario]`.
Then let's also add a deactivation test case for the device in that context that checks the status code and error message that the API gives back (probably something about not being able to deactivate devices in starter apps.
Hint to check the status code: `expect(error).to.have.property('statusCode', 401);`

Then lastly lets also add a test case for a failing deactivation under the `given an application > given an online device` that should give "401" & "Devices must be offline in order to be deactivated".

Even though these are just failing test cases, the current two do not test anything after the `device.get()`, so we are never checking whether the API knows the `is_active` field or not.

thgreasi (commented):
Other than the extra failing test cases LGTM :+1: 
Great first SDK PR. Nice digging.

thgreasi (approved):
Would prefer async await in the test but this is already great :+1:

Issue #1033, opened 2021-01-14
Limit supervisor upgrades to MC-compatible devices


xginn8:
As decided in a product thread, we want to limit the devices that can upgrade supervisors to only balenaOS >2.12 / supervisor >7. We'll need to do this from the UI as well.

see https://www.flowdock.com/app/rulemotion/r-product/threads/NLB1gISigptXnkF9fNmNWPVQrsR

xginn8:
this be done

PR #1034, opened 2021-01-18
supervisor-upgrades: limit to devices that support multicontainer

xginn8:
Connects-to: https://github.com/balena-io/balena-sdk/issues/1033
Change-type: patch

Resolves: # 
HQ: <url> 
See: <url> 
Depends-on: <url> 

---
##### Contributor checklist

- [ ] Includes tests
- [ ] Includes typings

### DIFF
--- a/lib/models/device.ts
+++ b/lib/models/device.ts
@@ -36,7 +36,6 @@ import * as url from 'url';
 
 import once = require('lodash/once');
 import * as bSemver from 'balena-semver';
-import * as semver from 'semver';
 import * as errors from 'balena-errors';
 
 import {
@@ -77,6 +76,7 @@ import { DeviceType } from '../../typings/balena-sdk/device-type-json';
 const MIN_SUPERVISOR_APPS_API = '1.8.0-alpha.0';
 
 const MIN_SUPERVISOR_MC_API = '7.0.0';
+const MIN_OS_MC = '2.12.0';
 
 // Degraded network, slow devices, compressed docker binaries and any combination of these factors
 // can cause proxied device requests to surpass the default timeout (currently 30s). This was
@@ -215,8 +215,8 @@ const getDeviceModel = function (
 	};
 
 	/**
-	 * @summary Ensure supervisor version compatibility using semver
-	 * @name ensureSupervisorCompatibility
+	 * @summary Ensure version compatibility using balena-semver
+	 * @name ensureVersionCompatibility
 	 * @private
 	 * @function
 	 *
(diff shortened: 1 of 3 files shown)


sradevski (approved):

tests/integration/models/device.spec.js
```diff
@@ -2725,6 +2728,25 @@ describe('Device Model', function () {
 						.expect(promise)
 						.to.be.rejectedWith(`Release not found: ${badRelease}`);
 				});
+
+				describe('given a device that does not support multicontainer', function () {
```
thgreasi, line 2732: Maybe we should wrap the old tests in a `given a device that does support multicontainer`, since both the `givenADevice(beforeEach` of that `describe()` & the `givenADevice(before` in here run for this test case.
What do you think @xginn8 ?

lib/models/device.ts
```diff
@@ -2998,6 +3005,25 @@ const getDeviceModel = function (
...
+			const device = (await exports.get(
+				uuidOrId,
+				deviceOpts,
+			)) as PineTypedResult<Device, typeof deviceOpts>;
+			ensureVersionCompatibility(
+				device.supervisor_version,
+				MIN_SUPERVISOR_MC_API,
+				'supervisor',
+			);
+			ensureVersionCompatibility(device.os_version, MIN_OS_MC, 'host OS');
```
thgreasi, line 3026: Does the API do any such check as well?
I'm not against this, since it gives a better error message, but having this in a hook of the API might be better.

tests/integration/models/device.spec.js
```diff
@@ -2679,6 +2679,9 @@ describe('Device Model', function () {
 			});
 
 			describe('balena.models.device.setSupervisorRelease()', function () {
+				givenADevice(beforeEach, {
+					os_version: 'balenaOS 2.56.1+rev1',
+				});
```
thgreasi, line 2684: NIT: How about just patching the device's os_version instead of creating a new one (on top of the `given a newly registered offline device` adds in its own `beforeEach`?
(Same below)

xginn8: sure, i avoided that as there may be side effects we are bypassing by faking a HUP.

lib/models/device.ts
xginn8: the API does not enforce this at the moment, as it felt a little overkill for a small edge case. i can enforce it as part of a hook if you'd be more comfortable enforcing it in both places.

tests/integration/models/device.spec.js
thgreasi: I'm fine if you prefer to do it like this (a POST and not a PATCH), I'm only trying to save the duplicate device creation in before each test case.
Feel free to move the whole `describe('balena.models.device.setSupervisorRelease()', function () {` outside the `'given a newly registered offline device'` and just have the `beforeEach` exactly as you prefer it :)

tests/integration/models/device.spec.js
xginn8: done, good idea.

tests/integration/models/device.spec.js
xginn8: done, thanks for the rec!

lib/models/device.ts
xginn8: thinking through this, we can't enforce this at the API level since devices that are being HUPed still require supervisor updates. we could potentially limit it to only device API keys, but i'm sure if that complexity is worth it.

thgreasi (commented):
I probably failed to describe this properly.
But I pushed two commits on this branch.
Let me know what you think, squash the typo commit and self-certify.

thgreasi (approved):

xginn8:
@balena-ci retest

thgreasi:
@balena-ci retest

PR #1042, opened 2021-01-26
tests: require an email appendage to avoid accidentally deleting acco…

xginn8:
…unts

Change-type: patch

Resolves: # 
HQ: <url> 
See: <url> 
Depends-on: <url> 

---
##### Contributor checklist

- [ ] Includes tests
- [ ] Includes typings

### DIFF
--- a/tests/integration/setup.ts
+++ b/tests/integration/setup.ts
@@ -68,6 +68,16 @@ const buildCredentials = function () {
 			username: env.TEST_REGISTER_USERNAME,
 		},
 	};
+	if (
+		// TODO: this should include the paid account eventually as well
+		![creds.email, creds.register.email].every(
+			(email) => email == null || email.includes('+testsdk'),
+		)
+	) {
+		throw new Error(
+			'Missing environment credentials, all emails must include `+testsdk` to avoid accidental deletion',
+		);
+	}
 
 	if (
 		!_.every([
--- a/README.md
+++ b/README.md
@@ -112,14 +112,15 @@ Users are encouraged to regularly update the balena SDK to the latest version.
 
 ## Tests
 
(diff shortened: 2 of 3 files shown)


xginn8:
this is a bit belt+suspenders, and i'm open to other approaches. we had a small incident in staging where i accidentally deleted my account, apps, and some shared apps i had access to (ref. https://www.flowdock.com/app/rulemotion/e-cloud-staging/threads/8uu60XY6PGfMeepodSaRta6As89).

thgreasi:
Seems that someone will also have to go to appveyor and update those env vars....

thgreasi:
Let me do that...

tests/integration/setup.ts
```diff
@@ -68,6 +68,19 @@ const buildCredentials = function () {
...
 		},
 	};
+	if (
+		![creds.email, creds.paid.email, creds.register.email].every((email) => {
+			if (email) {
+				return email.includes('+testsdk');
+			} else {
+				return true;
+			}
+		})
```
thgreasi, line 78:
```suggestion
		![creds.email, creds.paid.email, creds.register.email].every((email) => email == null || email.includes('+testsdk'))
```

Also, can we leave the paid user tests out? (I would need to reach to the customer success for this, which would take time, and we don't do any write operations for that user.)

xginn8: how about i leave it out for now, but we start the process of updating that as well? even though we do no writing now, that may change in the future and unless we enforce it as an invariant i worry about future data loss.

thgreasi: Sure

xginn8: thank you for all your help :heart:

thgreasi (approved):
:+1: just squash all commits

PR #1050, opened 2021-02-03
Add application membership methods

JSReds:
add application membership methods

Change-type: patch

---
##### Contributor checklist

- [ ] Includes tests
- [ ] Includes typings

### DIFF
--- a/tests/integration/models/application-membership.spec.ts
+++ b/tests/integration/models/application-membership.spec.ts
@@ -1,0 +1,365 @@
+// tslint:disable-next-line:import-blacklist
+import * as _ from 'lodash';
+import * as m from 'mochainon';
+import * as parallel from 'mocha.parallel';
+import {
+	balena,
+	credentials,
+	givenAnApplication,
+	givenLoggedInUser,
+} from '../setup';
+import type * as BalenaSdk from '../../..';
+const { expect } = m.chai;
+import { assertDeepMatchAndLength, timeSuite } from '../../util';
+
+describe('Application Membership Model', function () {
+	timeSuite(before);
+	givenLoggedInUser(before);
+	givenAnApplication(before);
+
+	let ctx: Mocha.Context;
+	before(async function () {
+		ctx = this;
+		const roles = await balena.pine.get({
+			resource: 'application_membership_role',
+			options: { $select: ['id', 'name'] },
+		});
+		this.applicationRoleMap = _.keyBy(roles, 'name');
+		roles.forEach((role) => {
+			expect(role).to.be.an('object');
+			expect(role).to.have.property('id').that.is.a('number');
+		});
+		this.applicationDeveloperRole = this.applicationRoleMap['developer'];
+		this.applicationOperatorRole = this.applicationRoleMap['operator'];
+		this.applicationObserverRole = this.applicationRoleMap['observer'];
+		[
+			this.applicationDeveloperRole,
+			this.applicationOperatorRole,
+			this.applicationObserverRole,
+		].forEach((role) => {
+			expect(role).to.be.an('object');
+		});
+	});
+
+	describe('balena.models.application.membership.create()', function () {
+		before(function () {
+			ctx = this;
+		});
+
+		parallel('[read operations]', function () {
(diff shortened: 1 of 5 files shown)


tests/integration/models/application-membership.spec.ts
```diff
@@ -0,0 +1,371 @@
...
+		[
+			this.applicationDeveloperRole,
+			this.applicationOperatorRole,
+			this.applicationObserverRole,
+		].forEach((role) => {
+			expect(role).to.be.an('object');
+		});
+	});
+
+	describe('given a new application', function () {
```
thgreasi, line 43: Let's remove this, since we the `givenAnApplication()` is one level up, so essentially all tests, just just those in here, have a that given app.

tests/integration/models/application-membership.spec.ts
```diff
@@ -0,0 +1,371 @@
...
+		describe('balena.models.application.membership.create()', function () {
+			before(function () {
+				ctx = this;
+			});
+
+			describe('[read operations]', function () {
```
thgreasi, line 49: You only need the `ctx = this` when using `parallel`, which I think we can use here right?
```suggestion
				ctx = this;
			});

			parallel('[read operations]', function () {
```

tests/integration/models/application-membership.spec.ts
```diff
@@ -0,0 +1,371 @@
...
+						await expect(promise).to.be.rejected.and.eventually.have.property(
+							'code',
+							'BalenaApplicationNotFound',
+						);
+					});
+				});
+
+				it(`should not be able to add a new member to a starter application with a given role operator`, async function () {
+					const promise = balena.models.application.membership.create({
+						application: this.application.id,
```
thgreasi, line 87: I guess that this one was failing and as a result you couldn't use `parallel` right?
```suggestion
						application: ctx.application.id,
```

tests/integration/models/application-membership.spec.ts
```diff
@@ -0,0 +1,371 @@
...
+
+					await expect(promise).to.be.rejectedWith(
+						'Application Membership not found',
+					);
+				});
+			});
+		});
+	});
+
+	describe('balena.models.application.membership.getAll()', function () {
```
thgreasi, line 258: I guess that this could also be `parallel` but not a blocker. Eg:
```
decribe('given a membership',
  before(...)
  after(...)
  parallel('balena.models.application.membership.getAll()', function () {
    let ctx
    before(function(){ ctx = this; });
```


I guess you can just move this to the `describe('given a membership [read operations]'` right below so that we also re-use the same membership for the tests.

thgreasi (approved):
:+1: 
Just NITs.
Feel free to merge but would be nice to see those improved before merging.

tests/integration/models/application-membership.spec.ts
JSReds: I actually started without using parallel saying "I'll add it later" and then I forgot it. Thank you for pointing it out :)

tests/integration/models/application-membership.spec.ts
JSReds: actually much better! thank you !

JSReds:
@balena-ci rebase

thgreasi:
@balena-ci rebase

tests/integration/models/application-membership.spec.ts
```diff
@@ -0,0 +1,365 @@
...
+	balena,
+	credentials,
+	givenAnApplication,
+	givenLoggedInUser,
+} from '../setup';
+import type * as BalenaSdk from '../../..';
+const { expect } = m.chai;
+import { assertDeepMatchAndLength, timeSuite } from '../../util';
+
+describe.only('Application Membership Model', function () {
```
thgreasi, line 15: `.only` :police_car:

JSReds: :prray:

thgreasi:
@JSReds I think this v PR will help you get your tests passing
See: https://github.com/balena-io/balena-sdk/pull/1052

thgreasi:
@balena-ci rebase

PR #1057, opened 2021-02-15
Support unique keys in organization & application membership.changeRole() & remove()

thgreasi:
Change-type: minor

---
##### Contributor checklist

- [ ] Includes tests
- [ ] Includes typings

### DIFF
--- a/tests/integration/models/application-membership.spec.ts
+++ b/tests/integration/models/application-membership.spec.ts
@@ -12,6 +12,19 @@ import type * as BalenaSdk from '../../..';
 const { expect } = m.chai;
 import { assertDeepMatchAndLength, timeSuite } from '../../util';
 
+const keyAlternatives = [
+	['id', (member: BalenaSdk.ApplicationMembership) => member.id],
+	[
+		'alternate key',
+		(member: BalenaSdk.ApplicationMembership) =>
+			_.mapValues(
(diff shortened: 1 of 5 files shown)


sradevski (commented):
@thgreasi what is the reason for us adding this? Code wise seems fine, but it is quite confusing what `uniqueKey` is and it ends up with a strange logic to support it.

thgreasi:
> it ends up with a strange logic to support it.

To which logic you refer? B/c supporting this didn't add any new code in the implementation, it only added tests.
Actually users could already do that if they didn't use/care about TS errors, b/c this PR is simply re-exposing what pine client already supports.

Imo it's more convenient than having to find the membership id and it also allows us to drop one extra custom method call from the UI.

tests/integration/models/application-membership.spec.ts
```diff
@@ -12,6 +12,19 @@ import type * as BalenaSdk from '../../..';
 const { expect } = m.chai;
 import { assertDeepMatchAndLength, timeSuite } from '../../util';
 
+const keyAlternatives = [
```
sradevski, line 15: I was referring to this one, I didn't notice that it is in the test 🤦‍♂️

sradevski (approved):

PR #1059, opened 2021-02-24
Add OSVersion archived property

JSReds:
Add OSVersion archived property

Change-type: minor

---
##### Contributor checklist

- [ ] Includes tests
- [ ] Includes typings

### DIFF
--- a/lib/models/hostapp.ts
+++ b/lib/models/hostapp.ts
@@ -33,7 +33,7 @@ export interface OsVersion {
 	osType: string;
 	line?: OsLines;
 	variant?: string;
-
+	archived?: boolean;
 	formattedVersion: string;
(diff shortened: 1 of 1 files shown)


lib/models/hostapp.ts
```diff
@@ -33,7 +33,7 @@ export interface OsVersion {
 	osType: string;
 	line?: OsLines;
 	variant?: string;
-
+	archived?: boolean;
```
sradevski, line 36: It's not the version that is archived, but the entire hostapp, so it doesn't make sense to put it on the `Release`, so to speak.

JSReds: got your point @sradevski, I'm closing this PR to do another solutions

PR #1062, opened 2021-03-08
Add getLatestOsVersions in hostapp model

JSReds:
Add getLatestOsVersions method in hostapp model

Change-type: minor

---
##### Contributor checklist

- [ ] Includes tests
- [ ] Includes typings

### DIFF
--- a/lib/models/hostapp.ts
+++ b/lib/models/hostapp.ts
@@ -9,6 +9,7 @@ import type {
 } from '../types/models';
 import { Dictionary } from '../../typings/utils';
 import { getAuthDependentMemoize } from '../util/cache';
+import { maxSatisfying } from 'semver';
 
 const RELEASE_POLICY_TAG_NAME = 'release-policy';
 const ESR_NEXT_TAG_NAME = 'esr-next';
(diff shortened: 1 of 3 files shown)


tests/integration/models/hostapp.spec.ts
```diff
@@ -75,4 +75,22 @@ describe('Hostapp model', function () {
 			expect(firstRes).to.equal(secondRes);
 		});
 	});
+
+	parallel('balena.models.hostapp.getAllOsVersions()', function () {
```
thgreasi, line 79:
```suggestion
	parallel('balena.models.hostapp.getLatestOsVersions()', function () {
```

lib/models/hostapp.ts
```diff
@@ -227,6 +234,14 @@ const getHostappModel = function (deps: InjectedDependenciesParam) {
 							is_invalidated: false,
 						},
 					},
+					should_be_running__release: {
```
thgreasi, line 237: How about creating a separate object, so that we don't retrieve all the data in both methods?
I might be missing something, but it looks like `should_be_running__release` isn't used anywhere in this file ?!?!

JSReds: yes sorry we can remove it, I was testing if it was possible to reuse the same method that handle the `owns__release` array  to also handle `should_be_running__release`and get the latest versions from there. But as far as I can see it's not exactly reliable

PR #1064, opened 2021-03-10
Added contract field to device type model

mehalter:
Change-type: minor
Signed-off-by: Micah Halter <micah@balena.io>

---
##### Contributor checklist

- [ ] Includes tests
- [ ] Includes typings

### DIFF
--- a/lib/types/contract.ts
+++ b/lib/types/contract.ts
@@ -1,0 +1,17 @@
+export interface Contract {
+	slug: string;
+	type: string;
+	name?: string;
+	version?: string;
+	externalVersion?: string;
+	contractVersion?: string;
+	description?: string;
+	aliases?: string[];
+	tags?: string[];
+	data?: object;
+	assets?: object;
+	requires?: string[];
+	provides?: string[];
+	composedOf?: object;
+	partials?: string[];
+}
--- a/lib/types/models.ts
+++ b/lib/types/models.ts
@@ -1,5 +1,6 @@
 import type { DeviceOverallStatus } from './device-overall-status';
 export type { DeviceOverallStatus } from './device-overall-status';
+import { Contract } from './contract';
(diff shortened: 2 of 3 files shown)


mehalter:
I think we should be able to move the `contract` field on `Releases` to a `Contract` type as well, but I'm not sure if that would break anything at the moment or where it's used to go check

lib/types/contract.ts
```diff
@@ -0,0 +1,17 @@
+export interface Contract {
```
thgreasi, line 1: Is this of the same format like the one we accept in the releases & the hostApps?
If not, then it might better be names like `DeviceTypeContract`.
Other than that LGTM (would be nice to have more accurate types than just `object` though). Feel free to self-certify.

lib/types/models.ts
```diff
@@ -380,6 +381,7 @@ export interface DeviceType {
 	name: string;
 	is_private: boolean;
 	logo: string | null;
+	contract: Contract | null;
```
thgreasi, line 384: I also just realized that the types of the image.contract & release.contract are `string | null`, which I guess are wrong...
Would be nice to fix them in a separate patch release.

lib/types/contract.ts
mehalter: This is the same format for all contracts based on the current contract specification. This is also why some of the types are objects

lib/types/models.ts
mehalter: Yeah, I noticed this. I'll fix these in a separate patch as well

lib/types/contract.ts
thgreasi: I'm not sure why this explains the why some props are `object`s, but we definitely don't need to block this PR any more.

thgreasi (approved):

PR #1070, opened 2021-03-30
Fix type of contract fields that should be any instead of object

mehalter:
Change-type: patch
Signed-off-by: Micah Halter <micah@balena.io>

Resolves: # 
HQ: <url> 
See: <url> 
Depends-on: <url> 

---
##### Contributor checklist

- [ ] Includes tests
- [ ] Includes typings

### DIFF
--- a/lib/types/contract.ts
+++ b/lib/types/contract.ts
@@ -1,3 +1,5 @@
+import { AnyObject } from '../../typings/utils';
+
 export interface Contract {
 	slug: string;
 	type: string;
@@ -8,10 +10,10 @@ export interface Contract {
 	description?: string;
 	aliases?: string[];
 	tags?: string[];
-	data?: object;
-	assets?: object;
+	data?: AnyObject;
+	assets?: AnyObject;
 	requires?: string[];
 	provides?: string[];
-	composedOf?: object;
-	partials?: string[];
+	composedOf?: AnyObject;
+	partials?: AnyObject;
 }


lib/types/contract.ts
```diff
@@ -8,10 +8,10 @@ export interface Contract {
...
 	aliases?: string[];
 	tags?: string[];
-	data?: object;
-	assets?: object;
+	data?: any;
+	assets?: any;
 	requires?: string[];
 	provides?: string[];
-	composedOf?: object;
+	composedOf?: any;
```
thgreasi, line 15: Would love to see an example of these.
Any chance the following would work?
```suggestion
	composedOf?: Dictionary<any>; // or AnyObject, both from `~/typings/utils.d.ts`
```
This is not great either, but if all you know is that this is an object, then it's better than `any`.

mehalter: For something like `data` and `assets` they are free form objects with structure that changes based on the type of contract that it is (i.e. hardware device type, software application, etc.) like the [fincm3 contract data field](https://github.com/balena-io/contracts/blob/cc2ca1ad8627e66f53c97a4ab07b928ede6da3ed/contracts/hw.device-type/fincm3/contract.json#L13) vs. the [balenaSound contract data field](https://github.com/balenalabs/balena-sound/blob/0f90bf0cf3819535b34c9ee2e24c3e805761b9ef/balena.yml#L16)

For `composedOf` I need to do some digging and see why this is an object in the specification.

mehalter: I did some local testing and `anyObject` is the best choice. Fixed.

thgreasi (approved):

Issue #1016, opened 2020-11-02
Should support querying applications by UUID


pdcastro:
The balena API has support for querying applications by UUID:
https://github.com/balena-io/open-balena-api/pull/491/files

The SDK should support this too. As for the model interface / function signatures, given that UUIDs and app names are both strings that cannot be readily unambiguously distinguished, the SDK should provide a way for the caller to explicitly specify whether the identifier is a UUID or application, for example:

```
application.get({ uuid: 'decade2' }, opts)
```

Some additional discussion took place in a [private thread](https://www.flowdock.com/app/rulemotion/i-cli/threads/g26bVg0Nc518ZZknqxtX0GmWhwb). Extract:

> Any application name with exactly 32 or 62 characters in the set `[a-fA-F0-9]` would be indistinguishable from an application UUID (while not necessarily being a UUID, or at least not a UUID in balena's database, perhaps a UUID in the user's own database), and thus subject to a query error. It's a rare corner case, but we are better off avoiding them. It would be an unnecessary restriction to tell users that they shouldn't name applications with 32 or 62 characters from a particular character set. I like `application.get({ uuid: '123456' }, opts)` better than trial-and-error heuristic probabilistic auto detection. :-)

> There is also the case of short UUIDs. For device UUIDs, I gather that the SDK currently issues a filtered query with `$startswith: uuidOrId`:  
> https://github.com/balena-io/balena-sdk/blob/v15.17.0/lib/models/device.ts#L514
> For devices, the SDK only needs to do this when the UUID is not a number (JS data type), so at least it knows that the short UUID is actually a UUID. With applications, 7-char UUIDs would more easily be confused with names. For example, the word "decade2" is both a valid app name and valid short UUID. I am OK with the occasional conflict and error if the user _chooses_ to use an ambiguous command line option/syntax like `--application decade2` (in fact in this thread I have voiced my preference to keep allowing this kind of auto detection for backwards compatibility even in future major versions of the CLI, when others suggested that the next major version of the CLI should disallow auto detection / ambiguous identifier specification). But I also think that the SDK and the CLI should provide a way of unambiguously specifying whether the identifier is a name, UUID or database ID. So while the SDK could provide the feature of best effort filtering by long UUID, short UUID, app name, slug and database ID, I think there should also be a way of querying only by a specific identifier type like UUID. Like `application.get({ uuid: 'decade2' }, opts)`.

[...]

Issue #1030, opened 2021-01-05
application.get should not support app names


thgreasi:
`application.get()` should not support searching by application name, so that it never returns ambiguous results.
Consumers that care about this use case should be using `application.getAll()`.

Discussed in product call 2021-01-05 as part of the CLI changes with regards to the release of orgs.
See: https://jel.ly.fish/brainstorm-call-product-5jan2021-d6f4871d-c9d9-4deb-82ca-7da2c25034a6

pdcastro:
During the product call, it was suggested that the SDK should not offer `application.get(appName)` because the app name on its own is ambiguous. However, I think that an alternative that would keep with the spirit of the suggestion is for the SDK to require upfront disambiguation in the form of an option that specified the "kind" of application that the caller would like to get: "direct access" (an app that belongs to an org that belongs to the user), "public app" or some other kind. If the app "kind" was not specified, the SDK would throw an error such as "Invalid arguments" (indicating a bug in the caller code, rather than user input ambiguity).

thgreasi:
@pdcastro I think that the point was to just make all app related calls unambiguous, but just teaching the users how to use app slug. This wouldn't just solve the public app issue, but also the ambiguous errors that cli users get when using an app name that is used by more than one directly accessible apps (in different orgs). If we go this way, then it would also simplify the CLI code as well, since it will no longer have to follow any  "ambiguous error => show a list of apps to select" flow.

That's a good reminder though to invite you to the respective call I'm planning about direct pine usage in the CLI & UI.

PR #1074, opened 2021-04-07
application.get: Add support for retrieving applications by uuid

thgreasi:
Originally draft just b/c I still haven't added tests.

Resolves: #1016
Change-type: minor

---
##### Contributor checklist

- [ ] Includes tests
- [ ] Includes typings

### DIFF
--- a/lib/models/application.ts
+++ b/lib/models/application.ts
@@ -271,7 +271,7 @@ const getApplicationModel = function (
 		 * @function
 		 * @memberof balena.models.application
 		 *
-		 * @param {String|Number} slugOrId - application slug (string) or id (number)
+		 * @param {String|Number} slugOrUuidOrId - application slug (string), uuid (string) or id (number)
 		 * @param {Object} [options={}] - extra pine options to use
 		 * @param {String} [context] - extra access filters, undefined or 'directly_accessible'
 		 * @fulfil {Object} - application
@@ -279,6 +279,11 @@ const getApplicationModel = function (
 		 *
 		 * @example
 		 * balena.models.application.get('myorganization/myapp').then(function(application) {
+		 * 	console.log(application);
+		 * });
+		 *
+		 * @example
+		 * balena.models.application.get('1bf99a68cf9e4266986e6dec7a6e8f46').then(function(application) {
 		 * 	console.log(application);
 		 * });
 		 *
@@ -294,17 +299,11 @@ const getApplicationModel = function (
 		 * });
 		 */
 		async get(
-			slugOrId: string | number,
+			slugOrUuidOrId: string | number,
 			options?: PineOptions<Application>,
 			context?: 'directly_accessible',
 		): Promise<Application> {
-			if (options == null) {
-				options = {};
-			}
-
-			if (slugOrId == null) {
-				throw new errors.BalenaApplicationNotFound(slugOrId);
-			}
+			options ??= {};
 
 			const accessFilter =
 				context === 'directly_accessible'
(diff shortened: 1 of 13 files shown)


lib/models/application.ts
```diff
@@ -359,20 +375,21 @@ const getApplicationModel = function (
 						{
 							$filter: {
 								$or: {
-									app_name: nameOrSlugOrId,
-									slug: nameOrSlugOrId.toLowerCase(),
+									app_name: nameOrSlugOrUuidOrId,
+									slug: nameOrSlugOrUuidOrId.toLowerCase(),
+									uuid: nameOrSlugOrUuidOrId.toLowerCase(),
```
thgreasi, line 380: This is essentially the only code change.

pdcastro: To confirm, this only accepts "long" or "full" UUID, right? Not something like `decade2` which is a valid 7-char short UUID. (I think that accepting short UUIDs here would be bad, creating more ambiguity that goes in the opposite direction of the spirit of #1030.) By the way, we discussed before (not necessarily decided, at least discussed) that there should be a way for the SDK to allow the caller (CLI/UI/user app) to query specifically and unambiguously by UUID or slug or name or ID, instead of `nameOrSlugOrUuidOrId`. For example, instead of:
```
    nameOrSlugOrUuidOrId: string | number
```
Perhaps it could be:
```
    nameOrSlugOrUuidOrId: string | number | object
```
And if it was an object, the object would be exactly one of:
* `{ id: number }`
* `{ uuid: string }`
* `{ name: string }`
* `{ slug: string }`

But, fair enough, a long / full UUID can be assumed to never legitimately clash with an unintended resource. I say "legitimately" because a bad user could copy and paste the UUID of an app and set it as the name of another app, in which case this query would be "illegitimately ambiguous" when `nameOrSlugOrUuidOrId` was a string.

pdcastro: Or maybe not that harmless... A bit far fetched and just hypothetical, but could user A create a sort of denial of service attack on user B by creating an app that was _named_ after the UUID of one of user B's apps? User A could become aware of the UUID of an app of user B for being a collaborator / observer of that app.

Or, what about public apps? If the UUID of a public app is public, could I create a DoS attack on the public app by _naming_ another public app (of mine?) after that UUID? Such that people's query for the legit UUID would return my illegit app (by name) instead. I haven't tested, just hypothetical.

thgreasi: I did have in mind the `| AlternateKeyObject`, but decided to not implement it in this PR, since I wouldn't want to also support names in that notation, since technicaly the name is not an alternate key (aka table wide natural key).

The denial of service scenario you describe is already possible by having user B name their app with the same name as user A. But user A can in that case use the app slug. This is going to be fully resolved though with #1030.

thgreasi: @pdcastro let me know if you would prefer to not merge this before the next major where we drop support for retrieving apps by name in this method.

pdcastro: > The denial of service scenario you describe is already possible by having user B name their app with the same name as user A. But user A can in that case use the app slug.

Is it really the same thing? The case of clashing names (multiple apps with the same name) can be resolved by using the app slug which the recommended practice, but for the scenario I described, querying by app UUID does not solve the problem -- and to say "use the slug instead of the UUID" sounds like a design failure. It sounds like we are saying that querying by app UUID could be ambiguous, an ambiguity being introduced by this PR / solution.

> if you would prefer to not merge this before the next major

I'm not worried about that, but I am worried about ambiguity when querying by UUID. I'd rather this PR implemented a solution where querying by UUID could not possibly be ambiguous, or subject to DoS.

thgreasi:
@balena-ci I self-certify!

PR #855, opened 2020-03-24
Add a device.getProgress convenience method

thgreasi:
The tests fail atm since the device.overall_progress returns a string atm, but we plan to fix that.

Resolves: #790
See: https://www.flowdock.com/app/rulemotion/resin-tech/threads/vFYF1_0T0MhlpvvnZHyko5BQRhK
Depends-on: https://github.com/balena-io/balena-api/issues/2515
Change-type: minor
Signed-off-by: Thodoris Greasidis <thodoris@balena.io>

---
##### Contributor checklist

- [ ] Includes tests
- [ ] Includes typings
- [ ] Includes updated documentation
- [ ] Includes updated build output

PR #1077, opened 2021-04-10
Add email verification & email request methods

nitishagar:
Connects-to: #855
Change-type: minor
HQ: https://github.com/balena-io/balena-io/pull/954
See: https://www.flowdock.com/app/rulemotion/resin-tech/threads/tNwuZRmXyhNdgKJZLq6-fckbOvC
Depends-on: https://github.com/balena-io/balena-api/pull/3134

---
##### Contributor checklist

- [ ] Includes tests
- [x] Includes typings

### DIFF
--- a/DOCUMENTATION.md
+++ b/DOCUMENTATION.md
@@ -363,6 +363,8 @@ const sdk = fromSharedOptions();
         * [.getEmail()](#balena.auth.getEmail) ⇒ <code>Promise</code>
         * [.logout()](#balena.auth.logout) ⇒ <code>Promise</code>
         * [.register(credentials)](#balena.auth.register) ⇒ <code>Promise</code>
+        * [.verifyEmail(verificationPayload)](#balena.auth.verifyEmail) ⇒ <code>Promise</code>
+        * [.requestVerificationEmail()](#balena.auth.requestVerificationEmail) ⇒ <code>Promise</code>
     * [.logs](#balena.logs) : <code>object</code>
         * [.subscribe(uuidOrId, [options])](#balena.logs.subscribe) ⇒ <code>Promise.&lt;LogSubscription&gt;</code>
         * [.history(uuidOrId, [options])](#balena.logs.history) ⇒ <code>Promise</code>
@@ -7272,6 +7274,8 @@ balena.models.billing.downloadInvoice(orgId, '0000').then(function(stream) {
     * [.getEmail()](#balena.auth.getEmail) ⇒ <code>Promise</code>
     * [.logout()](#balena.auth.logout) ⇒ <code>Promise</code>
     * [.register(credentials)](#balena.auth.register) ⇒ <code>Promise</code>
+    * [.verifyEmail(verificationPayload)](#balena.auth.verifyEmail) ⇒ <code>Promise</code>
+    * [.requestVerificationEmail()](#balena.auth.requestVerificationEmail) ⇒ <code>Promise</code>
 
 <a name="balena.auth.twoFactor"></a>
 
@@ -7741,6 +7745,43 @@ balena.auth.register({
 	if (error) throw error;
 	console.log(token);
 });
+```
+<a name="balena.auth.verifyEmail"></a>
+
+#### auth.verifyEmail(verificationPayload) ⇒ <code>Promise</code>
+**Kind**: static method of [<code>auth</code>](#balena.auth)  
+**Summary**: Verifies an email  
+**Access**: public  
+**Fulfil**: <code>String</code> - session token  
+
+| Param | Type | Description |
+| --- | --- | --- |
+| verificationPayload | <code>Object</code> | in the form of email, and token |
+| verificationPayload.email | <code>String</code> | the email |
+| verificationPayload.token | <code>String</code> | the verification token |
+
+**Example**  
+```js
+balena.auth.verifyEmail({
+	email: 'johndoe@gmail.com',
+	token: '5bb11d90eefb34a70318f06a43ef063f'
+}).then(function(jwt) {
+	console.log(jwt);
+});
+```
+<a name="balena.auth.requestVerificationEmail"></a>
+
+#### auth.requestVerificationEmail() ⇒ <code>Promise</code>
+This will only work if you used [login](#balena.auth.login) to log in.
+
+**Kind**: static method of [<code>auth</code>](#balena.auth)  
+**Summary**: Re-send verification email to the user  
+**Access**: public  
+**Example**  
+```js
+balena.auth.requestVerificationEmail().then(function() {
+	console.log('Requesting verification email operation complete!');
+})
 ```
 <a name="balena.logs"></a>
 
--- a/lib/auth.ts
+++ b/lib/auth.ts
@@ -469,6 +469,7 @@ const getAuth = function (
 	 * 	if (error) throw error;
 	 * 	console.log(token);
 	 * });
+	 *
 	 */
 	async function register(credentials: {
 		email: string;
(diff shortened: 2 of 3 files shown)


lib/auth.ts
```diff
@@ -450,7 +450,6 @@ const getAuth = function (
 	 * @param {String} credentials.password - the password
 	 * @param {(String|undefined)} [credentials.'g-recaptcha-response'] - the captcha response
 	 *
-	 * @fulfil {String} - session token
```
thgreasi, line 453: How about:
```
* @fulfil {String|void} - session token or undefined in case that email verification is required.
```

lib/auth.ts
```diff
@@ -462,26 +461,72 @@ const getAuth = function (
 	 * });
 	 *
 	 * @example
+	 * // If email verification is enabled
```
thgreasi, line 465: How about having a single register call and inside the `then` have an `if` statement that does a `console.log('registrered)` or `console.log('Email verification is required')`?

thgreasi (commented):
LGTM (Approved) but (reminder that) we need to wait for the API PR to get merged first, before merging this.

JSReds:
@thgreasi @nitishagar Could we also update the User model with the properties added in the API model? IMHO It might be useful to have those properties. 

Referring to 
```
is_verified: boolean;
must_be_verified: boolean;
unverified_email: string | null;
...
```

thgreasi:
@JSReds I think those are not whitelisted, so customers will not be able to use/know about these propeties.

JSReds:
Could we whitelist them, I'm just asking since these might be useful data to make a correct redirect in case of an email verification UI

lib/auth.ts
```diff
@@ -450,41 +450,120 @@ const getAuth = function (
...
+	 * balena.auth.requestVerificationEmail().then(function() {
+	 * 	console.log('Requesting verification email operation complete!');
+	 * }).catch(function(error) {
+	 * 	if (error) throw error;
+	 * 	console.log(email);
+	 * });
+	 *
+	 */
+	async function requestVerificationEmail() {
+		const { id } = await getUserDetails();
```
thgreasi, line 557: Let's use the public method:
```suggestion
		const id = await getUserId();
```

lib/auth.ts
```diff
@@ -450,41 +450,120 @@ const getAuth = function (
...
+	 * @example
+	 * balena.auth.requestVerificationEmail().then(function() {
+	 * 	console.log('Requesting verification email operation complete!');
+	 * }).catch(function(error) {
+	 * 	if (error) throw error;
+	 * 	console.log(email);
+	 * });
+	 *
+	 */
+	async function requestVerificationEmail() {
```
thgreasi, line 556: How about v ?
```suggestion
	async function resendVerificationEmail() {
```
I see a point though of this naming for pre-existing users that later try to verify their account, since "resend" doesn't make much sense if you have never been sent one before...

nitishagar: Confused here - do you prefer the "resend" over "request"? From your comment doesn't look like it; but you change shows me that suggested.

lib/auth.ts
```diff
@@ -450,41 +450,120 @@ const getAuth = function (
...
 			url: '/user/register',
 			baseUrl: apiUrl,
 			body: credentials,
 			sendToken: false,
 		});
+
+		// With email verification enabled we return with status(201) i.e. no auto-login
+		if (body === 'Created') {
+			return;
+		}
```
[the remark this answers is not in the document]
nitishagar: Yes - post our discussion, I have updated sign-up to have a similar flow to log-in.

lib/auth.ts
```diff
@@ -457,18 +457,12 @@ const getAuth = function (
 	 * balena.auth.register({
 	 * 	email: 'johndoe@gmail.com',
 	 * 	password: 'secret'
-	 * }).then(function(token) {
-	 * 	console.log(token);
+	 * }).then(function(result) {
+	 * 	console.log(`Token: ${result}`);
+	 * }).catch(function(error) {
+	 * 	throw error;
```
dfunckt, line 463: I don't see a reason for these changes. What is it you mean to convey to the user with the addition of the catch handler that merely rethrows the error?

lib/auth.ts
```diff
@@ -482,9 +476,84 @@ const getAuth = function (
...
+	 *
+	 * @fulfil {string} - returns with 'OK' status when successful
+	 * @returns {Promise}
+	 *
+	 * @example
+	 * balena.auth.requestVerificationEmail().then(function() {
+	 * 	console.log('Requesting verification email operation complete!');
+	 * }).catch(function(error) {
+	 * 	if (error) throw error;
+	 * });
```
dfunckt, line 543: Same comment here ^

Also, I'm either misreading the diff in the generated .md file above or maybe you forgot to regenerate that file after changes to this comment?

dfunckt (commented):
Just a couple of comments related to doc-strings. Otherwise LGTM.

lib/auth.ts
nitishagar: The reason I added them first was to cover both branches. But makes sense to not show the err' if we just show an example of throwing it again. Updated the PR.

lib/auth.ts
nitishagar: Replied Above. And updated with required changes here.

DOCUMENTATION.md
```diff
@@ -362,6 +362,8 @@ const sdk = fromSharedOptions();
         * [.getEmail()](#balena.auth.getEmail) ⇒ <code>Promise</code>
```
thgreasi, line 362: Let's title the commit like
> Add email verification & email resend methods

lib/auth.ts
```diff
@@ -482,9 +474,82 @@ const getAuth = function (
...
+		return body;
+	}
+
+	/**
+	 * @summary Re-send verification email to the user
+	 * @name requestVerificationEmail
+	 * @public
+	 * @function
+	 * @memberof balena.auth
+	 * @description This will only work if you used {@link balena.auth.login} to log in.
```
thgreasi, line 531:
```suggestion
```

lib/auth.ts
thgreasi: I'm in favor of my code change suggestion. The rest were just thoughts.

lib/auth.ts
```diff
@@ -457,18 +457,10 @@ const getAuth = function (
 	 * balena.auth.register({
 	 * 	email: 'johndoe@gmail.com',
 	 * 	password: 'secret'
-	 * }).then(function(token) {
-	 * 	console.log(token);
-	 * });
+	 * }).then(function(result) {
+	 * 	console.log(`Token: ${result}`);
+	 * })
```
thgreasi, line 462: Why did we remove the second/callback-based example?
Let's revert the example changes here to keep the PR clean, since at the end we didn't change that method at all.
```suggestion
	 * }).then(function(token) {
	 * 	console.log(token);
	 * });
	 *
	 * @example
	 * balena.auth.register({
	 * 	email: 'johndoe@gmail.com',
	 * 	password: 'secret'
	 * }, function(error, token) {
	 * 	if (error) throw error;
	 * 	console.log(token);
	 * });
```
Feel free to open a separate PR though if you think this is anyway improving the examples.

lib/auth.ts
nitishagar: Any reason we want to remove this? This would be only used after `login` and we need the following to work as well: https://github.com/balena-io/balena-sdk/pull/1077/files#diff-cd41b427b38a554f9a1a2d2e848e41ddc39bc7d68dba54c1398d7384d6aaf4f7R543

lib/auth.ts
nitishagar: As you also mentioned lets keep the `requestVerificationEmail()` name given we are using this for pre-existing users as well.

lib/auth.ts
thgreasi: I mean that I expect this to work even if you have logged in with an named user API key, which is possible for pre-existing users.
Not a big deal though I guess.

lib/auth.ts
```diff
@@ -482,9 +483,82 @@ const getAuth = function (
...
+
+	/**
+	 * @summary Re-send verification email to the user
+	 * @name requestVerificationEmail
+	 * @public
+	 * @function
+	 * @memberof balena.auth
+	 * @description This will only work if you used {@link balena.auth.login} to log in.
+	 *
+	 * @fulfil {string} - returns with 'OK' status when successful
```
thgreasi, line 542:
```suggestion
```
Let's drop this, since we might drop the textual return value in the API soon.

lib/auth.ts
```diff
@@ -482,9 +483,82 @@ const getAuth = function (
...
+	 *
+	 * @example
+	 * balena.auth.requestVerificationEmail().then(function() {
+	 * 	console.log('Requesting verification email operation complete!');
+	 * })
+	 *
+	 */
+	async function requestVerificationEmail() {
+		const id = await getUserId();
+		return await pine.patch({
```
thgreasi, line 553:
```suggestion
		await pine.patch({
```
Let's not return at all, since we might drop the textual return value in the API soon.

thgreasi (commented):
Other than those LGTM

thgreasi (approved):

PR #1079, opened 2021-04-12
**BREAKING** Change apiKey.getAll() to return all keys w/o any filter & add apiKey.getAllNamedUserApiKeys()

thgreasi:
The apiKeys.getAllNamedUserApiKeys will be offering
the same behavior.

Change-type: major
See: https://jel.ly.fish/287934d0-6fb2-4af2-888a-47966b9393f0
See: https://www.flowdock.com/app/rulemotion/resin-frontend/threads/aLBhe-Vfl7EBsb-oNp71Mm_TTUv

---
##### Contributor checklist

- [ ] Includes tests
- [ ] Includes typings

### DIFF
--- a/lib/models/api-key.ts
+++ b/lib/models/api-key.ts
@@ -24,7 +24,7 @@ const getApiKeysModel = function (
 	deps: InjectedDependenciesParam,
 	opts: InjectedOptionsParam,
 ) {
-	const { pine, request } = deps;
+	const { pine, request, sdkInstance } = deps;
 	const { apiUrl } = opts;
 	const exports = {
 		/**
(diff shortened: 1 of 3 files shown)


lib/models/api-key.ts
```diff
@@ -110,10 +110,47 @@ const getApiKeysModel = function (
...
+		 */
+		async getAllNamedUserApiKeys(
+			options: BalenaSdk.PineOptions<BalenaSdk.ApiKey> = {},
+		): Promise<BalenaSdk.ApiKey[]> {
+			return await pine.get({
+				resource: 'api_key',
+				options: mergePineOptions(
+					{
 						$filter: {
+							is_of__actor: await sdkInstance.auth.getUserActorId(),
```
nitishagar, line 150: Do we want to do a check for valid API keys also? (Assuming we have a time check or something similar)

tests/integration/models/api-key.spec.ts
```diff
@@ -196,13 +245,17 @@ describe('API Key model', function () {
...
-				return expect(balena.models.apiKey.revoke(this.apiKey.id))
-					.to.not.be.rejected.then(() => balena.models.apiKey.getAll())
-					.then(function (apiKeys) {
-						expect(apiKeys).to.be.an('array');
-						expect(apiKeys).to.have.lengthOf(0);
-					});
+			it('should be able to revoke an exising api key', async function () {
+				await expect(balena.models.apiKey.revoke(this.apiKey.id)).to.not.be
+					.rejected;
+				const apiKeys = await balena.models.apiKey.getAll({
```
nitishagar, line 251: Move this to a common method? I see this repeated @ 48, 68 and here (251).

lib/models/api-key.ts
thgreasi: We atm don't have such time constrained validity periods, but even if we had I wouldn't embed such a check here.
I mean, this major change was required just b/c we had an embedded check, so the less the better.

tests/integration/models/api-key.spec.ts
thgreasi: Indeed... It's called `getAllNamedUserApiKeys()`...

JSReds (approved):
nice one!

PR #1091, opened 2021-05-14
Add public organization types

JSReds:
Add public organization types

Change-type: minori

---
##### Contributor checklist

- [ ] Includes tests
- [ ] Includes typings

### DIFF
--- a/lib/types/models.ts
+++ b/lib/types/models.ts
@@ -47,6 +47,7 @@ export interface ResourceTypeMap {
 	plan__has__discount_code: PlanDiscountCode;
 	plan_addon: PlanAddon;
 	plan_feature: PlanFeature;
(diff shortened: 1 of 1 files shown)


lib/types/models.ts
```diff
@@ -169,6 +170,7 @@ export interface Application {
 	is_archived: boolean;
 	is_discoverable: boolean;
 	is_stored_at__repository_url: string | null;
+	public_organization: NavigationResource<PublicOrganization>;
```
thgreasi, line 173: I think that this is actually an OptionalNavigationResource b/c a if the org that owns the app has no public app, then the expand here will return an empty array :thinking:

thgreasi (approved):

PR #1101, opened 2021-06-10
Add contract partial based instruction generation

mehalter:
This is an initial implementation of a basic API call to interpolate partials within a contract. Discussion is needed to address the HostOS contract that is currently hard-coded (should be pulled from somewhere else). Also, this functionality applies not just to device types, but any contract could potentially have partials that need to be interpolated. There currently isn't a place to handle such abstract functionality in the SDK (to my knowledge) and it seems like this would be a good place to use Contrato as it is going to be (should be) the de facto implementation of the Contract data structure and operations surrounding them. It might be better for the partial interpolation to live there and then anything that is contract based in the SDK have the ability to use Contrato types and functionality.

This will also be able to easily fix the current issue where the instructions are not correct in BalenaHub and open fleets since new devices added do not appear in the dashboard. This will allow different interfaces and uses of the partials to provide their own overarching template (Discussion for this specific issue on Flowdock: https://www.flowdock.com/app/rulemotion/balenahub/threads/UQ29ow3_IXlkXMOnzT-ea9H6V-q).
https://balena.zulipchat.com/#narrow/stream/349104-balena-io.2Fcontracts/topic/contract.20partial.20based.20instruction.20generation

### TODO

- [x] Updating the contracts repo to have necessary partials
- [x] Improve implementation to handle the few edge cases that don't fall under the usual Etcher based install methods
  - [x] Fix cases with nested instructions like having different instructions for linux/mac/windows

Depends-on: https://github.com/balena-io/contracts/pull/236
Change-type: minor

---
##### Contributor checklist

- [x] Includes tests
- [x] Includes typings

### DIFF
--- a/lib/models/balenaos-contract.ts
+++ b/lib/models/balenaos-contract.ts
@@ -1,0 +1,76 @@
+import { Contract } from '../types/contract';
+
+// Hardcoded host OS contract, this should be moved to the Yocto build process with meta-balena.
+// Here for initial implementatin and testing purposes
+const BalenaOS: Contract = {
+	name: 'balenaOS',
+	slug: 'balenaos',
+	type: 'sw.os',
+	description: 'Balena OS',
+	partials: {
+		image: [`{{#each deviceType.partials.instructions}}{{{this}}} {{/each}}`],
+		internalFlash: [
+			`{{#each deviceType.partials.connectDevice}}{{{this}}} {{/each}}`,
+			`Write the {{name}} file you downloaded to the {{deviceType.name}}. We recommend using <a href="http://www.etcher.io/">Etcher</a>.`,
+			`Wait for writing of {{name}} to complete.`,
+			`{{#each deviceType.partials.disconnectDevice}}{{{this}}} {{/each}}`,
+			`{{{deviceType.partials.bootDevice}}} to boot the device.`,
+		],
+		externalFlash: [
+			`Insert the {{deviceType.data.media.altBoot.[0]}} to the host machine.`,
+			`Write the {{name}} file you downloaded to the {{deviceType.data.media.altBoot.[0]}}. We recommend using <a href="http://www.etcher.io/">Etcher</a>.`,
+			`Wait for writing of {{name}} to complete.`,
+			`Remove the {{deviceType.data.media.altBoot.[0]}} from the host machine.`,
+			`Insert the freshly flashed {{deviceType.data.media.altBoot.[0]}} into the {{deviceType.name}}.`,
+			`<strong role="alert">Warning!</strong> This will also completely erase internal storage medium, so please make a backup first.`,
+			`{{#each deviceType.partials.bootDeviceExternal}}{{{this}}} {{/each}}`,
+			`Wait for the {{deviceType.name}} to finish flashing and shutdown. {{#if deviceType.partials.flashIndicator}}Please wait until {{deviceType.partials.flashIndicator}}.{{/if}}`,
+			`Remove the {{deviceType.data.media.altBoot.[0]}} from the {{deviceType.name}}.`,
+			`{{#each deviceType.partials.bootDeviceInternal}}{{{this}}} {{/each}}`,
+			`{{{deviceType.partials.bootDevice}}} to boot the device.`,
+		],
+		externalBoot: [
+			`Insert the {{deviceType.data.media.defaultBoot}} to the host machine.`,
+			`Write the {{name}} file you downloaded to the {{deviceType.data.media.defaultBoot}}. We recommend using <a href="http://www.etcher.io/">Etcher</a>.`,
+			`Wait for writing of {{name}} to complete.`,
+			`Remove the {{deviceType.data.media.defaultBoot}} from the host machine.`,
+			`Insert the freshly flashed {{deviceType.data.media.defaultBoot}} into the {{deviceType.name}}.`,
+			`{{{deviceType.partials.bootDevice}}} to boot the device.`,
+		],
+		jetsonFlash: [
+			`Put the device in recovery mode and connect to the host computer via USB`,
+			`{{#if deviceType.partials.jetsonNotes}}{{#each deviceType.partials.jetsonNotes}}{{{this}}} {{/each}}{{/if}}`,
+			`Unzip the {{name}} image and use the Jetson Flash tool to flash the {{deviceType.name}} found at <a href="https://github.com/balena-os/jetson-flash">https://github.com/balena-os/jetson-flash</a>.`,
+			`Wait for writing of {{name}} to complete.`,
+			`{{{deviceType.partials.bootDevice}}} to boot the device.`,
+		],
+		edisonFlash: {
+			Linux: [
+				`{{#each deviceType.partials.Linux.flashDependencies}}{{{this}}} {{/each}}`,
+				`Unplug the {{deviceType.name}} from your system`,
+				`Unzip the downloaded {{name}} file`,
+				`{{#each deviceType.partials.Linux.flashInstructions}}{{{this}}} {{/each}}`,
+				`Plug the {{deviceType.name}} as per the instructions on your terminal.`,
+				`You can check the progress of the provisioning on your terminal.`,
+			],
+			MacOS: [
+				`{{#each deviceType.partials.MacOS.flashDependencies}}{{{this}}} {{/each}}`,
+				`Unplug the {{deviceType.name}} from your system`,
+				`Unzip the downloaded {{name}} file`,
+				`{{#each deviceType.partials.MacOS.flashInstructions}}{{{this}}} {{/each}}`,
+				`Plug the {{deviceType.name}} as per the instructions on your terminal.`,
+				`You can check the progress of the provisioning on your terminal.`,
+			],
+			Windows: [
+				`{{#each deviceType.partials.Windows.flashDependencies}}{{{this}}} {{/each}}`,
+				`Unplug the {{deviceType.name}} from your system`,
+				`Unzip the downloaded {{name}} file`,
+				`{{#each deviceType.partials.Windows.flashInstructions}}{{{this}}} {{/each}}`,
+				`Plug the {{deviceType.name}} as per the instructions on your terminal.`,
+				`You can check the progress of the provisioning on your terminal.`,
+			],
+		},
+	},
+};
+
+export { BalenaOS };
--- a/lib/models/device-type.ts
+++ b/lib/models/device-type.ts
@@ -16,8 +16,75 @@ limitations under the License.
 
 import type { InjectedDependenciesParam, PineOptions } from '..';
 import { DeviceType } from '../types/models';
+import { Partials, Contract } from '../types/contract';
 import { mergePineOptions } from '../util';
 import * as errors from 'balena-errors';
+import * as Handlebars from 'handlebars';
+import cloneDeep = require('lodash/cloneDeep');
+
+// REPLACE ONCE HOST OS CONTRACTS ARE GENERATED THROUGH YOCTO
+import { BalenaOS } from './balenaos-contract';
+
+const traversingCompile = (
+	partials: Partials,
+	initial: Contract,
+	path: string[],
+): Contract => {
+	let interpolated: Contract = { ...initial }
+	for (const partialKey of Object.keys(partials)) {
+		const current = partials[partialKey];
+		if (Array.isArray(current)) {
+			let location: any = interpolated;
+			for (const key of path) {
+				location = location[key];
+			}
+			// if array of partials, compile the template
+			location[partialKey] = current
+				.map((partial) => Handlebars.compile(partial)(interpolated))
+				.filter((n) => n);
+		} else {
+			// if it's another dictionary, keep traversing
+			interpolated = traversingCompile(
+				current,
+				interpolated,
+				path.concat([partialKey]),
+			);
+		}
+	}
+	return interpolated;
+}
+
+const interpolatedPartials = (contract: Contract): Contract => {
+	if (contract.partials) {
+		return traversingCompile(contract.partials, contract, ['partials']);
+	} else {
+		return contract;
+	}
+};
+
+const calculateInstallMethod = (contract: Contract): string => {
+	const flashProtocol = contract.data?.flashProtocol;
+	const defaultBoot = contract.data?.media?.defaultBoot;
+	if (flashProtocol) {
+		if (flashProtocol === 'RPIBOOT') {
+			return 'internalFlash';
+		} else {
+			return flashProtocol;
+		}
+	} else if (defaultBoot) {
+		if (defaultBoot === 'image') {
+			return 'image';
+		} else if (defaultBoot === 'internal') {
+			return 'externalFlash';
+		} else {
+			return 'externalBoot';
+		}
+	} else {
+		throw new errors.BalenaError(
+			`Unable to determine installation method for contract: ${contract.slug}`,
+		);
+	}
+};
 
 const getDeviceTypeModel = function (deps: InjectedDependenciesParam) {
 	const { pine } = deps;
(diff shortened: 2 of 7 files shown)


mehalter:
@balena-ci rebase

mehalter:
@balena-ci rebase

mehalter:
Tested this iterating over the `getAllSupported` result and got all of the instructions:

```js
(await sdk.models.deviceType.getAllSupported()).forEach(async (x, i) => {
  var instructions = await sdk.models.deviceType.getInstructions(x.slug)
  console.log("Instructions for " + x.slug + ":\n");
  console.log(instructions);
});
```

There are a couple devices that get returned by `getAllSupported` that do not have contracts (and therefore no partials): `smartcube-kbox-a150` and `smartcube-kbox-a250`. I'm not sure where these are coming from and why they are being returned, but I also don't see them as an option in Balena Cloud. These also do correctly throw an error indicating that the instruction partials are not defined rather than failing in an unexpected way.

I have also attached the following text file with the complete output from running the above code snippet (with a bit of prettying up of the formatting for easier reading and checking: [all_instructions.txt](https://github.com/balena-io/balena-sdk/files/9888752/all_instructions.txt)

lib/models/balenaos-contract.ts
```diff
@@ -0,0 +1,76 @@
...
+			`{{{deviceType.partials.bootDevice}}} to boot the device.`,
+		],
+		externalBoot: [
+			`Insert the {{deviceType.data.media.defaultBoot}} to the host machine.`,
+			`Write the {{name}} file you downloaded to the {{deviceType.data.media.defaultBoot}}. We recommend using <a href="http://www.etcher.io/">Etcher</a>.`,
+			`Wait for writing of {{name}} to complete.`,
+			`Remove the {{deviceType.data.media.defaultBoot}} from the host machine.`,
+			`Insert the freshly flashed {{deviceType.data.media.defaultBoot}} into the {{deviceType.name}}.`,
+			`{{{deviceType.partials.bootDevice}}} to boot the device.`,
+		],
```
alexgg, line 39: @mehalter the partials above should go into the existing balenaOS contract in https://github.com/balena-io/contracts/blob/master/contracts/sw.os/balenaos/contract.json.

This is already used by the OS build scripts so when added the contract will be deployed with the OS release.

lib/models/balenaos-contract.ts
```diff
@@ -0,0 +1,76 @@
...
+				`Unplug the {{deviceType.name}} from your system`,
+				`Unzip the downloaded {{name}} file`,
+				`{{#each deviceType.partials.Windows.flashInstructions}}{{{this}}} {{/each}}`,
+				`Plug the {{deviceType.name}} as per the instructions on your terminal.`,
+				`You can check the progress of the provisioning on your terminal.`,
+			],
+		},
+	},
+};
+
```
alexgg, line 75: Both `jetsonFlash` and `edisonFlash` are device specific and they do not belong in the OS contract. They need to be included into something like `sw.device-family`, and then pulled in when the specific device type contract is built.

alexgg (commented):
A couple of comments so we can move the partials into the OS contracts - I'll leave the ts review to someone more versed in the SDK.

lib/models/balenaos-contract.ts
mehalter: Awesome, I'll get this moved over. Is this accessible through the API as well?

lib/models/balenaos-contract.ts
mehalter: Hm, each of these partials are supported installation methods of the operating system and are not necessarily device specific. Any device that supports using the jetson flash tool should be allowed to get these instructions. These are just Balena OS's way of saying, "I support being installed with the jetson flash tool, and this is how. It doesn't provide any information that is specific to a device type. If that makes sense. Basically it's the BalenaOS's part in the flash process of telling the user "how to handle the file that I am giving you.

lib/models/balenaos-contract.ts
alexgg: @mehalter it's deployed with the OS release. For example if you do:
```
await sdk.pine.get({
    resource: 'release',
    id: nnnnn,
});
```
You will get something like:
```
belongs_to__application: {__id: 1520936, __deferred: {…}}
build_log:null
commit:"2c3a3a6c020e3ca5a2ae008f9de59643"
composition:{version: '2.1', networks: {…}, volumes: {…}, services: {…}}
contract:
"{\"name\":\"Balena OS for Raspberry Pi 4 (using 64bit OS)\",\"type\":\"sw.block\",\"description\":\"Balena OS for a Raspberry Pi 4 (using 64bit OS)\",\"provides\":[{\"type\":\"sw.os\",\"slug\":\"balena-os\"},{\"type\":\"hw.device-type\",\"slug\":\"raspberrypi4-64\"}],\"composedOf\":[\"balena-os\",\"raspberrypi4-64\"],\"version\":\"2.105.1+rev1\"}"
```
As you see at the moment the contract is very basic, but once you add the instructions they will appear there.

lib/models/balenaos-contract.ts
alexgg: Jetson is a device family, no other vendor will use Jetson tools. And the same goes for edison.

Why would a contract for the RaspberryPi need to mention other vendors like Jetson or edison? In my view this corresponds to a `sw.device-family` contract that is then pulled in only when building the Jetson/Edison families.

lib/models/device-type.ts
```diff
@@ -16,8 +16,74 @@ limitations under the License.
...
+					keys.concat([partialKey]),
+				);
+			}
+			return interpolated;
+		},
+		{ ...initial },
+	);
+};
+
+const interpolatedPartials = (contract: Contract, initial: any = {}) => {
```
Page-, line 55: Can you type this better than `any` please

lib/models/device-type.ts
```diff
@@ -16,8 +16,74 @@ limitations under the License.
...
 import * as errors from 'balena-errors';
+import * as Handlebars from 'handlebars';
+import cloneDeep = require('lodash/cloneDeep');
+
+// REPLACE ONCE HOST OS CONTRACTS ARE GENERATED THROUGH YOCTO
+import { BalenaOS } from './balenaos-contract';
+
+const traversingCompile = (partials: any, initial: any, keys: string[]) => {
+	return Object.keys(partials).reduce(
+		(interpolated: any, partialKey) => {
```
Page-, line 30: Can you type this better than `any` please

lib/models/device-type.ts
```diff
@@ -323,6 +389,113 @@ const getDeviceTypeModel = function (deps: InjectedDependenciesParam) {
...
+		 * 	// bootDevice: ["Connect power to the Raspberry Pi (v1 / Zero / Zero W)"]
+		 * });
+		 */
+		getInterpolatedPartials: async (
+			deviceTypeSlug: string,
+			initial: any = {},
+		): Promise<Contract> => {
+			const contract = (
+				await exports.getBySlugOrName(deviceTypeSlug, { $select: 'contract' })
+			).contract;
```
Page-, line 419:
```suggestion
			const { contract } = (
				await exports.getBySlugOrName(deviceTypeSlug, { $select: 'contract' })
			);
```

lib/models/device-type.ts
```diff
@@ -323,6 +389,113 @@ const getDeviceTypeModel = function (deps: InjectedDependenciesParam) {
...
+		 *  // Insert the freshly flashed sdcard into the Raspberry Pi (v1 / Zero / Zero W).
+		 *  // Connect power to the Raspberry Pi (v1 / Zero / Zero W) to boot the device.
+		 * });
+		 */
+		getInstructions: async (
+			deviceTypeSlug: string,
+		): Promise<any | string[]> => {
+			const contract = (
+				await exports.getBySlugOrName(deviceTypeSlug, { $select: 'contract' })
+			).contract;
```
Page-, line 455:
```suggestion
			const { contract } = (
				await exports.getBySlugOrName(deviceTypeSlug, { $select: 'contract' })
			);
```

lib/models/device-type.ts
```diff
@@ -323,6 +389,113 @@ const getDeviceTypeModel = function (deps: InjectedDependenciesParam) {
...
+		 * 	console.log(method);
+		 *  // externalBoot
+		 * });
+		 */
+		getInstallMethod: async (
+			deviceTypeSlug: string,
+		): Promise<string | null> => {
+			const contract = (
+				await exports.getBySlugOrName(deviceTypeSlug, { $select: 'contract' })
+			).contract;
```
Page-, line 492:
```suggestion
			const { contract } = (
				await exports.getBySlugOrName(deviceTypeSlug, { $select: 'contract' })
			);
```

package.json
```diff
@@ -123,6 +123,7 @@
     "balena-semver": "^2.3.0",
     "balena-settings-client": "^4.0.6",
     "date-fns": "^2.29.3",
+    "handlebars": "^4.7.7",
```
Page-, line 126: Do you know how much size this adds to the sdk bundle?

lib/types/device-type-json.ts
```diff
@@ -1,8 +1,9 @@
+import type { Contract } from './contract';
 import type { Dictionary } from '../../typings/utils';
 
 /* types for the /device-types/v1 endppoints */
 
-export interface DeviceType {
+export interface DeviceType extends Contract {
 	slug: string;
```
Page-, line 7: A bunch of the properties here should now be inherited from `Contract`, in which case it'd make sense imo to remove the duplication here

package.json
mehalter: looks like it adds 0.2Mb

lib/models/device-type.ts
```diff
@@ -16,8 +16,74 @@ limitations under the License.
...
+import { Contract } from '../types/contract';
 import { mergePineOptions } from '../util';
 import * as errors from 'balena-errors';
+import * as Handlebars from 'handlebars';
+import cloneDeep = require('lodash/cloneDeep');
+
+// REPLACE ONCE HOST OS CONTRACTS ARE GENERATED THROUGH YOCTO
+import { BalenaOS } from './balenaos-contract';
+
+const traversingCompile = (partials: any, initial: any, keys: string[]) => {
```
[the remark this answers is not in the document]
mehalter: Resolved in latest push

lib/models/device-type.ts
mehalter: Resolved in latest push

lib/models/device-type.ts
mehalter: Resolved in latest push

lib/models/device-type.ts
mehalter: Resolved in latest push

lib/models/device-type.ts
mehalter: Resolved in latest push

lib/types/device-type-json.ts
mehalter: Resolved in latest push

lib/types/device-type-json.ts
```diff
@@ -1,12 +1,9 @@
...
 import type { Dictionary } from '../../typings/utils';
 
 /* types for the /device-types/v1 endppoints */
 
-export interface DeviceType {
-	slug: string;
-	name: string;
-	aliases: string[];
-
+export interface DeviceType extends Contract {
```
thgreasi, line 6: Since this is exactly matching the device-type.json types, and also requires us to use `?.` whenever accessing `name` from now on, let's completely revert the changes in this file.
I hope we would actually be able to completely drop this file at some point...
If for the contracts work you can declare a more accurate DeviceType Contract definition, then let's do it in the new `lib/types/contract.ts` file w/ a specialized `DeviceTypeContract` interface.

package.json
thgreasi: I wonder whether we could rely on `_.template`, but we can change that later if we need to investigate & drop the size increase.

lib/models/config.ts
```diff
@@ -80,18 +80,18 @@ const getConfigModel = function (
...
 			}
 			if (deviceType.state === 'PREVIEW') {
 				deviceType.state = 'ALPHA';
-				deviceType.name = deviceType.name.replace('(PREVIEW)', '(ALPHA)');
+				deviceType.name = deviceType.name?.replace('(PREVIEW)', '(ALPHA)');
 			}
 			if (deviceType.state === 'EXPERIMENTAL') {
 				deviceType.state = 'NEW';
-				deviceType.name = deviceType.name.replace('(EXPERIMENTAL)', '(NEW)');
+				deviceType.name = deviceType.name?.replace('(EXPERIMENTAL)', '(NEW)');
```
thgreasi, line 94: Reminder to revert the `?.` here once we revert the DeviceType interface change (from the device-type-json.ts).

lib/models/device-type.ts
```diff
@@ -16,8 +16,78 @@ limitations under the License.
...
+
+// REPLACE ONCE HOST OS CONTRACTS ARE GENERATED THROUGH YOCTO
+import { BalenaOS } from './balenaos-contract';
+
+const traversingCompile = (
+	partials: Partials,
+	initial: Contract,
+	keys: string[],
+): Contract => {
+	return Object.keys(partials).reduce(
```
thgreasi, line 33: Let's make this a `for...of`.

lib/models/device-type.ts
```diff
@@ -16,8 +16,78 @@ limitations under the License.
...
+		(interpolated: Contract, partialKey) => {
+			const current = partials[partialKey];
+			if (Array.isArray(current)) {
+				let location: any = interpolated;
+				for (const key of keys) {
+					location = location[key];
+				}
+				// if array of partials, compile the template
+				location[partialKey] = current
+					.map((partial: string) => Handlebars.compile(partial)(interpolated))
```
thgreasi, line 43: I would expect TS to be able to properly infer this
```suggestion
					.map((partial) => Handlebars.compile(partial)(interpolated))
```

lib/models/device-type.ts
```diff
@@ -16,8 +16,78 @@ limitations under the License.
...
+	return Object.keys(partials).reduce(
+		(interpolated: Contract, partialKey) => {
+			const current = partials[partialKey];
+			if (Array.isArray(current)) {
+				let location: any = interpolated;
+				for (const key of keys) {
+					location = location[key];
+				}
+				// if array of partials, compile the template
+				location[partialKey] = current
```
thgreasi, line 42: Should this be inside the `for (const key of keys)`?

lib/models/device-type.ts
```diff
@@ -16,8 +16,78 @@ limitations under the License.
...
+	keys: string[],
+): Contract => {
+	return Object.keys(partials).reduce(
+		(interpolated: Contract, partialKey) => {
+			const current = partials[partialKey];
+			if (Array.isArray(current)) {
+				let location: any = interpolated;
+				for (const key of keys) {
+					location = location[key];
+				}
```
thgreasi, line 40: Let's rename the `keys` parameter to `path` or something and add a comment here on what the loop achieves.
```suggestion
				let location: any = interpolated;
				for (const key of path) {
					location = location[key];
				}
```

lib/models/device-type.ts
```diff
@@ -16,8 +16,78 @@ limitations under the License.
...
+					keys.concat([partialKey]),
+				);
+			}
+			return interpolated;
+		},
+		{ ...initial },
+	);
+};
+
+const interpolatedPartials = (contract: Contract, initial?: any): Contract => {
```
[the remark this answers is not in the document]
vipulgupta2048: It seems to be getting used to interpolate contract first from device type and then hostOS. So I left it in.

lib/models/device-type.ts
```diff
@@ -323,6 +393,113 @@ const getDeviceTypeModel = function (deps: InjectedDependenciesParam) {
...
+				throw new Error(
+					`Instruction partials not defined for ${deviceTypeSlug}`,
+				);
+			}
+			const installMethod = calculateInstallMethod(contract);
+			const interpolatedDeviceType = interpolatedPartials(contract);
+			const interpolatedHostOS = interpolatedPartials(cloneDeep(BalenaOS), {
+				deviceType: interpolatedDeviceType,
+			});
+
```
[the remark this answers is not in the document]
vipulgupta2048: This didn't work out, I instead did owing to other changes I have made. I don't have any context on what this amounts, since I haven't been able to test anything. 

```suggestion
			const interpolatedHostOS = interpolatedPartials({...cloneDeep(BalenaOS), ...interpolatedDeviceType});
```

vipulgupta2048:
Let me know what you think of the changes made here @thgreasi if they are good to go from this point, then I will start writing tests. I also considered use loadash's template instead of handlebars and decided against it as I still haven't gotten to the point of figuring out what the output will produce.

package.json
vipulgupta2048: Not changing this and can be discussed later.

lib/models/device-type.ts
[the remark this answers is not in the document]
vipulgupta2048: This wasn't working, fixed this in my latest commit.

vipulgupta2048:
@thgreasi PR ready to go. The methods have been tested and I will open a PR in docs to show how this looks on production for the final review.

thgreasi (approved):

PR #1102, opened 2021-06-14
models/apiKey: Add getProvisioningApiKeysByApplication() method

nitishagar:
Change-type: minor

---
##### Contributor checklist
- [x] Includes tests
- [ ] Includes typings

### DIFF
--- a/lib/models/api-key.ts
+++ b/lib/models/api-key.ts
@@ -13,7 +13,7 @@ WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
 See the License for the specific language governing permissions and
 limitations under the License.
 */
-
+import once = require('lodash/once');
 import * as errors from 'balena-errors';
 
 import type * as BalenaSdk from '..';
@@ -24,6 +24,13 @@ const getApiKeysModel = function (
 	deps: InjectedDependenciesParam,
 	opts: InjectedOptionsParam,
 ) {
+	const applicationModel = once(() =>
+		(require('./application') as typeof import('./application')).default(
+			deps,
+			opts,
+		),
+	);
+
 	const { pine, request, sdkInstance } = deps;
 	const { apiUrl } = opts;
 	const exports = {
@@ -162,6 +169,51 @@ const getApiKeysModel = function (
 							name: {
 								$ne: null,
 							},
+						},
+						$orderby: 'name asc',
+					},
+					options,
+				),
+			});
+		},
+
+		/**
+		 * @summary Get all provisioning API keys for an application
+		 * @name getProvisioningApiKeysByApplication
+		 * @public
+		 * @function
+		 * @memberof balena.models.apiKey
+		 *
+		 * @param {String|Number} nameOrSlugOrId - application name (string) (deprecated), slug (string) or id (number)
+		 * @param {Object} [options={}] - extra pine options to use
+		 * @fulfil {Object[]} - apiKeys
+		 * @returns {Promise}
+		 *
+		 * @example
+		 * balena.models.apiKey.getProvisioningApiKeysByApplication('myorganization/myapp').then(function(apiKeys) {
+		 * 	console.log(apiKeys);
+		 * });
+		 *
+		 * @example
+		 * balena.models.apiKey.getProvisioningApiKeysByApplication(123, function(error, apiKeys) {
+		 * 	if (error) throw error;
+		 * 	console.log(apiKeys);
+		 * });
+		 */
+		async getProvisioningApiKeysByApplication(
+			nameOrSlugOrId: string | number,
+			options: BalenaSdk.PineOptions<BalenaSdk.ApiKey> = {},
+		): Promise<BalenaSdk.ApiKey[]> {
+			const { actor } = await applicationModel().get(nameOrSlugOrId, {
+				$select: 'actor',
+			});
+
+			return await pine.get({
+				resource: 'api_key',
+				options: mergePineOptions(
+					{
+						$filter: {
+							is_of__actor: actor,
 						},
 						$orderby: 'name asc',
 					},
--- a/tests/integration/models/api-key.spec.ts
+++ b/tests/integration/models/api-key.spec.ts
@@ -2,7 +2,7 @@
 import * as _ from 'lodash';
 import * as m from 'mochainon';
 import * as parallel from 'mocha.parallel';
-import { balena, givenLoggedInUser } from '../setup';
+import { balena, givenAnApplication, givenLoggedInUser } from '../setup';
 import { timeSuite } from '../../util';
 const { expect } = m.chai;
 
@@ -121,6 +121,43 @@ describe('API Key model', function () {
 					expect(apiKey).to.have.property('created_at').that.is.a('string');
 				});
 			});
+		});
+	});
+
+	describe('balena.models.apiKey.getProvisioningApiKeysByApplication', function () {
+		givenLoggedInUser(before);
+		givenAnApplication(before);
+
+		before(async function () {
+			const apiKey = await balena.models.application.generateProvisioningKey(
+				this.application.id,
+			);
+			expect(apiKey).to.be.a('string').that.has.length(32);
+
+			this.apiKey = apiKey;
+		});
+
+		it('should fail when the application does not exist', function () {
+			return expect(
+				balena.models.apiKey.getProvisioningApiKeysByApplication(
+					'nonExistentApp',
+				),
+			).to.be.rejected.then((error) => {
+				return expect(error).to.have.property(
+					'code',
+					'BalenaApplicationNotFound',
+				);
+			});
+		});
+
+		it('should be able to retrieve the provisioning api keys of an application', function () {
+			return balena.models.apiKey
+				.getProvisioningApiKeysByApplication(this.application.id)
+				.then((apiKeys) => {
+					expect(apiKeys).to.be.an('array');
+					expect(apiKeys).to.have.lengthOf(1);
+					expect(apiKeys[0].name).to.equal(null);
+				});
 		});
 	});
 
--- a/DOCUMENTATION.md
+++ b/DOCUMENTATION.md
@@ -270,6 +270,7 @@ const sdk = fromSharedOptions();
             * [.create(name, [description])](#balena.models.apiKey.create) ⇒ <code>Promise</code>
             * [.getAll([options])](#balena.models.apiKey.getAll) ⇒ <code>Promise</code>
             * [.getAllNamedUserApiKeys([options])](#balena.models.apiKey.getAllNamedUserApiKeys) ⇒ <code>Promise</code>
+            * [.getProvisioningApiKeysByApplication(nameOrSlugOrId, [options])](#balena.models.apiKey.getProvisioningApiKeysByApplication) ⇒ <code>Promise</code>
             * [.update(id, apiKeyInfo)](#balena.models.apiKey.update) ⇒ <code>Promise</code>
             * [.revoke(id)](#balena.models.apiKey.revoke) ⇒ <code>Promise</code>
         * [.key](#balena.models.key) : <code>object</code>
@@ -640,6 +641,7 @@ balena.models.device.get(123).catch(function (error) {
         * [.create(name, [description])](#balena.models.apiKey.create) ⇒ <code>Promise</code>
         * [.getAll([options])](#balena.models.apiKey.getAll) ⇒ <code>Promise</code>
         * [.getAllNamedUserApiKeys([options])](#balena.models.apiKey.getAllNamedUserApiKeys) ⇒ <code>Promise</code>
+        * [.getProvisioningApiKeysByApplication(nameOrSlugOrId, [options])](#balena.models.apiKey.getProvisioningApiKeysByApplication) ⇒ <code>Promise</code>
         * [.update(id, apiKeyInfo)](#balena.models.apiKey.update) ⇒ <code>Promise</code>
         * [.revoke(id)](#balena.models.apiKey.revoke) ⇒ <code>Promise</code>
     * [.key](#balena.models.key) : <code>object</code>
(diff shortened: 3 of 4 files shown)


lib/models/api-key.ts
```diff
@@ -170,6 +171,56 @@ const getApiKeysModel = function (
...
+		 * 	console.log(apiKeys);
+		 * });
+		 *
+		 * @example
+		 * balena.models.apiKey.getProvisioningApiKeysByApplication(function(error, apiKeys) {
+		 * 	if (error) throw error;
+		 * 	console.log(apiKeys);
+		 * });
+		 */
+		async getProvisioningApiKeysByApplication(
```
thgreasi, line 196: Since we say "ByApplication" the first argument should be an "appIdOrSlug: number | string" and then use application.get() in the implementation.
Otherwise this would return all provisioning API keys across all apps.

PS: The method will still have an options argument as it has right now, but as a second param.

thgreasi (changes requested):

lib/models/api-key.ts
nitishagar: Right. Thanks for the feedback here. 

I took this as a method which should return keys "ByApplication" as-in all applications in order for a user. Hence the use of `getAll()`.

lib/models/api-key.ts
```diff
@@ -19,6 +19,7 @@ import * as errors from 'balena-errors';
 import type * as BalenaSdk from '..';
 import type { InjectedDependenciesParam, InjectedOptionsParam } from '..';
 import { mergePineOptions } from '../util';
+import { default as applicationModel } from './application';
```
thgreasi, line 22: Let's change this to a lazy loaded import, so that the CLI doesn't have to load all files unless it's really needed.
```
	const applicationModel = once(() =>
		(require('./application') as typeof import('./application')).default(
			deps,
			opts,
		),
	);
```

lib/models/api-key.ts
```diff
@@ -170,6 +171,56 @@ const getApiKeysModel = function (
...
+			const { actor } = await applicationModel(deps, opts).get(nameOrSlugOrId);
+
+			return await pine.get({
+				resource: 'api_key',
+				options: mergePineOptions(
+					{
+						$filter: {
+							is_of__actor: {
+								$in: actor,
+							},
```
thgreasi, line 209:
```suggestion
							is_of__actor: actor,
```

lib/models/api-key.ts
```diff
@@ -170,6 +171,56 @@ const getApiKeysModel = function (
...
+							// the only way to reason whether it's
+							// a named user api key vs a deprecated user api key
+							// is whether it has a name.
+							name: {
+								$ne: null,
+							},
```
thgreasi, line 215:
```suggestion
```

lib/models/api-key.ts
```diff
@@ -170,6 +171,56 @@ const getApiKeysModel = function (
...
+		 * balena.models.apiKey.getProvisioningApiKeysByApplication(function(error, apiKeys) {
+		 * 	if (error) throw error;
+		 * 	console.log(apiKeys);
+		 * });
+		 */
+		async getProvisioningApiKeysByApplication(
+			nameOrSlugOrId: string | number,
+			options: BalenaSdk.PineOptions<BalenaSdk.ApiKey> = {},
+		): Promise<BalenaSdk.ApiKey[]> {
+			const { actor } = await applicationModel(deps, opts).get(nameOrSlugOrId);
```
thgreasi, line 200:
```suggestion
			const { actor } = await applicationModel(deps, opts).get(nameOrSlugOrId, { $select: 'actor' });
```

lib/models/api-key.ts
```diff
@@ -170,6 +171,56 @@ const getApiKeysModel = function (
...
+		 * @example
+		 * balena.models.apiKey.getProvisioningApiKeysByApplication().then(function(apiKeys) {
+		 * 	console.log(apiKeys);
+		 * });
+		 *
+		 * @example
+		 * balena.models.apiKey.getProvisioningApiKeysByApplication(function(error, apiKeys) {
+		 * 	if (error) throw error;
+		 * 	console.log(apiKeys);
+		 * });
```
thgreasi, line 194:
```suggestion
		 * balena.models.apiKey.getProvisioningApiKeysByApplication('myorganization/myapp').then(function(apiKeys) {
		 * 	console.log(apiKeys);
		 * });
		 *
		 * @example
		 * balena.models.apiKey.getProvisioningApiKeysByApplication(123, function(error, apiKeys) {
		 * 	if (error) throw error;
		 * 	console.log(apiKeys);
		 * });
```

lib/models/api-key.ts
```diff
@@ -170,6 +171,56 @@ const getApiKeysModel = function (
...
 		},
 
+		/**
+		 * @summary Get all provisioning API keys by application (asc)
+		 * @name getProvisioningApiKeysByApplication
+		 * @public
+		 * @function
+		 * @memberof balena.models.apiKey
+		 *
+		 * @param {Object} [options={}] - extra pine options to use
```
thgreasi, line 181:
```suggestion
		 * @param {String|Number} nameOrSlugOrId - application name (string) (deprecated), slug (string) or id (number)
		 * @param {Object} [options={}] - extra pine options to use
```

lib/models/api-key.ts
```diff
@@ -170,6 +171,56 @@ const getApiKeysModel = function (
 			});
 		},
 
+		/**
+		 * @summary Get all provisioning API keys by application (asc)
```
thgreasi, line 175:
```suggestion
		 * @summary Get all provisioning API keys for an application
```

lib/models/api-key.ts
```diff
@@ -170,6 +171,56 @@ const getApiKeysModel = function (
...
+		 * @public
+		 * @function
+		 * @memberof balena.models.apiKey
+		 *
+		 * @param {Object} [options={}] - extra pine options to use
+		 * @fulfil {Object[]} - apiKeys
+		 * @returns {Promise}
+		 *
+		 * @example
+		 * balena.models.apiKey.getProvisioningApiKeysByApplication().then(function(apiKeys) {
```
thgreasi, line 186: Can we also get a test case, where we create a provisioning api key using `balena.models.application.generateApiKey()` and then retrieve it here?

thgreasi (changes requested):

tests/integration/models/api-key.spec.ts
```diff
@@ -124,6 +124,33 @@ describe('API Key model', function () {
...
 	});
 
+	describe('balena.models.apiKey.getProvisioningApiKeysByApplication', function () {
+		givenLoggedInUser(before);
+
+		before(function () {
+			this.appName = 'SampleOrg/sampleApp';
+			return balena.models.application
+				.generateApiKey('SampleOrg/sampleApp')
+				.then(function (apiKey) {
```
thgreasi, line 134:
```suggestion
				.then((apiKey) => {
```
or convert to async-await, b/c the `this` in here is probably not what we think.

tests/integration/models/api-key.spec.ts
```diff
@@ -124,6 +124,33 @@ describe('API Key model', function () {
...
+
+		describe('balena.models.apiKey.getProvisioningApiKeysByApplication()', function () {
+			return balena.models.apiKey
+				.getProvisioningApiKeysByApplication(this.appName)
+				.then(function (apiKeys) {
+					expect(apiKeys).to.be.an('array');
+					expect(apiKeys).to.have.lengthOf(1);
+					expect(apiKeys[0].name).to.equal(this.appName);
+					expect(apiKeys[0].id).to.equal(this.apiKey);
+				});
```
thgreasi, line 150:
```suggestion
				.then((apiKeys) => {
					expect(apiKeys).to.be.an('array');
					expect(apiKeys).to.have.lengthOf(1);
					expect(apiKeys[0].name).to.equal(null); // but it might be `''`, plz double check.
					expect(apiKeys[0].id).to.equal(this.apiKey.id);
				});
```
or convert to async-await

thgreasi (changes requested):

tests/integration/models/api-key.spec.ts
```diff
@@ -124,6 +124,33 @@ describe('API Key model', function () {
...
+	describe('balena.models.apiKey.getProvisioningApiKeysByApplication', function () {
+		givenLoggedInUser(before);
+
+		before(function () {
+			this.appName = 'SampleOrg/sampleApp';
+			return balena.models.application
+				.generateApiKey('SampleOrg/sampleApp')
+				.then(function (apiKey) {
+					expect(_.isString(apiKey)).to.be.true;
+					expect(apiKey).to.have.length(32);
```
Page-, line 136: Mostly I want to use the `a('string')`, but combining it into one sentence is a bit nicer overall imo
```suggestion
					expect(apiKey).to.be.a('string').that.has.length(32);
```

tests/integration/models/api-key.spec.ts
```diff
@@ -124,6 +124,31 @@ describe('API Key model', function () {
...
+
+		describe('balena.models.apiKey.getProvisioningApiKeysByApplication()', function () {
+			return balena.models.apiKey
+				.getProvisioningApiKeysByApplication('SampleOrg/sampleApp')
+				.then(function (apiKeys) {
+					expect(apiKeys).to.be.an('array');
+					expect(apiKeys).to.have.lengthOf(1);
+					expect(apiKeys[0].name).to.equal('SampleOrg/sampleApp');
+					expect(apiKeys[0].id).to.equal(this.apiKey);
+				});
```
thgreasi, line 148: I think you missed this
```suggestion
				.then((apiKeys) => {
					expect(apiKeys).to.be.an('array');
					expect(apiKeys).to.have.lengthOf(1);
					expect(apiKeys[0].name).to.equal(null); // but it might be `''`, plz double check.
					expect(apiKeys[0].id).to.equal(this.apiKey.id);
				});
```

tests/integration/models/api-key.spec.ts
```diff
@@ -124,6 +124,31 @@ describe('API Key model', function () {
 		});
 	});
 
+	describe('balena.models.apiKey.getProvisioningApiKeysByApplication', function () {
```
thgreasi, line 127: This doesn't run at all atm, b/c there is no `it()` in here.
Please test locally by pointing your sdk against staging first.

tests/integration/models/api-key.spec.ts
```diff
@@ -124,6 +124,31 @@ describe('API Key model', function () {
...
+		givenLoggedInUser(before);
+
+		before(function () {
+			return balena.models.application
+				.generateApiKey('SampleOrg/sampleApp')
```
thgreasi, line 132: It doesn't look like this application is created anywhere, so I expect this to reject in all cases.
You should add a `givenAnApplication(before);` right after the `givenLoggedInUser` and then in here use 
```suggestion
				.generateApiKey(this.application.id)
```

tests/integration/models/api-key.spec.ts
```diff
@@ -124,6 +124,31 @@ describe('API Key model', function () {
...
+				.then((apiKey) => {
+					expect(apiKey).to.be.a('string').that.has.length(32);
+
+					this.apiKey = apiKey;
+				});
+		});
+
+		describe('balena.models.apiKey.getProvisioningApiKeysByApplication()', function () {
```
thgreasi, line 140:
```suggestion
		it('should fail when the application does not exist' , function () { /*...*/ });

		it('should be able to retrieve the provisioning api keys of an application' , function () {
```

thgreasi (changes requested):

tests/integration/models/api-key.spec.ts
```diff
@@ -124,6 +124,44 @@ describe('API Key model', function () {
 		});
 	});
 
+	describe.only('balena.models.apiKey.getProvisioningApiKeysByApplication', function () {
```
thgreasi, line 127:
```suggestion
	describe('balena.models.apiKey.getProvisioningApiKeysByApplication', function () {
```

tests/integration/models/api-key.spec.ts
```diff
@@ -247,3 +285,6 @@ describe('API Key model', function () {
 		});
 	});
 });
+function givenInitialOrganization(before: Mocha.HookFunction) {
```
thgreasi, line 288: ???

tests/integration/models/api-key.spec.ts
```diff
@@ -124,6 +124,44 @@ describe('API Key model', function () {
...
+			const apiKey = await balena.models.application.generateProvisioningKey(
+				this.application.id,
+			);
+			expect(apiKey).to.be.a('string').that.has.length(32);
+		});
+
+		// it('should fail when the application does not exist', function () {
+		// 	return expect(
+		// 		balena.models.apiKey.getProvisioningApiKeysByApplication(
+		// 			'nonExistentApp',
```
thgreasi, line 141:
```suggestion
		// 			'someorg/nonExistentApp',
```
Also, any reason this test case is commented out?

thgreasi:
@balena-ci rebase

tests/integration/models/api-key.spec.ts
nitishagar: Was WIP test, added to discuss the is-of mapping work.

tests/integration/models/api-key.spec.ts
nitishagar: Deleted.

tests/integration/models/api-key.spec.ts
nitishagar: Fixed and test locally.

thgreasi (approved):

tests/integration/models/api-key.spec.ts
```diff
@@ -124,6 +124,44 @@ describe('API Key model', function () {
...
+			return expect(
+				balena.models.apiKey.getProvisioningApiKeysByApplication(
+					'nonExistentApp',
+				),
+			).to.be.rejected.then((error) => {
+				expect(error).to.have.property('code', 'BalenaApplicationNotFound');
+
+				return expect(error)
+					.to.have.property('message')
+					.that.matches(/\bunique\b/i);
```
thgreasi, line 150: Well, the returned error is not a unique error, so this fails atm.
Let's just drop this expectation since we already have checked for the `BalenaApplicationNotFound` code, and it should be enough.

PR #1107, opened 2021-07-13
Update generateProvisioningKey to include keyName

nitishagar:
Change-Type: patch
Signed-off-by: Nitish Agarwal 1592163+nitishagar@users.noreply.github.com

---
##### Contributor checklist

- [x] Includes tests 
- [ ] Includes typings

Note: The test here are not retrieving key and checking the assigned name, but checking on whether we can set a value with the new API call.

### DIFF
--- a/lib/models/application.ts
+++ b/lib/models/application.ts
@@ -940,6 +940,7 @@ const getApplicationModel = function (
 		 * @memberof balena.models.application
 		 *
 		 * @param {String|Number} nameOrSlugOrId - application name (string) (deprecated), slug (string) or id (number)
+		 * @param {String} [keyName] - Provisioning key name
 		 * @fulfil {String} - device provisioning key
 		 * @returns {Promise}
 		 *
@@ -961,13 +962,20 @@ const getApplicationModel = function (
 		 */
 		generateProvisioningKey: async (
 			nameOrSlugOrId: string | number,
+			keyName?: string,
 		): Promise<string> => {
 			try {
 				const applicationId = await getId(nameOrSlugOrId);
 				const { body } = await request.send({
 					method: 'POST',
-					url: `/api-key/application/${applicationId}/provisioning`,
+					url: '/api-key/v1/',
 					baseUrl: apiUrl,
+					body: {
+						actorType: 'application',
+						actorTypeId: applicationId,
+						roles: ['provisioning-api-key'],
+						name: keyName,
+					},
 				});
 				return body;
 			} catch (err) {
--- a/tests/integration/models/application.spec.js
+++ b/tests/integration/models/application.spec.js
@@ -623,13 +623,13 @@ describe('Application Model', function () {
 
 			describe('balena.models.application.generateApiKey()', function () {
 				applicationRetrievalFields.forEach((prop) =>
-					it(`should be able to generate an API key by ${prop}`, function () {
-						return balena.models.application
-							.generateApiKey(this.application[prop])
-							.then(function (apiKey) {
-								expect(_.isString(apiKey)).to.be.true;
-								return expect(apiKey).to.have.length(32);
-							});
+					it(`should be able to generate an API key by ${prop}`, async function () {
+						const apiKey = await balena.models.application.generateApiKey(
+							this.application[prop],
+						);
+
+						expect(apiKey).to.be.a('string');
+						expect(apiKey).to.have.length(32);
 					}),
 				);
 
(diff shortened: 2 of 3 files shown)


tests/integration/models/application.spec.js
```diff
@@ -661,6 +661,17 @@ describe('Application Model', function () {
 					}),
 				);
 
+				applicationRetrievalFields.forEach((prop) =>
+					it(`should be able to generate a provisioning key by ${prop} with key name as key_${prop}`, function () {
+						return balena.models.application
+							.generateProvisioningKey(this.application[prop], `key_${prop}`)
+							.then(function (key) {
+								expect(_.isString(key)).to.be.true;
+								return expect(key).to.have.length(32);
```
thgreasi, line 670:
```suggestion
								expect(key).to.be.a('string');
								expect(key).to.have.length(32);
```

tests/integration/models/application.spec.js
```diff
@@ -661,6 +661,17 @@ describe('Application Model', function () {
 					}),
 				);
 
+				applicationRetrievalFields.forEach((prop) =>
+					it(`should be able to generate a provisioning key by ${prop} with key name as key_${prop}`, function () {
+						return balena.models.application
+							.generateProvisioningKey(this.application[prop], `key_${prop}`)
+							.then(function (key) {
```
thgreasi, line 668: `async await` would be nice for new code.

thgreasi (commented):
That's probably a `Change-Type: minor` right? Since we are adding a new feature in a pre-existing method.

nitishagar:
> That's probably a `Change-Type: minor` right? Since we are adding a new feature in a pre-existing method.

I put `patch` initially with the thinking that this is addition to an existing method (which is optional val). But updated to `minor`.

thgreasi:
To which do you refer to? `ResinCI/build/CI` ?
I've also restarted the appVeyor, which is the primary required one here.

nitishagar:
> To which do you refer to? `ResinCI/build/CI` ?
> I've also restarted the appVeyor, which is the primary required one here.

@thgreasi Yes, I meant the `ResinCI...` one. Rebased the changes. Take a look when you get a chance.

lib/models/application.ts
```diff
@@ -940,6 +940,7 @@ const getApplicationModel = function (
 		 * @memberof balena.models.application
 		 *
 		 * @param {String|Number} nameOrSlugOrId - application name (string) (deprecated), slug (string) or id (number)
+		 * @param {String} keyName - Provisioning key name (string) - [OPTIONAL]
```
thgreasi, line 943:
```suggestion
		 * @param {String} [keyName] - Provisioning key name
```
The docs generator should do the rest :)

tests/integration/models/application.spec.js
```diff
@@ -623,13 +623,13 @@ describe('Application Model', function () {
...
-								expect(_.isString(apiKey)).to.be.true;
-								return expect(apiKey).to.have.length(32);
-							});
+					it(`should be able to generate an API key by ${prop}`, async function () {
+						const apiKey = await balena.models.application.generateApiKey(
+							this.application[prop],
+						);
+
+						expect(apiKey).to.be.a('string');
+						expect(apiKey).to.have.length(32);
```
thgreasi, line 632: Can we also get the latest api key that we have for this app and check it's name property?
We can also store the last known api key id and use it in the filter so that we can be more confident that the one that we got back is the one that we just created.

tests/integration/models/application.spec.js
```diff
@@ -661,6 +661,18 @@ describe('Application Model', function () {
...
 
+				applicationRetrievalFields.forEach((prop) =>
+					it(`should be able to generate a provisioning key by ${prop} with key name as key_${prop}`, async function () {
+						const key = await balena.models.application.generateProvisioningKey(
+							this.application[prop],
+							`key_${prop}`,
+						);
+
+						expect(key).to.be.a('string');
+						expect(key).to.have.length(32);
```
thgreasi, line 672: > Can we also get the latest api key that we have for this app and check it's name property?
We can also store the last known api key id and use it in the filter so that we can be more confident that the one that we got back is the one that we just created.

thgreasi (commented):
Other than that LGTM

thgreasi:
Also, `ResinCI/build/CI` is failing b/c we still are not able to test the SDK against balenaCI node runners.
In appVeyor we inject user credentials to be used by the runner, which balenaCI doesn't support, and as a result we have those marked as non-required.

tests/integration/models/application.spec.js
nitishagar: This will not have `name` field assigned. Only refactored because of the `await` thingy.

Updated the provisioning call for the name check.

thgreasi (approved):

PR #1111, opened 2021-07-15
os.getConfig: Add typings for the provisioningKeyName option

nitishagar:
Change-Type: minor

Depends-on: https://github.com/balena-io/open-balena-api/pull/716
Depends-on: https://github.com/balena-io/balena-api/pull/3298

---
##### Contributor checklist

- [ ] Includes tests
- [ ] Includes typings

### DIFF
--- a/lib/models/os.ts
+++ b/lib/models/os.ts
@@ -30,6 +30,7 @@ const BALENAOS_VERSION_REGEX = /v?\d+\.\d+\.\d+(\.rev\d+)?((\-|\+).+)?/;
 export interface ImgConfigOptions {
 	network?: 'ethernet' | 'wifi';
 	appUpdatePollInterval?: number;
+	provisioningKeyName?: string;
 	wifiKey?: string;
 	wifiSsid?: string;
 	ip?: string;
@@ -483,6 +484,7 @@ const getOsModel = function (
 	 * the device will use, one of 'ethernet' or 'wifi'.
 	 * @param {Number} [options.appUpdatePollInterval] - How often the OS checks
 	 * for updates, in minutes.
+	 * @param {String} [options.provisioningKeyName] - Name assigned to API key
 	 * @param {String} [options.wifiKey] - The key for the wifi network the
 	 * device will connect to.
 	 * @param {String} [options.wifiSsid] - The ssid for the wifi network the
--- a/tests/integration/models/os.spec.ts
+++ b/tests/integration/models/os.spec.ts
@@ -664,6 +664,36 @@ describe('OS model', function () {
 					});
 			});
 
+			it('should be able to configure v2 image with a provisioning key name', async function () {
+				const provisioningKeyName = `provision-key-${new Date()}`;
+				const configOptions = {
+					appUpdatePollInterval: 72,
+					network: 'wifi' as const,
+					wifiKey: 'foobar',
+					wifiSsid: 'foobarbaz',
+					ip: '1.2.3.4',
+					gateway: '5.6.7.8',
+					netmask: '9.10.11.12',
+					version: '2.11.8+rev1.prod',
+					provisioningKeyName,
+				};
+
+				await balena.models.os.getConfig(ctx.application.id, configOptions);
+
+				const provisioningKeys = _.filter(
+					await balena.models.apiKey.getProvisioningApiKeysByApplication(
+						ctx.application.id,
+					),
+					['name', provisioningKeyName],
+				);
+
+				expect(provisioningKeys).to.be.an('array');
+				expect(provisioningKeys.length).is.equal(1);
+				expect(provisioningKeys[0])
+					.to.have.property('name')
+					.to.be.equal(provisioningKeyName);
+			});
+
 			it('should be rejected if the application id does not exist', function () {
 				const promise = balena.models.os.getConfig(999999, {
 					version: DEFAULT_OS_VERSION,


thgreasi (commented):
LGTM as far as we get the respective API changes on prod first

lib/models/os.ts
```diff
@@ -30,6 +30,7 @@ const BALENAOS_VERSION_REGEX = /v?\d+\.\d+\.\d+(\.rev\d+)?((\-|\+).+)?/;
 export interface ImgConfigOptions {
 	network?: 'ethernet' | 'wifi';
 	appUpdatePollInterval?: number;
+	provisioningKeyName?: string;
```
thgreasi, line 33: Let's also add a test case to test that `getConfig` also sets the name that we are providing to the generated provisioning api key.

thgreasi:
@nitishagar let's also change the commit message to something like:
> os.getConfig: Add typings for the provisioningKeyName option

lib/models/os.ts
nitishagar: I added a `provisioningKeyName` for the `getConfig` spec, but we don't get back the key name only the actual key. I tested this functionality in the core-api here: https://github.com/balena-io/open-balena-api/pull/716/files#diff-1880e5d78ef6ce81f0e72c9bf111a7783e5dd39c5de9278026b13fae90104f4dR81

Should you see a way to access the key object on the SDK spec side?

nitishagar:
This will pass only with `open-balena-api` release with following: https://github.com/balena-io/open-balena-api/pull/716

tests/integration/models/os.spec.ts
```diff
@@ -664,6 +665,33 @@ describe('OS model', function () {
 					});
 			});
 
+			it.only('should be able to configure v2 image with a provisioning key name', async function () {
```
thgreasi, line 668:
```suggestion
			it('should be able to configure v2 image with a provisioning key name', async function () {
```

thgreasi (approved):
LGTM but reminder that this needs to wait for the API to go on prod first

tests/integration/models/os.spec.ts
nitishagar: Ahh - Done!

nitishagar:
> LGTM but reminder that this needs to wait for the API to go on prod first

PR raised: https://github.com/balena-io/balena-api/pull/3298

tests/integration/models/os.spec.ts
```diff
@@ -664,6 +665,33 @@ describe('OS model', function () {
...
+
+				await balena.models.os.getConfig(ctx.application.id, configOptions);
+
+				const provisioningKeys =
+					await balena.models.apiKey.getProvisioningApiKeysByApplication(
+						ctx.application.id,
+					);
+
+				expect(provisioningKeys).to.be.an('array');
+				expect(provisioningKeys).to.have.length(1);
```
thgreasi, line 689: This still failing.
For sure there are more than 1 keys at this point in the test suite.
Why don't you just do it just like in the API PR?
See: https://github.com/balena-io/balena-sdk/pull/1107/files#diff-6e5edef2fac65498ec08862f218a130979112a22de7b03b6405eda18c16eb445R688-R702

nitishagar: I know - I was looking to move away from depending on order and checking `[0]` index as well. Updated to generic `find` operation now.

tests/integration/models/os.spec.ts
```diff
@@ -649,6 +649,7 @@ describe('OS model', function () {
 					gateway: '5.6.7.8',
 					netmask: '9.10.11.12',
 					version: '2.0.8+rev1.prod',
+					provisioningKeyName: 'provision-key',
```
thgreasi, line 652:
```suggestion
```
If all test cases provide a name, then the is no reason to have a separate one that tests the name was indeed assigned.

tests/integration/models/os.spec.ts
```diff
@@ -664,6 +665,33 @@ describe('OS model', function () {
...
+				const configOptions = {
+					appUpdatePollInterval: 72,
+					network: 'wifi' as const,
+					wifiKey: 'foobar',
+					wifiSsid: 'foobarbaz',
+					ip: '1.2.3.4',
+					gateway: '5.6.7.8',
+					netmask: '9.10.11.12',
+					version: '2.11.8+rev1.prod',
+					provisioningKeyName: 'provision-key',
```
thgreasi, line 678: We should either give a unique name here (since this name is also used in the right above test case) or not give an name in any other case.

tests/integration/models/os.spec.ts
```diff
@@ -664,6 +665,33 @@ describe('OS model', function () {
...
+				const provisioningKeys =
+					await balena.models.apiKey.getProvisioningApiKeysByApplication(
+						ctx.application.id,
+					);
+
+				expect(provisioningKeys).to.be.an('array');
+				expect(provisioningKeys.length).is.not.equal(0);
+				expect(
+					_.findIndex(provisioningKeys, ['name', 'provision-key']),
+				).to.not.equal(-1);
```
thgreasi, line 692: 1. get the provisioning api keys before your call
2. generate the named config
3. fetch the provisioning api keys ordered by created_at desc
4. filter the newly retrieved api keys to exclude the ones that are in the list of IDs of the api keys of step 1
5. expect that what's left is has length of 1
6. expect hat the single new apikey has a property name that equals the name you provided.

thgreasi (changes requested):

tests/integration/models/os.spec.ts
nitishagar: Okay - so on very similar lines like the last time (https://github.com/balena-io/balena-sdk/pull/1107/files#diff-6e5edef2fac65498ec08862f218a130979112a22de7b03b6405eda18c16eb445R692).

tests/integration/models/os.spec.ts
```diff
@@ -664,6 +664,36 @@ describe('OS model', function () {
...
+				};
+
+				await balena.models.os.getConfig(ctx.application.id, configOptions);
+
+				const provisioningKeys = _.filter(
+					await balena.models.apiKey.getProvisioningApiKeysByApplication(
+						ctx.application.id,
+					),
+					['name', provisioningKeyName],
+				);
```
thgreasi, line 688:
```suggestion
				const provisioningKeys = await balena.models.apiKey.getProvisioningApiKeysByApplication(
					ctx.application.id,
					{ $filter: { name: provisioningKeyName } },
				);
```

tests/integration/models/os.spec.ts
```diff
@@ -664,6 +664,36 @@ describe('OS model', function () {
 					});
 			});
 
+			it('should be able to configure v2 image with a provisioning key name', async function () {
+				const provisioningKeyName = `provision-key-${new Date()}`;
```
thgreasi, line 668:
```suggestion
				const provisioningKeyName = `provision-key-${Date.now()}`;
```

thgreasi (approved):

PR #1113, opened 2021-07-28
apiKey.getAll: Return only NamedUserApiKeys for backwards compatibility

thgreasi:
So far `getAll()` was only returning named user api keys. With the current code, once we add support for naming provisioning api keys as well, `getAll()` will start returning those as well. but not the non-named ones, which would lead to confusion.
Changing it to return all apiKeys should better be in the next major, so the safest thing to do for users of the current major is probably to change it so that it will only be returning named user api keys, like it always did.

Change-type: patch

---
##### Contributor checklist

- [ ] Includes tests
- [ ] Includes typings

### DIFF
--- a/lib/models/api-key.ts
+++ b/lib/models/api-key.ts
@@ -110,26 +110,10 @@ const getApiKeysModel = function (
 		 * 	console.log(apiKeys);
 		 * });
 		 */
-		getAll(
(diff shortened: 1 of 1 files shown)


thgreasi:
@balena-ci I self-certify!

PR #1121, opened 2021-09-06
Use the hostApp releases for os.download() & os.getSupportedOsUpdateVersions()

thgreasi:
These are cloud only atm, so it's non-breaking to switch them to rely on the hostApps instead of `/device-types/v1`.

Change-type: minor

---
##### Contributor checklist

- [ ] Includes tests
- [ ] Includes typings


thgreasi:
@balena-ci I self-certify!

Issue #1122, opened 2021-09-09
Add `finalize` method to the release object


toochevere:
## Description
Releases can now by `draft` or `final`.  A `draft` release can be promoted into a `final` status.  However this functionality currently only exists on the api and needs to be reflected in the SDK.

## References
See:  https://www.flowdock.com/app/rulemotion/r-supervisor/threads/JN2-gnspQ-v6WaeWCvm9T8NYDY1

PR #1123, opened 2021-09-10
Add release.finalize to promote draft releases to final

toochevere:
Change-type: minor
Signed-off-by: Paul Jonathan <pj@balena.io>
Resolves: https://github.com/balena-io/balena-sdk/issues/1122

---
##### Contributor checklist

- [x] Includes tests

### DIFF
--- a/lib/models/release.ts
+++ b/lib/models/release.ts
@@ -390,6 +390,39 @@ const getReleaseModel = function (
 	}
 
 	/**
+	 * @summary Finalizes a draft release
+	 * @name finalize
+	 * @public
+	 * @function
+	 * @memberof balena.models.release
+	 *
+	 * @param {String|Number} commitOrId - release commit (string) or id (number)
+	 * @fulfil {void}
+	 * @returns {Promise}
+	 *
+	 * @example
+	 * balena.models.release.finalize(123).then(function() {
+	 * 	console.log('finalized!');
+	 * });
+	 *
+	 * @example
+	 * balena.models.release.finalize('7cf02a6').then(function() {
+	 * 	console.log('finalized!');
+	 * });
+	 *
+	 */
+	async function finalize(commitOrId: string | number): Promise<void> {
+		const { id } = await get(commitOrId, { $select: 'id' });
+		await pine.patch<Release>({
+			resource: 'release',
+			id,
+			body: {
+				is_final: true,
+			},
+		});
+	}
+
+	/**
 	 * @namespace balena.models.release.tags
 	 * @memberof balena.models.release
 	 */
@@ -577,6 +610,7 @@ const getReleaseModel = function (
 		getLatestByApplication,
 		getWithImageDetails,
 		createFromUrl,
+		finalize,
 		tags,
 	};
 };
--- a/tests/integration/models/release.spec.js
+++ b/tests/integration/models/release.spec.js
@@ -20,6 +20,8 @@ import {
 	itShouldGetAllTagsByResource,
 } from './tags';
 
+const releaseRetrievalFields = ['id', 'commit'];
+
 describe('Release Model', function () {
 	timeSuite(before);
 	givenLoggedInUser(before);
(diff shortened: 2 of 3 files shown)


lib/models/release.ts
```diff
@@ -389,6 +389,56 @@ const getReleaseModel = function (
...
+	 * @example
+	 * balena.models.release.finalize('7cf02a6').then(function() {
+	 * 	console.log('finalized!');
+	 * });
+	 *
+	 */
+	async function finalize(commitOrId: string | number): Promise<void> {
+		if (commitOrId == null) {
+			throw new errors.BalenaReleaseNotFound(commitOrId);
+		}
```
thgreasi, line 417: Let's use `const { id } = await models.release.get(commitOrId, { $select: 'id' });` and use the returned `id` so that we also automatically throw in case more than one releases are returned.

lib/models/release.ts
```diff
@@ -389,6 +389,56 @@ const getReleaseModel = function (
...
+				body: {
+					is_final: true,
+				},
+			});
+		} else {
+			await pine.patch<Release>({
+				resource: 'release',
+				body: {
+					is_final: true,
+				},
```
thgreasi, line 432: NIT: In general we prefer having the `body` after the `options` so that we always first have to think what the `.patch` is going to affect and only after that think which props are going to change.

tests/integration/models/release.spec.js
```diff
@@ -286,6 +286,66 @@ describe('Release Model', function () {
...
+						source: 'cloud',
+						is_final: false,
+						composition: '{}',
+						start_timestamp: new Date().toISOString(),
+					},
+				});
+			};
+
+			it('should finalize a release by id', function () {
+				return createDraftRelease('abcdefg').then(function (draftRelease) {
```
thgreasi, line 308: async await :cry: :grey_question:

tests/integration/models/release.spec.js
```diff
@@ -286,6 +286,66 @@ describe('Release Model', function () {
...
+							});
+							expect(finalRelease).to.deep.match({
+								commit: 'efghijklmnop',
+								id: draftRelease.id,
+								is_final: true,
+							});
+						});
+				});
+			});
+		});
```
thgreasi, line 347: How about the following ?
```suggestion
		describe('balena.model.release.finalize()', function () {
			const fields = ['id', 'commit'] as const;
			const testReleaseByField: Dictionary<BalenaSdk.Release> = {};
			const createDraftRelease = async function (commit) {
				const userId = await balena.auth.getUserId();
				return await balena.pine.post<BalenaSdk.Release>({
					resource: 'release',
					body: {
						belongs_to__application: this.application.id,
						is_created_by__user: userId,
						commit: commit,
						status: 'success',
						source: 'cloud',
						is_final: false,
						composition: '{}',
						start_timestamp: new Date().toISOString(),
					},
				});
			};

			before(async function () {
				let i = 0;
				await Promise.all(fields.map(async (field, i)) => {
					testReleaseByField[field] = await createDraftRelease(`abcdef${i}`);
				});
			});

			after(async function () {
				await balena.pine.delete({
					resource: 'release',
					options: {
						$filter: {
							id: { $in: Object.values(testReleaseIdByField) }
						}
					}
				});
			})

			parallel('', function () {
				fields.forEach((field) => {
					it(`should finalize a release by ${field}`, async function () {
						const draftRelease = testReleaseIdByField[field];
						await balena.models.release .finalize(draftRelease[field]);

						const finalRelease = await balena.models.release.get(
							draftRelease.id,
						);
						expect(finalRelease).to.deep.match({
							id: draftRelease.id,
							commit: draftRelease.commit,
							is_final: true,
						});
					});
				});
			});
		});
```
Afair we need the extra `decribe` b/c `parallel` is a little picky with `before`s & `after`s.

lib/models/release.ts
```diff
@@ -389,6 +389,56 @@ const getReleaseModel = function (
 		);
 	}
 
+	/**
+	 * @summary Promotes a 'draft' release to 'final' (non-draft)
```
thgreasi, line 393: I'm fine with it as is, but it felt like it's referring too much to the allowed values of the deprecated release_type field. How about simply something like v ?
```suggestion
	 * @summary Finalizes a draft release
```

lib/models/release.ts
toochevere: I had thought about doing something like this, but it's significantly less performant.  Probably makes no difference, and as you mention, safer.

thgreasi: I also had complains about this pattern, but:
* we re-use the same "what to query on" logic
* it also throws 404 if the item isn't found (pine doesn't do that for PATCHes, but it should)

tests/integration/models/release.spec.js
toochevere: I took a slightly different approach because my first way of doing things was breaking other tests.  Tell me what you think of it now.

tests/integration/models/release.spec.js
toochevere: isn't returning the promise more performant than awaiting in cases like this?  In any case, take a look at the new approach.

lib/models/release.ts
```diff
@@ -389,6 +389,53 @@ const getReleaseModel = function (
...
+		} else {
+			const { id } = await get(commitOrId, { $select: 'id' });
+			await pine.patch<Release>({
+				resource: 'release',
+				id,
+				body: {
+					is_final: true,
+				},
+			});
+		}
```
thgreasi, line 436: @toochevere what I meant was actually:
```suggestion
		const { id } = await get(commitOrId, { $select: 'id' });
		await pine.patch<Release>({
			resource: 'release',
			id: commitOrId,
			body: {
				is_final: true,
			},
		});
```

tests/integration/models/release.spec.js
thgreasi: Afair since node 12 there is ~0 cost in async await vs promises.
In general all new code we write since then is all in async-await, and use promise chaining only in rare cases.
Moreover since this is about tests, imo optimizing for readability rather than perf makes sense (especially since the bottleneck in the sdk tests is the API requests and not the js execution time).

lib/models/release.ts
toochevere: oh, of course

tests/integration/models/release.spec.js
thgreasi: @toochevere in general we try to keep the number of app creation-deletions to a minimum, since that's a slow operation atm in the API.
The problem with deleting the release was that the release was still the target release of the application when trying to delete it, and that's why you needed to delete the whole app.

I moved this at the end of the above `describe()` section, just before the app gets deleted.
If you like it, then feel free to squash the commit with yours and we can get this merged.

thgreasi (approved):

PR #1124, opened 2021-09-10
typings: Drop the v5-model-only application_type.is_host_os

thgreasi:
This is no longer a property in the v6 model and we forgot to drop it when we upgraded the SDK to v6.

Change-type: patch

Resolves: # 
HQ: <url> 
See: <url> 
Depends-on: <url> 

---
##### Contributor checklist

- [ ] Includes tests
- [ ] Includes typings


thgreasi:
@balena-ci I self-certify!

PR #1140, opened 2021-11-01
Add 400 error handling for email verification.

nitishagar:
Change-Type: patch

See: https://www.flowdock.com/app/rulemotion/r-product/threads/GTH4UJIdUSFRAc_OhXnsCfe80m7
See: https://docs.google.com/document/d/197za9xgUfhzstuPxCy18z0YVnWkpK6ce0geX7B6T7iE/edit

### DIFF
--- a/lib/auth.ts
+++ b/lib/auth.ts
@@ -520,18 +520,30 @@ const getAuth = function (
 		const email = verificationPayload.email;
 		const verificationToken = verificationPayload.token;
 
-		const { body } = await request.send({
-			method: 'POST',
-			url: '/user/v1/verify-email',
-			body: {
-				verificationToken,
-				email,
-			},
-			baseUrl: apiUrl,
-			sendToken: false,
-		});
+		try {
+			const { body } = await request.send({
+				method: 'POST',
+				url: '/user/v1/verify-email',
+				body: {
+					verificationToken,
+					email,
+				},
+				baseUrl: apiUrl,
+				sendToken: false,
+			});
 
-		return body;
+			return body;
+		} catch (err) {
+			if (err.statusCode === 400) {
+				throw new errors.BalenaRequestError(
+					'Your verification link has expired - log in again to request another one.',
+					err.statusCode,
+					{},
+				);
+			}
+
+			throw err;
+		}
 	}
 
 	/**


thgreasi:
@nitishagar I now lean towards moving this to the API itself so that all error messaging is in one place and we don't have to repeat this in the python sdk as well.

nitishagar:
@thgreasi We already have more granular API error messaging. This was to make it more generic and consistent with dashboard messaging. 

Where do we have the Python SDK in use? As far as I know we don't have any updates to Python SDK yet to even invoke verification APIs.

lib/auth.ts
```diff
@@ -520,18 +520,30 @@ const getAuth = function (
...
+					email,
+				},
+				baseUrl: apiUrl,
+				sendToken: false,
+			});
 
-		return body;
+			return body;
+		} catch (err) {
+			if (err.statusCode === 400) {
```
thgreasi, line 537: This could also be b/c of providing an invalid email or verification code and doesn't suffice to reason whether the provided code was expired. Maybe we should also match the returned error message as well and throw a `new BalenaError`.

lib/auth.ts
```diff
@@ -520,18 +520,30 @@ const getAuth = function (
...
+				throw new errors.BalenaRequestError(
+					'Your verification link has expired - log in again to request another one.',
```
thgreasi, line 539: I took a better look of our messaging and flow and I'm still not convinced that the SDK is the right place for this.
Especially since `link` and `log in again` refer to a UI specific user flow that the SDK doesn't necessarily care about (the sdk really just takes a code and doesn't require you to re-login if you are already logged-in in order to request a new code).

What if we later want to add a "request new code" in the same page that the user verification link navigates you to, thus the user not having to re-login to request a new one if expired?
Doing such a UI-only user flow change should not also require changes to the SDK code right (the error messaging in this case)?

Imo let's do the simplest thing of having UI-specific user messaging be in the UI.

nitishagar: @thgreasi Didn't quite get the scenario you mentioned below:
> What if we later want to add a "request new code" in the same page that the user verification link navigates you to, thus the user not having to re-login to request a new one if expired?
Doing such a UI-only user flow change should not also require changes to the SDK code right (the error messaging in this case)?

Also, I agree to having this part of UI as this is only UI specific messaging. Do we want to segregate the type of error on the UI or update the wording of the above error msg to be a bit more generic? e.g. `'Issue with the verification link provided - log in again to generate a new one.'`

lib/auth.ts
nitishagar: Do we want to segregate the type of error on the UI or update the wording of the above error msg to be a bit more generic? e.g. `'Issue with the verification link provided - log in again to generate a new one.'`

thgreasi: Let's just go with "Your verification link has expired or is invalid - log in again to request another one."

PR #1152, opened 2021-11-23
Add release setIsInvalidated function

myarmolinsky:
Add release setIsInvalidated function

Change-type: minor
Signed-off-by: Matthew Yarmolinsky <matthew-timothy@balena.io>

---
##### Contributor checklist

- [ ] Includes tests
- [ ] Includes typings

### DIFF
--- a/lib/models/release.ts
+++ b/lib/models/release.ts
@@ -423,6 +423,53 @@ const getReleaseModel = function (
 	}
 
 	/**
+	 * @summary Set the is_invalidated property of a release to true or false
+	 * @name setIsInvalidated
+	 * @public
+	 * @function
+	 * @memberof balena.models.release
+	 *
+	 * @param {String|Number} commitOrId - release commit (string) or id (number)
+	 * @param {Boolean} isInvalidated - boolean value, true for invalidated, false for validated
+	 * @fulfil {void}
+	 * @returns {Promise}
+	 *
+	 * @example
+	 * balena.models.release.setIsInvalidated(123, true).then(function() {
+	 * 	console.log('invalidated!');
+	 * });
+	 *
+	 * @example
+	 * balena.models.release.setIsInvalidated('7cf02a6', true).then(function() {
+	 * 	console.log('invalidated!');
+	 * });
+	 *
+	 * @example
+	 * balena.models.release.setIsInvalidated(123, false).then(function() {
+	 * 	console.log('validated!');
+	 * });
+	 *
+	 * @example
+	 * balena.models.release.setIsInvalidated('7cf02a6', false).then(function() {
+	 * 	console.log('validated!');
+	 * });
+	 *
+	 */
+	async function setIsInvalidated(
+		commitOrId: string | number,
+		isInvalidated: boolean,
+	): Promise<void> {
+		const { id } = await get(commitOrId, { $select: 'id' });
+		await pine.patch<Release>({
+			resource: 'release',
+			id,
+			body: {
+				is_invalidated: isInvalidated,
+			},
+		});
+	}
+
+	/**
 	 * @summary Add a note to a release
 	 * @name note
 	 * @public
@@ -643,6 +690,7 @@ const getReleaseModel = function (
 		getWithImageDetails,
 		createFromUrl,
 		finalize,
+		setIsInvalidated,
 		note,
 		tags,
 	};
--- a/tests/integration/models/release.spec.js
+++ b/tests/integration/models/release.spec.js
@@ -276,6 +276,44 @@ describe('Release Model', function () {
 					});
 				});
 			});
+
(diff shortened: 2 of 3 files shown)


lib/models/release.ts
```diff
@@ -422,6 +422,53 @@ const getReleaseModel = function (
...
+	 * @returns {Promise}
+	 *
+	 * @example
+	 * balena.models.release.setIsInvalidated(123, true).then(function() {
+	 * 	console.log('invalidated!');
+	 * });
+	 *
+	 * @example
+	 * balena.models.release.setIsInvalidated('7cf02a6', true).then(function() {
+	 * 	console.log('isvalidated!');
```
thgreasi, line 444:
```suggestion
	 * 	console.log('invalidated!');
```

tests/integration/models/release.spec.js
```diff
@@ -276,6 +276,44 @@ describe('Release Model', function () {
...
+								draftRelease.id,
+							);
+							expect(invalidatedRelease).to.deep.match({
+								id: draftRelease.id,
+								commit: draftRelease.commit,
+								is_invalidated: false,
+							});
+						});
+					});
+				});
```
thgreasi, line 315: There is obviously a race condition when running these tests in parallel, right :) ?
(plus you are checking for `is_invalidated: false` in both tests)
Try the following:
```suggestion
				releaseRetrievalFields.forEach((field) => {
					parallel('', function () {
							it(`should invalidate a release by ${field}`, async function () {
								const release = testReleaseByField[field];
								await balena.models.release.setIsInvalidated(
									release[field],
									true,
								);
								const invalidatedRelease = await balena.models.release.get(
									release.id,
									{ $select: 'is_invalidated' }
								);
								expect(invalidatedRelease).to.deep.match({
									is_invalidated: true,
								});
							});
						});
					});

					parallel('', function () {
						it(`should validate a release by ${field}`, async function () {
							const release = testReleaseByField[field];
							await balena.models.release.setIsInvalidated(
								release[field],
								false,
							);
							const validatedRelease = await balena.models.release.get(
								release.id,
								{ $select: 'is_invalidated' }
							);
							expect(validatedRelease).to.deep.match({
								is_invalidated: false,
							});
						});
					});
				});
```

lib/models/release.ts
myarmolinsky: Nice catch, thank you 👍

tests/integration/models/release.spec.js
myarmolinsky: Didn't notice the `is_invalidated: false` 😖, thanks for catching that
Tried your suggestion locally and the tests passed :)

tests/integration/models/release.spec.js
```diff
@@ -276,6 +276,44 @@ describe('Release Model', function () {
 					});
 				});
 			});
+
+			describe('balena.model.release.setIsInvalidated()', function () {
+				releaseRetrievalFields.forEach((field) => {
+					parallel('', function () {
+						it(`should validate a release by ${field}`, async function () {
```
thgreasi, line 283: This release was already valid even before the test :)
We might better move it below the other.

tests/integration/models/release.spec.js
```diff
@@ -276,6 +276,44 @@ describe('Release Model', function () {
 					});
 				});
 			});
+
+			describe('balena.model.release.setIsInvalidated()', function () {
+				releaseRetrievalFields.forEach((field) => {
+					parallel('', function () {
```
thgreasi, line 282: This is causing a race condition between invalidating & validating the same release.
Either drop the parallel and let them run in order or check my earlier suggestion which parallelizes the independent test cases.

tests/integration/models/release.spec.js
```diff
@@ -276,6 +276,44 @@ describe('Release Model', function () {
...
+						it(`should validate a release by ${field}`, async function () {
+							const invalidRelease = testReleaseByField[field];
+							await balena.models.release.setIsInvalidated(
+								invalidRelease[field],
+								false,
+							);
+							const validatedRelease = await balena.models.release.get(
+								invalidRelease.id,
```
thgreasi, line 290: NIT: Add a `$select` here as per my earlier comment (and in the lower `release.get()` ) so that we don't waste resources in retrieving unnecessary properties.

thgreasi (approved):

tests/integration/models/release.spec.js
```diff
@@ -276,6 +276,44 @@ describe('Release Model', function () {
 					});
 				});
 			});
+
+			describe('balena.model.release.setIsInvalidated()', function () {
+				releaseRetrievalFields.forEach((field) => {
```
thgreasi, line 281: Oh I just realized that this should be INSIDE both `parallel()` calls :man_facepalming: 
Anyway, no big deal...

PR #1165, opened 2021-12-21
os.getMaxSatisfyingVersion: Add support for ESR releases

thgreasi:
Resolves: #389
Change-type: minor

Resolves: # 
HQ: <url> 
See: <url> 
Depends-on: <url> 

---
##### Contributor checklist

- [ ] Includes tests
- [ ] Includes typings

### DIFF
--- a/tests/integration/models/os.spec.ts
+++ b/tests/integration/models/os.spec.ts
@@ -381,7 +381,30 @@ describe('OS model', function () {
 	});
 
 	describe('balena.models.os._getMaxSatisfyingVersion()', function () {
-		const osVersions = [
+		const esrOsVersions = [
+			{
+				rawVersion: '2021.10.2.prod',
+				isRecommended: true,
+			},
+			{ rawVersion: '2021.10.2.dev' },
+			{ rawVersion: '2021.07.1.prod' },
+			{ rawVersion: '2021.07.1.dev' },
+			{ rawVersion: '2021.04.0.prod' },
+			{ rawVersion: '2021.04.0.dev' },
+			{ rawVersion: '2021.01.0.prod' },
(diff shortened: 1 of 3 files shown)


thgreasi:
@balena-ci I self-certify!

tests/integration/models/os.spec.ts
```diff
@@ -438,24 +473,39 @@ describe('OS model', function () {
...
+				'2020.07.2.dev',
+			));
+
 		it('should return an equivalent result, if no exact result exists, when given a specific version', () =>
 			expect(_getMaxSatisfyingVersion('2.73.1+rev1', osVersions)).to.equal(
 				'2.73.1+rev1.prod',
 			));
 
-		it('should support ^ semver ranges', () =>
+		it('should support ^ semver ranges in default OS releases', () =>
```
pdcastro, line 486: If we had an additional a test case here like:
```ts
		it('should support > semver ranges', () =>
			expect(_getMaxSatisfyingVersion('>2.80.3', osVersions)).to.equal(
				'2021.07.1.prod',
			));
```
I believe it would it produce `2021.07.1.prod`, which may or may not be an expected result (e.g. for users who don't have access to ESR images). But I think it is OK, because:
1. I see that `getMaxSatisfyingVersion` now accepts a 3rd argument to filter on `esr` versions, which allows clients (CLI, `balena-image-manager`, UI) to restrict `MaxSatisfyingVersion` to ESR or non ESR versions.
2. In practice, I think it is very unlike that any user would want to specify `>2.80.3` when downloading an OS image.

thgreasi: See: https://github.com/balena-io/balena-sdk/pull/1167

PR #1203, opened 2022-03-10
patch: Add trying SDK in the browser

vipulgupta2048:
Used partials from HandlebarJS to add a note about using SDK in the browser. 
Since, the docs iterate through all orphans in order to generate the docs. This note that I added was being generated twice. 
I had to do some bash trickery to remove the note that got generated at the end of the file which you will find in package.json
Also, not sure how these many changes are creeping into the documentation. I am thinking we haven't run this command in a while?

Impact:
This should be pulled by the main docs. Business as usual. 

Signed-off-by: Vipul Gupta (@vipulgupta2048) <vipul@balena.io>
Resolves: https://github.com/balena-io/docs/issues/2205

### DIFF
--- a/DOCUMENTATION.md
+++ b/DOCUMENTATION.md
@@ -1,9 +1,59 @@
-Welcome to the Balena SDK documentation.
+Welcome to the Balena SDK documentation. This document aims to describe all the functions supported by the SDK, as well as showing examples of their expected usage.
 
-This document aims to describe all the functions supported by the SDK, as well as showing examples of their expected usage.
+## Installation
+
+Install the balena SDK by running:
+
+```sh
+$ npm install --save balena-sdk
+```
+
+### Bundling for browsers
+
+If you're using webpack, browserify, or a similar tool then you probably want to bundle the balena SDK into your application yourself, rather than using the pre-built `balena-browser.js` bundle. 
+
+If you intend to do that, be sure to remove the following dependencies that are actually unnecessary in the browser, because they're only used in Node environments. This will significantly reduce the size of your resulting bundle:
+
+* fs
+* path
+* balena-settings-client
+* node-localstorage
+
+### Bundling with pkg
+
+The balena SDK includes builds for various ECMAScript versions that are
+dynamically selected at runtime (using 
+[@balena/es-version](https://github.com/balena-io-modules/balena-es-version)).
+For this reason, packagers like [pkg](https://github.com/vercel/pkg) are not
+able to automatically detect which assets to include in the output package. The
+following sample `pkg` section should be added to your application's
+`package.json` file to instruct `pkg` to bundle the required assets:
+
+```json
+  "pkg": {
+    "scripts": [
+      "node_modules/balena-sdk/**/*.js"
+    ],
+    "assets": [
+      "node_modules/pinejs-client-core/**/*"
+    ]
+  }
+```
+
+For more information, please refer to the respective
+[documentation from the `pkg` project](https://github.com/vercel/pkg#config).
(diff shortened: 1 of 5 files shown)


vipulgupta2048:
@balena-ci rebase

DOCUMENTATION.md
```diff
@@ -93,6 +93,14 @@ same defaults as the main SDK factory function.
 import { fromSharedOptions } from 'balena-sdk';
 const sdk = fromSharedOptions();
 ```
+
+## Trying balenaSDK in the browser
```
thgreasi, line 97: It feels weird that this appears in between the documentation for the exported methods & the methods for the balena instance.
How about just using the pre-existing DOCUMENTATION.hbs and just add this section right above `{{>main}}` for simplicity? The current content can also be changed to better match the python sdk docs page
See: https://www.balena.io/docs/reference/sdk/python-sdk/

IE copy over the `## Installation` & the unpkg.com sections from the README.

DOCUMENTATION.md
```diff
@@ -107,94 +115,94 @@ const sdk = fromSharedOptions();
     * [.models](#balena.models) : <code>object</code>
         * [.application](#balena.models.application) : <code>object</code>
             * [.tags](#balena.models.application.tags) : <code>object</code>
-                * [.getAllByApplication(slugOrUuidOrId, [options])](#balena.models.application.tags.getAllByApplication) ⇒ <code>Promise</code>
+                * [.getAllByApplication(slugOrId, [options])](#balena.models.application.tags.getAllByApplication) ⇒ <code>Promise</code>
```
thgreasi, line 118: I think that you need to run the full `npm run build`, since the docs are generated from the jsdocs found in the unminified js build output.
I don't get any diff locally on master after an `npm run build`.

DOCUMENTATION.md
vipulgupta2048: We thought about that and it looked weird to me on first thought. Check out our same conversation here: https://github.com/balena-io/docs/issues/2205

 IMHO ^^ this was waaay simpler for me to do then go hacking around the partials.

DOCUMENTATION.md
vipulgupta2048: Ahh yes you are right

vipulgupta2048:
@balena-ci rebase

vipulgupta2048:
Good to go @thgreasi

DOCUMENTATION.md
```diff
@@ -1,10 +1,60 @@
-Welcome to the Balena SDK documentation.
+Welcome to the Balena SDK documentation. This document aims to describe all the functions supported by the SDK, as well as showing examples of their expected usage.
+
+## Installation
+
+Install the balena SDK by running:
+
+```sh
+$ npm install --save balena-sdk
+```
```
thgreasi, line 9: Would be nice to have an `### npm` & `### CDN` subsection in the `## Installation`, to mention unpkg like in the README:
> The balena SDK also provides a bundled single file for browsers, which allows you to include a single file with all dependencies included, available as [balena-browser.min.js](https://unpkg.com/balena-sdk/es2015/balena-browser.min.js) (or [balena-browser.js](https://unpkg.com/balena-sdk/es2015/balena-browser.js) if you'd like the much larger unminified version). This uses the [UMD format](https://github.com/umdjs/umd), and will register itself as either a CommonJS or AMD module called `balena-sdk` if possible, or create a `balenaSdk` global if not.

thgreasi (approved):
:+1: :+1: :+1:

PR #1204, opened 2022-03-10
device.getWithServiceDetails: Add the release id in the service info

myarmolinsky:
device.getWithServiceDetails: Add the release id in the service info

Change-type: minor
Signed-off-by: Matthew Yarmolinsky <matthew-timothy@balena.io>

---
##### Contributor checklist

- [ ] Includes tests
- [ ] Includes typings

### DIFF
--- a/lib/util/device-service-details.ts
+++ b/lib/util/device-service-details.ts
@@ -19,6 +19,7 @@ export interface CurrentService {
 
 export interface CurrentServiceWithCommit extends CurrentService {
 	commit: string;
+	release_id: number;
 }
 
 export interface CurrentGatewayDownload {
@@ -70,7 +71,7 @@ export const getCurrentServiceDetailsPineExpand = (expandRelease: boolean) => {
 				},
 				...(expandRelease && {
 					is_provided_by__release: {
-						$select: ['commit'],
+						$select: ['id', 'commit'],
 					},
 				}),
 			},
@@ -114,7 +115,7 @@ function getSingleInstallSummary(
 	const image = (rawData.image as Image[])[0];
 	const service = (image.is_a_build_of__service as Service[])[0];
 
-	let releaseInfo: { commit?: string } = {};
+	let releaseInfo: { commit?: string; release_id?: number } = {};
 	if (
 		'is_provided_by__release' in rawData &&
 		rawData.is_provided_by__release != null
@@ -122,6 +123,7 @@ function getSingleInstallSummary(
 		const release = (rawData.is_provided_by__release as Release[])[0];
 		releaseInfo = {
 			commit: release != null ? release.commit : undefined,
+			release_id: release != null ? release.id : undefined,
 		};
 	}
 
--- a/tests/integration/models/device.spec.js
+++ b/tests/integration/models/device.spec.js
@@ -2261,6 +2261,12 @@ describe('Device Model', function () {
 					// Augmented properties
 					// Should filter out deleted image installs
 					expect(deviceDetails.current_services.db).to.have.lengthOf(1);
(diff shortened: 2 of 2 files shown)


lib/util/device-service-details.ts
```diff
@@ -19,6 +19,7 @@ export interface CurrentService {
 
 export interface CurrentServiceWithCommit extends CurrentService {
 	commit: string;
+	commitId: number;
```
thgreasi, line 22:
```suggestion
	release_id: number;
```
So that it's clear what id this is for.

thgreasi (commented):
Let's also change the commit to something like:
device.getWithServiceDetails: Add the release id in the returned service info

lib/util/device-service-details.ts
```diff
@@ -19,6 +19,7 @@ export interface CurrentService {
 
 export interface CurrentServiceWithCommit extends CurrentService {
 	commit: string;
+	release_id: number;
```
thgreasi, line 22: Would be nice extend the current `getWithServiceDetails` test case to also check for this new field.

thgreasi:
BTW the commit message doesn't fit @myarmolinsky and needs some rephrasing

lib/util/device-service-details.ts
myarmolinsky: I've added a test to check that both `commit` and `release_id` exist on the services retrieved

tests/integration/models/device.spec.js
```diff
@@ -2261,6 +2261,12 @@ describe('Device Model', function () {
 					// Augmented properties
 					// Should filter out deleted image installs
 					expect(deviceDetails.current_services.db).to.have.lengthOf(1);
+					const currentServices =
+						_.flatten(Object.values(deviceDetails.current_services)) ?? [];
+					currentServices.forEach((currentService) => {
+						expect(currentService).to.have.property('commit');
+						expect(currentService).to.have.property('release_id');
```
thgreasi, line 2268: You can actually also add this a few lines above after the `expect(imageInstall).to.not.have.property('commit');` since the `current_services` is just a dictionary of the image installs :)

thgreasi (commented):
LGTM (self-certify).
Would be nice if you could make the commit message fit though b/c atm it's
> device.getWithServiceDetails: Add the release id in the returned serv…

Maybe something as simple as dropping the `returned`, like:
> device.getWithServiceDetails: Add the release id in the service info

tests/integration/models/device.spec.js
myarmolinsky: Do you mean right after the `forEach` ends? Or *inside* the forEach before it ends?

thgreasi: I mean on line 2256
https://github.com/balena-io/balena-sdk/pull/1204/files#diff-64d3e0fa4d8bb01064ac099bacb99a8bff3c732683a5931c7edd36e7a401bb8eR2256
image

thgreasi (approved):

tests/integration/models/device.spec.js
myarmolinsky: Yes but I do not understand why we can put these lines right after that line. If I put them there as is, then that check will run for each imageInstall. If you meant I should put the checks there alter them a bit, what kind of alteration would have been necessary? Because `imageInstall` does not immediately have the `release_id` property on it so I can't just do `expect(imageInstall).to.have.property('release_id')`
I'm very likely missing something 😅

thgreasi: Nevermind that's already fine anyway :+1:

PR #1243, opened 2022-05-31
Deprecate the needsPasswordReset field of the JWTUser

thgreasi:
Change-type: patch
See: https://github.com/balena-io/balena-api/pull/3665


thgreasi:
@balena-ci I self-certify!

Issue #1255, opened 2022-08-02
Should add support for batch operations in the application model


thgreasi:
This originally came up b/c many individual application delete requests cause the usage based ratelimiter to kick in.

See: https://www.flowdock.com/app/rulemotion/resin-tech/threads/fJCpffcn2oM7dVOnaiWu-d8gCyi

PR #1257, opened 2022-08-03
application.remove: Support batch deletions by providing multiple IDs

thgreasi:
Resolves: #1255
Change-type: minor

---
##### Contributor checklist

- [ ] Includes tests
- [ ] Includes typings

### DIFF
--- a/lib/models/application.ts
+++ b/lib/models/application.ts
@@ -134,6 +134,16 @@ const getApplicationModel = function (
 		},
 	);
 
+	const batchApplicationOperation = once(() =>
+		(
+			require('../util/request-batching') as typeof import('../util/request-batching')
+		).batchResourceOperationFactory<Application>({
+			getAll: exports.getAll,
+			NotFoundError: errors.BalenaApplicationNotFound,
+			AmbiguousResourceError: errors.BalenaAmbiguousApplication,
+		}),
+	);
+
 	// Infer dashboardUrl from apiUrl if former is undefined
 	const dashboardUrl = opts.dashboardUrl ?? apiUrl!.replace(/api/, 'dashboard');
 
@@ -798,7 +808,7 @@ const getApplicationModel = function (
 		 * @function
 		 * @memberof balena.models.application
 		 *
-		 * @param {String|Number} slugOrUuidOrId - application slug (string), uuid (string) or id (number)
+		 * @param {String|Number|Number[]} slugOrUuidOrIdOrIds - application slug (string), uuid (string) or id (number) or array of ids
 		 * @returns {Promise}
 		 *
 		 * @example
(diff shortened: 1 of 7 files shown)


lib/util/request-batching.ts
```diff
@@ -0,0 +1,139 @@
...
+		}
+
+		// create a list of UUIDs or chunks of IDs
+		const chunks: Array<string | number[]> =
+			typeof uuidOrIdOrIds === 'string'
+				? [uuidOrIdOrIds]
+				: chunk(
+						!Array.isArray(uuidOrIdOrIds) ? [uuidOrIdOrIds] : uuidOrIdOrIds,
+						CHUNK_SIZE,
+				  );
```
Page-, line 74: I think you can move the `!Array.isArray` outside of the chunks section, since it'll be an array of 1 anyway and you can't exactly chunk it any smaller:
```suggestion
			typeof uuidOrIdOrIds === 'string' || !Array.isArray(uuidOrIdOrIds)
				? [uuidOrIdOrIds]
				: chunk(
						 uuidOrIdOrIds,
						CHUNK_SIZE,
				  );
```

Page-: Oh actually I see you're doing some magic around an array vs array of arrays below so you'd need
```suggestion
			typeof uuidOrIdOrIds === 'string'
				? [uuidOrIdOrIds]
				:  !Array.isArray(uuidOrIdOrIds)
				? [[uuidOrIdOrIds]]
				: chunk(
						 uuidOrIdOrIds,
						CHUNK_SIZE,
				  );
```
in which case maybe the existing version is better :/

lib/util/request-batching.ts
```diff
@@ -0,0 +1,139 @@
...
+				: {
+						uuid: { $startswith: uuidOrIdOrIdsChunk },
+				  };
+			const combinedOptions = mergePineOptions(
+				{
+					$select: [
+						'id',
+						...(groupByNavigationPoperty ? [groupByNavigationPoperty] : []),
+					] as Array<PineSelectableProps<T>>,
+					// @ts-expect-error
```
Page-, line 96: I like to add a little blurb as to why we expect the error as it can be pretty helpful later

lib/models/application.ts
```diff
@@ -812,19 +822,37 @@ const getApplicationModel = function (
...
+		remove: async (
+			slugOrUuidOrIdOrIds: string | number | number[],
+		): Promise<void> => {
+			if (typeof slugOrUuidOrIdOrIds === 'string') {
+				try {
+					const applicationId = await getId(slugOrUuidOrIdOrIds);
+					await pine.delete({
+						resource: 'application',
+						id: applicationId,
+					});
```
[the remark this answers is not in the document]
thgreasi: I was also thinking about that, but in order to do that we have to make the batchOperations helper to also handle `slug`s (or any unique prop really). But I just decided that this was already a decent PR to go in and left it for later.

lib/util/request-batching.ts
thgreasi: Exactly... I hope to untangle it a bit once I also add support for slugs.

lib/util/request-batching.ts
thgreasi: It's b/c `PineFilter<{ id: number; uuid: number; }>` is not assignable to `PineFilter<T>` :(

thgreasi:
@balena-ci I self-certify!

PR #1275, opened 2022-11-04
Use deep imports for date-fns to improve tree-shaking

thgreasi:
Change-type: patch
See: https://balena.zulipchat.com/#narrow/stream/346007-balena-io.2FbalenaCloud/topic/moment.20-.3E.20date-fns

Resolves: # 
HQ: <url> 
See: <url> 
Depends-on: <url> 

---
##### Contributor checklist

- [ ] Includes tests
- [ ] Includes typings

### DIFF
--- a/lib/util/date.ts
+++ b/lib/util/date.ts
@@ -1,5 +1,5 @@
 import { formatDistance } from 'date-fns';
-import * as memoizee from 'memoizee';
+import memoizee from 'memoizee';
 
 const now = memoizee(() => new Date(), { maxAge: 1000 });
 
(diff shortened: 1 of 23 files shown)


lib/util/date.ts
```diff
@@ -1,4 +1,6 @@
-import { formatDistance } from 'date-fns';
+// TODO: Remvove the deep imports once we start offering an esm build
+// and confirm that tree-shaking date-fns still works.
```
thgreasi, line 2: Actually, I think that the CLI would benefit from a deep import, since it will always have to parse less files, regardless of switching to esm or not :thinking:

thgreasi:
@balena-ci I self-certify!

PR #1282, opened 2022-11-22
Add utils and export mergePineOptions `balena.utils.mergePineOptions()`

JSReds:
Change-type: minor

Resolves: # 
HQ: <url> 
See: <url> 
Depends-on: <url> 

---
##### Contributor checklist

- [ ] Includes tests
- [ ] Includes typings

### DIFF
--- a/lib/index.ts
+++ b/lib/index.ts
@@ -174,6 +174,7 @@ export type BalenaSDK = {
 	interceptors: Interceptor[];
 	request: BalenaRequest;
 	pine: Pine;
+	utils: import('./util').BalenaUtils;
 	errors: typeof import('balena-errors');
 	version: string;
 };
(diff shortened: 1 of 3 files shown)


lib/index.ts
```diff
@@ -396,6 +406,24 @@ export const getSdk = function ($opts?: SdkOptions) {
...
+	 * The utils instance used internally. This should not be necessary
+	 * in normal usage, but can be useful to handle some specific cases.
+	 *
+	 * @example
+	 * balena.utils.mergePineOptions(
+	 *  { $expand: { device: { $select: ['id'] } } },
+	 *  { $expand: { device: { $select: ['name'] } } },
+	 * );
+	 */
+	sdk.utils = utils;
```
thgreasi, line 425: Given that `util/index.ts` loads all utils that the SDK has, maybe it would make sense lazy load this property like we do for the models. Otherwise the individual file loading would slow down the CLI's startup time.
```suggestion
		Object.defineProperty(sdk, 'utils', {
			enumerable: true,
			configurable: true,
			get() {
				const { mergePineOptions } = require('./util') as typeof import('./util');
				return { mergePineOptions };
			},
		});
```
On top of (but we can postpone for later) we could make `util/index.ts` only include the exported functions and have the SDK source (& tests) do exact path imports for anything they need.

PR #1291, opened 2023-01-03
add methods necessary for preloading

aethernet:
- Get "Fleet level Target State v3" on devices
- Get "supervisor image for version and dt" on os
- Util to list images name from a target state

Change-type: patch
Signed-off-by: Edwin Joassart <edwin.joassart@balena.io>

See: 
- [Zulip preloading thread](https://balena.zulipchat.com/#narrow/stream/346008-balena-io.2Fetcher/topic/.2Eetch.20.2F.20preloading.20status.20.28and.20some.20more.29/near/318093854)

Depends-on: https://github.com/balena-io/open-balena-api/pull/1081
Change-type: patch

---
##### Contributor checklist
- [x] Includes tests
- [x] Includes typings
- [/] Includes updated documentation -> should be handled by CI from jsdoc
- [/] Includes updated build output -> should be handled by CI

### DIFF
--- a/src/models/device.ts
+++ b/src/models/device.ts
@@ -1248,6 +1248,7 @@ const getDeviceModel = function (
 		 * @memberof balena.models.device
 		 *
 		 * @param {String|Number} uuidOrId - device uuid (string) or id (number)
+		 * @param {Number} version - (optional) target state version (2 or 3), default to 2
 		 * @returns {Promise}
 		 *
 		 * @example
@@ -1261,6 +1262,11 @@ const getDeviceModel = function (
 		 * });
 		 *
 		 * @example
+		 * balena.models.device.getSupervisorTargetState(123, 3).then(function(state) {
+		 * 	console.log(state);
+		 * });
+		 *
+		 * @example
 		 * balena.models.device.getSupervisorTargetState('7cf02a6', function(error, state) {
 		 * 	if (error) throw error;
 		 * 	console.log(state);
@@ -1268,10 +1274,52 @@ const getDeviceModel = function (
 		 */
 		getSupervisorTargetState: async (
 			uuidOrId: string | number,
+			version: 2 | 3 = 2,
 		): Promise<DeviceState.DeviceState> => {
 			const { uuid } = await exports.get(uuidOrId, { $select: 'uuid' });
 			const { body } = await request.send({
-				url: `/device/v2/${uuid}/state`,
+				url: `/device/v${version}/${uuid}/state`,
+				baseUrl: apiUrl,
+			});
+			return body;
+		},
+
+		/**
+		 * @summary Get the target supervisor state on a "generic" device on a fleet
+		 * @name getSupervisorTargetStateForApp
+		 * @public
+		 * @function
+		 * @memberof balena.models.device
+		 *
+		 * @param {String|Number} uuidOrId - fleet uuid (string) or id (number)
+		 * @param {String} release - (optional) release uuid (default tracked)
+		 * @returns {Promise}
+		 *
+		 * @example
+		 * balena.models.device.getSupervisorTargetStateForApp('7cf02a6').then(function(state) {
+		 * 	console.log(state);
+		 * });
+		 *
+		 * @example
+		 * balena.models.device.getSupervisorTargetStateForApp(123).then(function(state) {
+		 * 	console.log(state);
+		 * });
+		 *
+		 * @example
+		 * balena.models.device.getSupervisorTargetStateForApp(123, '7cf02a6').then(function(state) {
+		 * 	console.log(state);
+		 * });
+		 *
+		 */
+		getSupervisorTargetStateForApp: async (
+			slugOrUuidOrId: string | number,
+			release?: string | number,
+		): Promise<DeviceState.DeviceStateV3> => {
+			const { uuid } = await applicationModel().get(slugOrUuidOrId, {
+				$select: 'uuid',
+			});
+			const { body } = await request.send({
+				url: `/device/v3/fleet/${uuid}/state/?releaseUuid=${release ?? ''}`,
 				baseUrl: apiUrl,
 			});
 			return body;
--- a/src/util/device.ts
+++ b/src/util/device.ts
@@ -1,4 +1,5 @@
 import type * as BalenaSdk from '..';
+import type * as DeviceState from '../types/device-state';
 
 export const isProvisioned = (
 	device: Partial<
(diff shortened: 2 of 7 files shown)


lib/models/device.ts
```diff
@@ -1288,17 +1289,64 @@ const getDeviceModel = function (
...
+		 * 	console.log(state);
+		 * });
+		 *
+		 * @example
+		 * balena.models.device.getSupervisorTargetStateForFleet(123, '7cf02a6').then(function(state) {
+		 * 	console.log(state);
+		 * });
+		 *
+		 */
+		getSupervisorTargetStateForApp: async (
```
thgreasi, line 1341: Since that's more related to the app rather than the device, how about moving it under the application model like: `balena.models.application.getSupervisorTargetState`?

thgreasi (commented):
Let's also make clear in the commit title that this is about application target state, like:
> Add a method for retrieving the application's target supervisor state

or similar

lib/models/device.ts
aethernet: @thgreasi 
On the API it's attached to the device endpoint as it represent the "virtual target state" of a "future" device on fleet.

For coherence between the two, I'd prefer to keep it on device.

lib/models/os.ts
```diff
@@ -957,6 +957,32 @@ const getOsModel = function (
...
+	 *
+	 */
+	const getSupervisorImageForArch = async (
+		architecture: string,
+		version: string,
+	): Promise<string | null> => {
+		const { body } = await request.send({
+			url: `/v6/supervisor_release?\$top=1&\$select=image_name&\$filter=(supervisor_version%20eq%20%27${version}%27)%20and%20(is_for__device_type/any(ifdt:ifdt/is_of__cpu_architecture/any(ioca:ioca/slug%20eq%20%27${architecture}%27)))`,
+			baseUrl: apiUrl,
+		});
```
thgreasi, line 982: Reminder to change this to a proper `pine.get()`.

aethernet:
Note: While waiting for https://github.com/balena-io/open-balena-api/pull/1081 I'm skipping the test for that endpoint so I can use the draft is published to npm and I can use it for dev purposes.

lib/models/os.ts
aethernet: done

aethernet:
@thgreasi It's ready to be reviewed :)

lib/models/os.ts
```diff
@@ -957,6 +962,41 @@ const getOsModel = function (
...
+	 * @param {Number} deviceTypeId - The Id for the Device Type
+	 * @param {String} version - The semver version string for the supervisor
+	 * @returns {Promise<String>} - Docker image name for the Supervisor version and arch
+	 *
+	 * @example
+	 * const result = balena.models.os.getSupervisorImageForDT(60, 'v12.11.0').then(result => console.log(result))
+	 * // 60 would be raspberrypi4-64 on balena-cloud
+	 *
+	 */
+	const getSupervisorImageForDT = async (
```
thgreasi, line 981: NIT
```suggestion
	const getSupervisorReleaseByDeviceType = async (
```
which should return the whole record and consumers can grab the part they need.
Keep in mind that this will not include invalidated or draft releases.

aethernet: There's not much more to it, but I get your point, pushing a fix.

lib/models/device.ts
```diff
@@ -1288,17 +1289,64 @@ const getDeviceModel = function (
...
+		 *
+		 * @example
 		 * balena.models.device.getSupervisorTargetState('7cf02a6', function(error, state) {
 		 * 	if (error) throw error;
 		 * 	console.log(state);
 		 * });
 		 */
 		getSupervisorTargetState: async (
 			uuidOrId: string | number,
+			version?: number,
```
dfunckt, line 1304:
```suggestion
			version: 2 | 3 = 2
```

Arguably though, the correct thing to do is default to latest but I don’t know the implications

lib/models/device.ts
```diff
@@ -1288,17 +1289,64 @@ const getDeviceModel = function (
...
 		): Promise<DeviceState.DeviceState> => {
 			const { uuid } = await exports.get(uuidOrId, { $select: 'uuid' });
 			const { body } = await request.send({
-				url: `/device/v2/${uuid}/state`,
+				url: `/device/v${version !== 3 ? 2 : 3}/${uuid}/state`,
```
dfunckt, line 1308:
```suggestion
				url: `/device/v${version}/${uuid}/state`,
```

lib/models/device.ts
aethernet: agreed, but as they are not the same at all, I prefer to play it safe.

aethernet:
with open-balena-api 7.13.0 the missing endpoint are now in. 
but we still need for the update then release to prod env for tests to pass https://github.com/balena-io/balena-api/pull/4114

aethernet:
The previously missing endpoint is now in production, so I turn the tests back on and from my POV this is ready for a last review then merge.

vipulgupta2048:
Very interesting @aethernet will this PR be a step in supporting preloading using the balena-SDK instead of relying on balenaCLI. If yes, then absolutely the best thing I have seen happen. 
cc: @rcooke-warwick

aethernet:
I Self-Certify!

src/util/device.ts
```diff
@@ -11,3 +12,33 @@ export const isProvisioned = (
 		device.last_connectivity_event != null
 	);
 };
+
+/**
+ *
+ * @param targetState
+ * @returns array containing all images for all services for all releases for all apps for the device
+ */
+export const listImagesFromTargetState = (
```
thgreasi, line 21: If this is not used by the SDK at all, why do we have it here? Which are the clients will be consuming this?

src/models/device.ts
```diff
@@ -1261,17 +1262,64 @@ const getDeviceModel = function (
...
+		 *
+		 * @example
 		 * balena.models.device.getSupervisorTargetState('7cf02a6', function(error, state) {
 		 * 	if (error) throw error;
 		 * 	console.log(state);
 		 * });
 		 */
 		getSupervisorTargetState: async (
 			uuidOrId: string | number,
+			version: 2 | 3 = 2,
```
thgreasi, line 1277: Doesn't this need an overload in order to have the correct return type when using v3?

src/util/device.ts
aethernet: `balena-img` will consume this

thgreasi: If it is only `balena-img` will be consuming this, then it doesn't need to be part of the sdk's exposed API, and moving it closer to/in the consumer imo makes more sense.

But TBH even if the CLI would also need to use this, imo it's just 4 nested for loops which is trivial to inline.
Btw
See: https://github.com/balena-io/balena-sdk/pull/1347

thgreasi: In other words, I would imagine adding a comment like `// don't remove this b/c the img-maker needs it` would not look great. For example maybe preload could have such a helper or even better use it internally for its own needs and expose a higher level method.

PR #1304, opened 2023-02-21
lib/models: Add usbMassStorage protocol for Radxa like boards

acostach:
The Radxa boards use different flashing mechanisms depending on the device model. Let's add a common
step for them, which is entering maskrom mode. The actual procedure for entering this mode is detailed in the contract partials for each device.

Change-type: patch

See: https://github.com/balena-io/contracts/pull/363
Change-type: patch

---
##### Contributor checklist

- [ ] Includes tests
- [ ] Includes typings

### DIFF
--- a/lib/models/balenaos-contract.ts
+++ b/lib/models/balenaos-contract.ts
@@ -44,8 +44,8 @@ const BalenaOS: Contract = {
 			`Wait for writing of {{name}} to complete.`,
 			`{{{deviceType.partials.bootDevice}}} to boot the device.`,
 		],
-		radxaFlash: [
-			`Enable maskrom mode and connect the device to your computer.`,
+		usbMassStorage: [
+			`{{#each deviceType.partials.instructions}}{{{this}}} {{/each}}`,
 		],
 		edisonFlash: {
 			Linux: [
--- a/lib/models/balenaos-contract.ts
+++ b/lib/models/balenaos-contract.ts
@@ -44,6 +44,9 @@ const BalenaOS: Contract = {
 			`Wait for writing of {{name}} to complete.`,
 			`{{{deviceType.partials.bootDevice}}} to boot the device.`,
 		],
+		radxaFlash: [
+			`Enable maskrom mode and connect the device to your computer.`,
+		],
 		edisonFlash: {
 			Linux: [
 				`{{#each deviceType.partials.Linux.flashDependencies}}{{{this}}} {{/each}}`,
--- a/tests/integration/models/device-type.spec.ts
+++ b/tests/integration/models/device-type.spec.ts
@@ -4,11 +4,14 @@ import { balena, givenLoggedInUser } from '../setup';
 import { timeSuite } from '../../util';
 import type * as BalenaSdk from '../../..';
 
-const DEVICE_TYPE_NAME = 'Raspberry Pi 2';
-const DEVICE_TYPE_SLUG = 'raspberry-pi2';
-const DEVICE_TYPE_ALIAS = 'raspberrypi2';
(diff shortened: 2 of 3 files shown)


lib/models/balenaos-contract.ts
```diff
@@ -1,12 +1,12 @@
-import { Contract } from '../types/contract';
+import { Contract } from "../types/contract";
```
thgreasi, line 1: The quotes would need to be reverted to single ones, otherwise the linter will complain :)

lib/models/balenaos-contract.ts
```diff
@@ -44,6 +44,9 @@ const BalenaOS: Contract = {
 			`Wait for writing of {{name}} to complete.`,
 			`{{{deviceType.partials.bootDevice}}} to boot the device.`,
 		],
+		radxaFlash: [
```
thgreasi, line 47: Can you also add a test case at
See: https://github.com/balena-io/balena-sdk/blob/master/tests/integration/models/device-type.spec.ts#L112-L128

so that we can see how this works for such a device type?

lib/models/balenaos-contract.ts
acostach: Sure, done, thanks for letting me know. Prettify caused this and I later found out the hard way that balena-lint with --fix needs to be used. I've added that line to the README.

lib/models/balenaos-contract.ts
acostach: I added one now but it will be necessary for https://github.com/balena-io/contracts/pull/363 to be merged first, right @vipulgupta2048 ?

vipulgupta2048: Yeah, we need to change the implementation of a method first as well for these instructions to be used. Making a commit now

vipulgupta2048:
@thgreasi It seems that the tests are failing due to no credentials being found? What happened over here?

tests/integration/models/device-type.spec.ts
```diff
@@ -4,11 +4,14 @@ import { balena, givenLoggedInUser } from '../setup';
...
-const DEVICE_TYPE_INSTALL_METHOD = 'externalBoot';
-const DEVICE_TYPE_ID = 1;
+const RPI2_DEVICE_TYPE_NAME = 'Raspberry Pi 2';
+const RPI2_DEVICE_TYPE_SLUG = 'raspberry-pi2';
+const RPI2_DEVICE_TYPE_ALIAS = 'raspberrypi2';
+const RPI2_DEVICE_TYPE_INSTALL_METHOD = 'externalBoot';
+const RPI2_DEVICE_TYPE_ID = 1;
+
+const RADXA_ZERO_DEVICE_TYPE_SLUG = 'radxa-zero';
+const RADXA_ZERO_DEVICE_TYPE_INSTALL_METHOD = 'radxaFlash';
```
thgreasi, line 14: @acostach @vipulgupta2048 given v
See: https://github.com/balena-io/contracts/pull/366
and that the tests atm fail w/ a related complain, should you change the expected installation method to be `usbMassStorage`?


>       balena.models.deviceType.getInstallMethod()
>         should get device type installation method:
>
>      AssertionError: expected 'usbMassStorage' to equal 'radxaFlash'
>      + expected - actual
>
>      -usbMassStorage
>      +radxaFlash

thgreasi: Feel free to self-certify once you fix the tests :+1:

vipulgupta2048: I see the issue, fixing

vipulgupta2048:
@balena-ci I self-certify!

acostach:
@balena-ci I self-certify!

PR #1314, opened 2023-03-24
Add `device history` model

fisehara:
Add `device history` model

`device history` is read only and model is added with
- getAllByDevice
- getAllByApplication

Changelog-entry: Enable audit logging and reviewing device changes. The SDK defaults to fetch last 7 days of history

Change-type: minor

### DIFF
--- a/lib/models/device.ts
+++ b/lib/models/device.ts
@@ -19,6 +19,7 @@ import type {
 	InjectedDependenciesParam,
 	PineOptions,
 	PineTypedResult,
+	PineFilter,
 } from '..';
 import type {
 	Device,
@@ -26,6 +27,7 @@ import type {
 	DeviceVariable,
 	DeviceTag,
 	Application,
+	DeviceHistory,
 } from '../types/models';
 import { DeviceOverallStatus as OverallStatus } from '../types/device-overall-status';
 import type * as DeviceState from '../types/device-state';
@@ -48,6 +50,7 @@ import {
 	isId,
 	isNoDeviceForKeyResponse,
 	isNotFoundResponse,
+	isFullUuid,
 	mergePineOptions,
 	treatAsMissingDevice,
 	withSupervisorLockedError,
@@ -83,6 +86,8 @@ import type {
 import type { AtLeast } from '../../typings/utils';
 import type { DeviceType } from '../types/models';
 
+import subDays from 'date-fns/subDays';
+
 const MIN_OS_MC = '2.12.0';
 const OVERRIDE_LOCK_ENV_VAR = 'RESIN_OVERRIDE_LOCK';
 
@@ -102,6 +107,8 @@ export type DeviceMetrics = Pick<
 	| 'cpu_id'
 	| 'is_undervolted'
 >;
+
+const DEFAULT_DAYS_OF_REQUESTED_HISTORY = 7;
 
 const getDeviceModel = function (
 	deps: InjectedDependenciesParam,
@@ -296,6 +303,33 @@ const getDeviceModel = function (
 				});
 			},
 		});
+	};
+
+	const historyTimeRangeFilterWithGuard = (
+		fromDate?: Date,
+		toDate?: Date,
+	): PineFilter<DeviceHistory> => {
+		let fromDateFilter;
+		let toDateFilter;
+		if (fromDate != null) {
+			if (!(fromDate instanceof Date)) {
+				throw new errors.BalenaInvalidParameterError('fromDate', fromDate);
+			}
+			fromDateFilter = { $ge: fromDate };
+		}
+
+		if (toDate != null) {
+			if (!(toDate instanceof Date)) {
+				throw new errors.BalenaInvalidParameterError('toDate', toDate);
+			}
+			toDateFilter = { $le: toDate };
+		}
+
+		const timeRangeFilter =
+			fromDateFilter || toDateFilter
+				? { created_at: { ...fromDateFilter, ...toDateFilter } }
+				: {};
+		return timeRangeFilter;
 	};
 
 	const exports = {
@@ -521,9 +555,7 @@ const getDeviceModel = function (
 			}
 
 			let device;
-			const isPotentiallyFullUuid =
-				typeof uuidOrId === 'string' &&
-				(uuidOrId.length === 32 || uuidOrId.length === 62);
+			const isPotentiallyFullUuid = isFullUuid(uuidOrId);
 			if (isPotentiallyFullUuid || isId(uuidOrId)) {
 				device = await pine.get({
 					resource: 'device',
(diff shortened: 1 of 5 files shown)


lib/models/device.ts
```diff
@@ -3436,6 +3436,125 @@ const getDeviceModel = function (
...
+			 * });
+			 */
+			async getAllByApplication(
+				slugOrUuidOrId: string | number,
+				options?: PineOptions<DeviceHistory>,
+			): Promise<DeviceHistory[]> {
+				if (options == null) {
+					options = {};
+				}
+
```
Page-, line 3537: There's no need to default this at all since `mergePineOptions` is fine with nullish being passed in (and in fact can short circuit in that case to avoid work)
```suggestion
```

lib/models/device.ts
```diff
@@ -3436,6 +3436,125 @@ const getDeviceModel = function (
...
+			 * });
+			 */
+			async getAllByDevice(
+				uuidOrId: string | number,
+				options?: PineOptions<DeviceHistory>,
+			): Promise<DeviceHistory[]> {
+				if (options == null) {
+					options = {};
+				}
+
```
Page-, line 3480:
```suggestion
```

lib/models/device.ts
```diff
@@ -103,6 +107,8 @@ export type DeviceMetrics = Pick<
 	| 'is_undervolted'
 >;
 
+const defaultDaysOfRequestedHistory = 7;
```
myarmolinsky, line 110: nit
```suggestion
const DEFAULT_DAYS_OF_REQUESTED_HISTORY = 7;
```

lib/util/index.ts
```diff
@@ -19,7 +19,9 @@ export const onlyIf =
 		}
 	};
 
-export const isId = (v?: any): v is number => typeof v === 'number';
+export const isId = (v?: unknown): v is number => typeof v === 'number';
+export const isUuid = (v?: unknown): v is string =>
```
thgreasi, line 23:
```suggestion
export const isFullUuid = (v?: unknown): v is string =>
```

lib/models/device.ts
fisehara: Thanks!

lib/models/device.ts
```diff
@@ -3436,6 +3440,140 @@ const getDeviceModel = function (
...
+					created_at: {
+						$ge: fromDate,
+						...(toDate ? { $le: toDate } : null),
+					},
+				};
+
+				if (isId(uuidOrId)) {
+					$filter = { ...$filter, ...{ tracks__device: uuidOrId } };
+				} else if (isUuid(uuidOrId)) {
+					$filter = { ...$filter, ...{ uuid: uuidOrId } };
```
thgreasi, line 3498:
```suggestion
					$filter = { ...$filter, tracks__device: uuidOrId} };
				} else if (isUuid(uuidOrId)) {
					$filter = { ...$filter, uuid: uuidOrId };
```

lib/models/device.ts
```diff
@@ -3436,6 +3440,140 @@ const getDeviceModel = function (
...
+				uuidOrId: string | number,
+				{
+					fromDate,
+					toDate,
+					...options
+				}: PineOptions<DeviceHistory> & { fromDate?: Date; toDate?: Date } = {
+					fromDate: subDays(new Date(), defaultDaysOfRequestedHistory),
+				},
+			): Promise<DeviceHistory[]> {
+				let $filter = {
```
thgreasi, line 3488: Can you try doing v :
```suggestion
				const $filter: PineFilter<DeviceHistory> = {
```
and then see whether TS complains if you do `$filter.tracks__device = uuidOrId`.


If that doesn't work, then let's just add the type here anyway even w/ a `let`:
```
let $filter: PineFilter<DeviceHistory>
```

lib/models/device.ts
```diff
@@ -3436,6 +3440,140 @@ const getDeviceModel = function (
...
+					created_at: {
+						$ge: fromDate,
```
thgreasi, line 3490: I think that you still need to check whether `fromDate` has a value or not here since atm it can be null.

lib/models/device.ts
```diff
@@ -3436,6 +3440,140 @@ const getDeviceModel = function (
...
+			 */
+			async getAllByDevice(
+				uuidOrId: string | number,
+				{
+					fromDate,
+					toDate,
+					...options
+				}: PineOptions<DeviceHistory> & { fromDate?: Date; toDate?: Date } = {
+					fromDate: subDays(new Date(), defaultDaysOfRequestedHistory),
+				},
```
thgreasi, line 3486: Just a heads-up that as this is right now `getAllByDevice('uuid', {})` doesn't use a default value for the `fromDate` and fetches ALL history records.

If you intended making fetching all history more explicit eg `getAllByDevice('uuid', { fromDate: null })`, then you need to do:
```suggestion
				{
					fromDate = subDays(new Date(), defaultDaysOfRequestedHistory),
					toDate,
					...options
				}: PineOptions<DeviceHistory> & { fromDate?: Date; toDate?: Date } = {},
```

lib/models/device.ts
```diff
@@ -3436,6 +3440,140 @@ const getDeviceModel = function (
...
+					},
+				);
+
+				return await pine.get<DeviceHistory>({
+					resource: 'device_history',
+					options: mergePineOptions(
+						{
+							$filter: {
+								created_at: {
+									$ge: fromDate,
```
thgreasi, line 3566: I think that you still need to check whether `fromDate` has a value or not here since atm it can be null.

lib/types/models.ts
```diff
@@ -744,3 +744,23 @@ export interface SubscriptionAddonDiscount {
 	discount_percentage: number;
 	discounts__plan_addon: NavigationResource<PlanAddon>;
 }
+
+export interface DeviceHistory {
```
thgreasi, line 748: You should also add a new row in `ResourceTypeMap` like:
```
device_history: DeviceHistory;
```
This is the resource-to-interface map that the SDK typings use to enable fully typed results.

lib/models/device.ts
fisehara: thanks!

lib/models/device.ts
fisehara: done

lib/models/device.ts
fisehara: thanks, moved into the helper `historyTimeRangeFilterWithGuard`

lib/models/device.ts
fisehara: Typescript unfortunately complains and the $filter comes without member knowledge inthe IDE.

lib/models/device.ts
```diff
@@ -298,6 +305,35 @@ const getDeviceModel = function (
...
+		fromDate?: Date,
+		toDate?: Date,
+	): PineFilter<DeviceHistory> => {
+		let fromDateFilter;
+		let toDateFilter;
+		if (fromDate != null) {
+			if (fromDate instanceof Date) {
+				fromDateFilter = { $ge: fromDate };
+			} else {
+				throw new errors.BalenaInvalidParameterError('fromDate', fromDate);
```
thgreasi, line 318: NIT: This could be using an early return pattern like `if (!(fromDate instanceof Date)) { throw ... }`.

thgreasi (approved):

lib/models/device.ts
fisehara: Yeah, will follow the fail early pattern

PR #1329, opened 2023-05-12
Add support to short uuid on device.serviceVar.set

otaviojacobi:
Change-type: patch

Resolves: # 
HQ: <url> 
See: <url> 
Depends-on: <url> 

---
##### Contributor checklist

- [ ] Includes tests
- [ ] Includes typings

### DIFF
--- a/src/models/device.ts
+++ b/src/models/device.ts
@@ -2898,18 +2898,24 @@ const getDeviceModel = function (
 			): Promise<void> {
 				value = String(value);
 
-				const deviceFilter = isId(uuidOrId)
-					? uuidOrId
-					: {
-							$any: {
-								$alias: 'd',
-								$expr: {
-									d: {
-										uuid: uuidOrId,
-									},
+				let deviceFilter;
+				if (isId(uuidOrId)) {
+					deviceFilter = uuidOrId;
+				} else if (isFullUuid(uuidOrId)) {
+					deviceFilter = {
+						$any: {
+							$alias: 'd',
+							$expr: {
+								d: {
+									uuid: uuidOrId,
 								},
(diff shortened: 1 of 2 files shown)


lib/models/device.ts
```diff
@@ -3392,18 +3392,24 @@ const getDeviceModel = function (
...
-				const deviceFilter = isId(uuidOrId)
-					? uuidOrId
-					: {
-							$any: {
-								$alias: 'd',
-								$expr: {
-									d: {
-										uuid: uuidOrId,
-									},
+				let deviceFilter;
```
thgreasi, line 3395: This atm is typed as `any`, so let's make it:
```suggestion
				let deviceFilter: ResourceObjFilter<Device>[keyof ResourceObjFilter<Device>];
```

lib/models/device.ts
thgreasi, line 1: Let's also copy paste the `.set` test case and make it use a short uuid
See: https://github.com/balena-io/balena-sdk/blob/6c3a9cc36e7f0c0b85ea3b373696651d1c9486c0/tests/integration/models/device.spec.js#L2381-L2389

thgreasi (commented):
Other than that LGTM :+1:

lib/models/device.ts
otaviojacobi: nice catch

lib/models/device.ts
thgreasi: @otaviojacobi I would prefer to not monkey patch new props. Let's do the simple thing of adding a new test case especially for testing short uuids.

otaviojacobi:
LGTM

PR #1330, opened 2023-05-15
Add device type yocto properties to typings

otaviojacobi:
Change-type: patch
Depends-on: https://github.com/balena-io/open-balena-api/pull/1351
See: https://github.com/balena-io/balena-cli/pull/2619

Resolves: # 
HQ: <url> 
See: <url> 
Depends-on: <url> 

---
##### Contributor checklist

- [ ] Includes tests
- [ ] Includes typings

### DIFF
--- a/.github/actions/test/action.yml
+++ b/.github/actions/test/action.yml
@@ -41,17 +41,17 @@ runs:
         node -v
         npm -v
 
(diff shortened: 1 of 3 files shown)


thgreasi (approved):
Even though we try to move away from the device-type.json and this is in the opposite direction, I'm fine with it for now, even though I'm afraid anyone would start using them, increasing our dependencies to it.

PR #1331, opened 2023-05-16
Add OS download custom properties

otaviojacobi:
Change-type: minor

Resolves: # 
HQ: <url> 
See: <url> 
Depends-on: <url> 

---
##### Contributor checklist

- [ ] Includes tests
- [x] Includes typings

### DIFF
--- a/src/models/os.ts
+++ b/src/models/os.ts
@@ -107,6 +107,21 @@ export interface OsUpdateVersions {
 	current: string | undefined;
 }
 
+export interface DownloadConfig
+	extends Pick<
+		ImgConfigOptions,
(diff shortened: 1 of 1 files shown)


otaviojacobi:
@thgreasi I didn't really know how/if to test this automatically. I tested manually by passing some parameters to the sdk call and then provisioning an actual device, for example, now if we pass the appId this will make the device provision directly to the fleet and connect to it, which worked.

otaviojacobi:
and FWIW this was already working on the JS version, even in previous one if you just ignored the typings

thgreasi:
:+1:

PR #1337, opened 2023-05-25
Fix auth.twoFactor.isEnabled() regression returning always true

thgreasi:
Update balena-auth from 4.1.0 to 4.2.0

Change-type: patch

Resolves: # 
HQ: <url> 
See: <url> 
Depends-on: <url> 

---
##### Contributor checklist

- [ ] Includes tests
- [ ] Includes typings


thgreasi:
Yes I've already added them @otaviojacobi :thinking: 
I see that the envs were missing from the karma.conf.js, but that doesn't explain why the tests didn't run in node :thinking:

thgreasi:
I self-certify!

PR #1355, opened 2023-06-28
application.create: Enable creating fleets with archived device types 

myarmolinsky:
application.create: Enable creating fleets with archived device types

See: https://balena.zulipchat.com/#narrow/stream/345746-aspect.2Fproduct/topic/balenaFin.20discontinuation
Connects-to: https://github.com/balena-io/balena-ui/pull/6030
Change-type: patch

---
##### Contributor checklist

- [ ] Includes tests
- [ ] Includes typings

### DIFF
--- a/tests/integration/models/application.spec.ts
+++ b/tests/integration/models/application.spec.ts
@@ -118,17 +118,6 @@ describe('Application Model', function () {
 					);
 				});
 
-				it('should be rejected if the device type is discontinued', function () {
-					const promise = balena.models.application.create({
-						name: 'FooBar',
-						deviceType: 'edge',
-						organization: ctx.initialOrg.id,
-					});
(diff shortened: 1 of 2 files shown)


thgreasi (approved):
Go for it, no need to wait to decide whether the API should have the old check.

tests/integration/models/application.spec.ts
```diff
@@ -118,15 +118,13 @@ describe('Application Model', function () {
 					);
 				});
 
-				it('should be rejected if the device type is discontinued', function () {
+				it('should succeed even if the device type is discontinued', function () {
```
thgreasi, line 121: This should no longer be in the `[read operations]` section since it actually manages to create a fleet. We need to move it to the `[mutating operations]`.

thgreasi (changes requested):

PR #1370, opened 2023-08-08
V18 Major

otaviojacobi:
Change-type: major

See: https://balena.zulipchat.com/#narrow/stream/345890-balena-io/topic/node-sdk-v18

---
##### Contributor checklist

- [x] Includes tests
- [ ] Includes typings

### DIFF
--- a/src/auth.ts
+++ b/src/auth.ts
@@ -17,7 +17,7 @@ limitations under the License.
 import * as errors from 'balena-errors';
 import memoizee from 'memoizee';
 import type { InjectedDependenciesParam, InjectedOptionsParam } from '.';
-import { WhoamiResult } from './types/auth';
+import { UserInfo, WhoamiResult } from './types/auth';
 
 const getAuth = function (
 	deps: InjectedDependenciesParam,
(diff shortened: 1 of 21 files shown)


thgreasi:
How about a commit message like:
**BREAKING**: Support non-user API keys in auth.isLoggedIn() & whoami()

Ideally the change in the return type of the whoami could be in a separate commit, but I'm fine even w/o that.

src/auth.ts
```diff
@@ -352,9 +350,13 @@ const getAuth = function (
...
-	async function getEmail(): Promise<string> {
-		const { email } = await getUserDetails();
-		return email;
+	async function getEmail(): Promise<string | null> {
+		const result = await getActorDetails();
+
+		if (result.actorType === 'user') {
+			return result.email;
+		}
+		return null;
```
thgreasi, line 359:
```suggestion
		if (result.actorType !== 'user') {
			throw new Error(`The authentication credentials in use are not of a user`);
		}
		return result.email;
```

src/auth.ts
```diff
@@ -303,8 +297,12 @@ const getAuth = function (
...
 	 */
 	async function getUserId(): Promise<number> {
-		const { id } = await getUserDetails();
-		return id;
+		const actor = await getActorDetails();
+
+		if (actor.actorType === 'user') {
+			return actor.actorTypeId;
+		}
+		throw new Error(`${actor.actorType} key is not associated to any user`);
```
thgreasi, line 305: NIT: Let's follow the early return(/throw) pattern.
```suggestion
		if (actor.actorType !== 'user') {
			throw new Error(`The authentication credentials in use are not of a user`);
		}
		return actor.actorTypeId;
```

src/auth.ts
```diff
@@ -303,8 +297,12 @@ const getAuth = function (
 	 * });
 	 */
 	async function getUserId(): Promise<number> {
```
thgreasi, line 299: I'm wondering whether now that whoami returns the whole result, it would make sense to drop this :thinking: 
Alternatively if the argument could be that this saves us from repeating the `if not user actor then throw` check, then it might make sense to also add a `getUsername` method, but I'm not really enthusiastic about that...

otaviojacobi: @thgreasi Well the good thing about having getUsername is that it would be a very similar replacement to the current `whoami` which can make the transition from v17 -> v18 slight easier. Why you deslike the idea?

thgreasi: That was what I was thinking as well @otaviojacobi .
I'm fine with it just:
* afraid that ppl might just replace whoami w/ getUsername w/o taking the time to think whether there is a better alternative
* not loving the amount of separate special methods that we have for each field of the user whoami.

Let's add it if you are more convinced than me.

thgreasi: @otaviojacobi how about `getUserInfo` or `getUserDetails`?

otaviojacobi: @thgreasi both work for me, I think getUserInfo is slighlt more clear than details

PR #1386, opened 2023-10-09
Add typings for the user_profile resource

thgreasi:
Change-type: minor
Depends-on: https://github.com/balena-io/balena-api/pull/4626


thgreasi:
I self-certify!

PR #1408, opened 2024-03-18
Fix `application.create` method being wrongly marked as deprecated

myarmolinsky:
Fix `application.create` method being wrongly marked as deprecated

The intention was to mark `options.applicationType` as deprecated
Change-type: patch

---
##### Contributor checklist

- [ ] Includes tests
- [ ] Includes typings


vipulgupta2048 (commented):
Are we not marking the intended one deprecated or we don't surface applicationType from the SDK. If it's the latter, then LGTM.

vipulgupta2048 (approved):

PR #1420, opened 2024-06-20
Update to the v7 model

thgreasi:
Change-type: major
See: https://balena.fibery.io/Work/Project/Server-side-pagination-for-devices-Cycle-3-539


JSReds:
testing it I've noticed that now we cannot make any requests without passing at least a $select, is this wanted ? 
if so we should probably make options to be compulsory as methods like getAllByApplication are not working anymore:

```
await sdk.models.device.getAllByApplication(200555)

request.js:192 Uncaught BalenaRequestError: Request error: Must include a $select for top level 'device'
    at Object.eval (request.js:191:27)
    at Generator.next (<anonymous>)
    at fulfilled (tslib.es6.mjs:153:56)
        
 ```

thgreasi:
@JSReds are you using v7 from a branch or are you targeting /resin ? `$select`s should only be necessary on the /resin model.

When I tested the sdk against v7 on my bM last weekend, the tests cases passed as they are (w/o adding the missing `$select`s).
Moreover there are some test cases in the API on the v7 branch that do pass w/o a $select.
See: https://github.com/balena-io/balena-api/blob/7b5af1a2e1a22b968fc381b06e614484bc410d9f/test/01_user.ts#L1999

JSReds:
got it @thgreasi, I'm effectively using `/resin`, this is probably because of it :) thanks for the clarification :pray:

thgreasi:
I've just added test cases to make sure v7 works w/o $selects as well in the API PR :)
See: https://github.com/balena-io/balena-api/pull/5120/files#diff-50786f06ef687c91c0d34ed511be85399f074d7b4cd4387900c26bc15f35afa7R86

JSReds (approved):

PR #1427, opened 2024-07-12
pinejs-client-core: Add some missing methods to the custom typings

thgreasi:
Will allow to reduce some casting in https://github.com/balena-io/balena-cli/pull/2798

Change-type: patch

Resolves: # 
HQ: <url> 
See: <url> 
Depends-on: <url> 

---
##### Contributor checklist

- [ ] Includes tests
- [ ] Includes typings

### DIFF
--- a/typings/pinejs-client-core.d.ts
+++ b/typings/pinejs-client-core.d.ts
@@ -403,17 +403,17 @@ type ResourceAlternateKey<T> = SubmitBody<T>;
 type ResourceId<T> = BaseResourceId | ResourceAlternateKey<T>;
 
 export interface ParamsObj<T> {
-	resource?: string;
-	body?: SubmitBody<T>;
-	id?: ResourceId<T>;
-	options?: ODataOptions<T>;
-
 	apiPrefix?: string;
 	method?: ODataMethod;
+	resource?: string;
+	id?: ResourceId<T>;
 	url?: string;
+	body?: SubmitBody<T>;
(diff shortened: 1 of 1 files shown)


typings/pinejs-client-core.d.ts
```diff
@@ -485,9 +505,19 @@ export interface SubscribeParamsWithId<T> extends ParamsObjWithId<T> {
...
 
+export type ConstructorParams = Pick<
+	ParamsObj<unknown>,
+	(typeof validParams)[number]
+>;
+
 export interface Pine<ResourceTypeMap extends object = object> {
 	apiPrefix: string;
-	delete<T>(params: ParamsObjWithId<T> | ParamsObjWithFilter<T>): Promise<'OK'>;
+	passthrough: AnyObject;
```
otaviojacobi, line 515: What does passthrough means in the context of the SDK?

thgreasi: We are not using it in the SDK atm, but if we don't blindly copy it from pinejs-client-core, this will be an extra missing property of the base class typings that TS will be complaining about.

thgreasi: Effectively, this while PR is copying typings from pinejs-client-core, until we completely drop the custom typings.
So all these props are going to be added anyway once the custom typings get dropped.

otaviojacobi (approved):

PR #1441, opened 2024-10-01
tests: Use @balena/env-parsing

thgreasi:
Change-type: patch

Resolves: # 
HQ: <url> 
See: <url> 
Depends-on: <url> 
Change-type: major|minor|patch 

---
##### Contributor checklist

- [ ] Includes tests
- [ ] Includes typings
- [ ] Includes updated documentation
- [ ] Includes updated build output

PR #1442, opened 2024-10-01
Update @balena/lint to v9

thgreasi:
Update @balena/lint from 8.2.7 to 9.0.1

Depends-on: https://github.com/balena-io/balena-sdk/pull/1440
~~Depends-on: https://github.com/balena-io/balena-sdk/pull/1441~~
Change-type: patch


eslint.config.js
```diff
@@ -0,0 +1,20 @@
...
+		sourceType: 'module',
+	},
+	env: {
+		jest: true,
+	},
+	root: true,
+	rules: {
+		'no-restricted-imports': ['error', 'date-fns', 'lodash'],
+	},
+});
```
JSReds, line 20: as far as I know Page is pretty against custom config for a specific repo, did you discuss it with him ?

thgreasi: I only focused on bumping balena-lint in this PR, so this rule was there before, but tbh blocking top level imports in the SDK & the UI is definitely something that we need to keep the bundle size small, while on the other hand it would legitimately be adding friction to all other projects...

JSReds (approved):

PR #1443, opened 2024-10-03
Add option for detached HUP on startOsUpdate

jaomaloy:
This will call the v2 actions endpoint for resinhup which runs a detached version of HUP that increases HUP reliability on slow networks but will offer
no status updates such as in_progress.

Resolves: # 
HQ: <url> 
See: <url> 
Depends-on: <url> 

---
##### Contributor checklist

- [ ] Includes tests
- [ ] Includes typings

### DIFF
--- a/src/models/device.ts
+++ b/src/models/device.ts
@@ -151,8 +151,9 @@ const getDeviceModel = function (
 
 	const getOsUpdateHelper = once(async () => {
 		const $deviceUrlsBase = await getDeviceUrlsBase();
-		const { getOsUpdateHelper: _getOsUpdateHelper } =
-			require('../util/device-actions/os-update') as typeof import('../util/device-actions/os-update');
+		const _getOsUpdateHelper = (
+			await import('../util/device-actions/os-update')
+		).getOsUpdateHelper;
 		return _getOsUpdateHelper($deviceUrlsBase, request);
 	});
 	/* eslint-enable @typescript-eslint/no-require-imports */
@@ -334,14 +335,17 @@ const getDeviceModel = function (
 	async function startOsUpdate(
 		uuidOrUuids: string,
 		targetOsVersion: string,
+		options?: { runDetached?: boolean },
 	): Promise<OsUpdateActionResult>;
 	async function startOsUpdate(
 		uuidOrUuids: string[],
 		targetOsVersion: string,
+		options?: { runDetached?: boolean },
 	): Promise<Dictionary<OsUpdateActionResult>>;
 	async function startOsUpdate(
 		uuidOrUuids: string | string[],
 		targetOsVersion: string,
+		options: { runDetached?: boolean } = { runDetached: false },
 	): Promise<OsUpdateActionResult | Dictionary<OsUpdateActionResult>> {
 		if (!targetOsVersion) {
 			throw new errors.BalenaInvalidParameterError(
@@ -369,6 +373,7 @@ const getDeviceModel = function (
 		);
 
 		const osUpdateHelper = await getOsUpdateHelper();
+
 		const results: Dictionary<
 			ResolvableReturnType<typeof osUpdateHelper.startOsUpdate>
 		> = {};
@@ -410,10 +415,13 @@ const getDeviceModel = function (
 						);
 					}
 				}
+
+				// use the v2 device actions api for detached updates
 				await limitedMap(devices, async (device) => {
 					results[device.uuid] = await osUpdateHelper.startOsUpdate(
 						device.uuid,
 						targetOsVersion,
+						options.runDetached === true ? 'v2' : 'v1',
 					);
 				});
 			},
@@ -2262,6 +2270,10 @@ const getDeviceModel = function (
 		 * Unsupported (unpublished) version will result in rejection.
 		 * The version **must** be the exact version number, a "prod" variant and greater than the one running on the device.
 		 * To resolve the semver-compatible range use `balena.model.os.getMaxSatisfyingVersion`.
+		 * @param {Object} [options] - options
+		 * @param {Boolean} [options.runDetached] - run the update in detached mode.
+		 * Default behaviour is runDetached=false but is DEPRECATED and will be removed in a future release. Use runDetached=true
+		 * for more reliable updates.
 		 * @fulfil {Object} - action response
 		 * @returns {Promise}
 		 *
@@ -2273,7 +2285,8 @@ const getDeviceModel = function (
 		startOsUpdate,
 
 		/**
-		 * @summary Get the OS update status of a device
+		 * @deprecated
+		 * @summary Get the OS update status of a device. This will no longer return a useful status for runDetached=true updates.
 		 * @name getOsUpdateStatus
 		 * @public
 		 * @function
--- a/src/util/device-actions/device-actions-service.ts
+++ b/src/util/device-actions/device-actions-service.ts
@@ -3,6 +3,7 @@ import type { BalenaRequest } from 'balena-request';
 interface MakeActionRequestParams {
 	uuid: string;
 	actionNameOrId: string | number;
+	deviceActionsApiVersion: 'v1' | 'v2';
 	params?: any;
 	method: 'GET' | 'POST' | 'PUT' | 'DELETE' | 'PATCH' | 'HEAD' | 'OPTIONS';
 	extraOptions: any;
@@ -11,6 +12,7 @@ interface MakeActionRequestParams {
 interface StartActionParams {
 	uuid: string;
 	actionName: string;
+	deviceActionsApiVersion: 'v1' | 'v2';
 	params: any;
 	extraOptions?: any;
 }
@@ -21,8 +23,6 @@ interface GetActionStatusParams {
 	extraOptions?: any;
 }
 
-const DEVICE_ACTIONS_API_VERSION = 'v1';
-
 export class DeviceActionsService {
 	private actionsEndpoint: string;
 
@@ -30,12 +30,13 @@ export class DeviceActionsService {
 		deviceUrlsBase: string,
 		private request: BalenaRequest,
 	) {
-		this.actionsEndpoint = `https://actions.${deviceUrlsBase}/${DEVICE_ACTIONS_API_VERSION}`;
+		this.actionsEndpoint = `https://actions.${deviceUrlsBase}`;
 	}
 
 	public startAction = <T>({
 		uuid,
 		actionName,
+		deviceActionsApiVersion,
 		params,
 		extraOptions,
 	}: StartActionParams) =>
@@ -43,6 +44,7 @@ export class DeviceActionsService {
 			method: 'POST',
 			uuid,
 			actionNameOrId: actionName,
+			deviceActionsApiVersion,
 			params,
 			extraOptions,
 		});
@@ -55,6 +57,7 @@ export class DeviceActionsService {
 		this.makeActionRequest<T>({
 			method: 'GET',
 			uuid,
+			deviceActionsApiVersion: 'v1',
 			actionNameOrId: actionId,
 			extraOptions,
 		});
@@ -63,6 +66,7 @@ export class DeviceActionsService {
 		method,
 		uuid,
 		actionNameOrId,
+		deviceActionsApiVersion,
 		params,
 		extraOptions,
 	}: MakeActionRequestParams): Promise<T> => {
@@ -70,7 +74,7 @@ export class DeviceActionsService {
 
 		const { body } = await this.request.send<T>({
 			method,
-			url: `${this.actionsEndpoint}/${uuid}/${actionNameOrId}`,
+			url: `${this.actionsEndpoint}/${deviceActionsApiVersion}/${uuid}/${actionNameOrId}`,
 			body: data,
 			...extraOptions,
 		});
--- a/DOCUMENTATION.md
+++ b/DOCUMENTATION.md
@@ -301,8 +301,8 @@ const sdk = fromSharedOptions();
             * [.pinToRelease(uuidOrIdOrArray, fullReleaseHashOrId)](#balena.models.device.pinToRelease) ⇒ <code>Promise</code>
             * [.trackApplicationRelease(uuidOrIdOrArray)](#balena.models.device.trackApplicationRelease) ⇒ <code>Promise</code>
             * [.setSupervisorRelease(uuidOrIdOrArray, supervisorVersionOrId)](#balena.models.device.setSupervisorRelease) ⇒ <code>Promise</code>
-            * [.startOsUpdate(uuidOrUuids, targetOsVersion)](#balena.models.device.startOsUpdate) ⇒ <code>Promise</code>
-            * [.getOsUpdateStatus(uuid)](#balena.models.device.getOsUpdateStatus) ⇒ <code>Promise</code>
+            * [.startOsUpdate(uuidOrUuids, targetOsVersion, [options])](#balena.models.device.startOsUpdate) ⇒ <code>Promise</code>
+            * ~~[.getOsUpdateStatus(uuid)](#balena.models.device.getOsUpdateStatus) ⇒ <code>Promise</code>~~
             * [.ping(uuidOrId)](#balena.models.device.ping) ⇒ <code>Promise</code>
             * ~~[.getApplicationInfo(uuidOrId)](#balena.models.device.getApplicationInfo) ⇒ <code>Promise</code>~~
             * [.identify(uuidOrId)](#balena.models.device.identify) ⇒ <code>Promise</code>
@@ -708,8 +708,8 @@ balena.models.device.get(123).catch(function (error) {
         * [.pinToRelease(uuidOrIdOrArray, fullReleaseHashOrId)](#balena.models.device.pinToRelease) ⇒ <code>Promise</code>
         * [.trackApplicationRelease(uuidOrIdOrArray)](#balena.models.device.trackApplicationRelease) ⇒ <code>Promise</code>
         * [.setSupervisorRelease(uuidOrIdOrArray, supervisorVersionOrId)](#balena.models.device.setSupervisorRelease) ⇒ <code>Promise</code>
-        * [.startOsUpdate(uuidOrUuids, targetOsVersion)](#balena.models.device.startOsUpdate) ⇒ <code>Promise</code>
-        * [.getOsUpdateStatus(uuid)](#balena.models.device.getOsUpdateStatus) ⇒ <code>Promise</code>
+        * [.startOsUpdate(uuidOrUuids, targetOsVersion, [options])](#balena.models.device.startOsUpdate) ⇒ <code>Promise</code>
+        * ~~[.getOsUpdateStatus(uuid)](#balena.models.device.getOsUpdateStatus) ⇒ <code>Promise</code>~~
         * [.ping(uuidOrId)](#balena.models.device.ping) ⇒ <code>Promise</code>
         * ~~[.getApplicationInfo(uuidOrId)](#balena.models.device.getApplicationInfo) ⇒ <code>Promise</code>~~
         * [.identify(uuidOrId)](#balena.models.device.identify) ⇒ <code>Promise</code>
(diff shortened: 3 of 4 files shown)


src/models/device.ts
```diff
@@ -342,6 +351,7 @@ const getDeviceModel = function (
 	async function startOsUpdate(
```
kb2ma, line 351: I believe the alternate signatures for this function above also need an options parameter.

src/models/device.ts
```diff
@@ -350,6 +360,10 @@ const getDeviceModel = function (
 			);
 		}
 
+		if (options == null) {
+			options = {};
```
kb2ma, line 364: Define the default value for runDetached as false.

src/models/device.ts
```diff
@@ -368,7 +382,10 @@ const getDeviceModel = function (
 			{ primitive: true, promise: true },
 		);
 
-		const osUpdateHelper = await getOsUpdateHelper();
+		const osUpdateHelper = options.runDetached === true
+		? await getOsUpdateHelperv2()
```
kb2ma, line 386: I don't see the value of a second getOsUpdateHelper function. Just parameterize getOsUpdateHelper() to include a version number parameter. Use that parameter to pass the right endpoint version to DeviceActionsService. Also, for version 2, don't return a getOsUpdateStatus() function.

kb2ma:
One more thing to add OsUpdateActionResult needs a 'triggered' status option. For v2, the status is either 'idle', 'triggered', or possibly 'error'. It will error for example if there is a bad parameter in the request. In that case the HUP is not started on the device. See `balena-proxy src/services/actions-backend/app.ts`.

src/models/device.ts
```diff
@@ -350,6 +360,10 @@ const getDeviceModel = function (
 			);
 		}
 
+		if (options == null) {
+			options = {};
+		}
```
Page-, line 365: Use `options: { runDetached?: boolean } = {}` above to provide a default argument and avoid the need for this
```suggestion
```

src/models/device.ts
```diff
@@ -157,6 +157,15 @@ const getDeviceModel = function (
 	});
 	/* eslint-enable @typescript-eslint/no-require-imports */
 
+	const getOsUpdateHelperv2 = once(async () => {
+		const $deviceUrlsBase = await getDeviceUrlsBase();
+		const { getOsUpdateHelperv2: _getOsUpdateHelperv2 } =
+			require('../util/device-actions/os-update') as typeof import('../util/device-actions/os-update');
```
Page-, line 163: I think you can use a normal import like:
```suggestion
		const { getOsUpdateHelperv2: _getOsUpdateHelperv2 } = await('../util/device-actions/os-update');
```
since we're already in an async context and don't need to worry about using require for synchronous loading

src/util/device-actions/device-actions-service.ts
```diff
@@ -21,16 +21,15 @@ interface GetActionStatusParams {
...
 }
 
-const DEVICE_ACTIONS_API_VERSION = 'v1';
-
 export class DeviceActionsService {
 	private actionsEndpoint: string;
 
 	constructor(
 		deviceUrlsBase: string,
+		deviceActionsApiVersion: string,
```
Page-, line 29: We can be a bit more precise with:
```suggestion
		deviceActionsApiVersion: 'v1' | 'v2',
```

src/models/device.ts
```diff
@@ -149,11 +149,10 @@ const getDeviceModel = function (
 		}),
 	);
 
-	const getOsUpdateHelper = once(async () => {
+	const getOsUpdateHelper = once(async (deviceActionsApiVersion: string) => {
```
Page-, line 152: We can be a bit more precise with:
```suggestion
	const getOsUpdateHelper = once(async (deviceActionsApiVersion: 'v1' | 'v2') => {
```

src/models/device.ts
```diff
@@ -334,14 +341,17 @@ const getDeviceModel = function (
...
 	): Promise<OsUpdateActionResult>;
 	async function startOsUpdate(
 		uuidOrUuids: string[],
 		targetOsVersion: string,
+		options?: { runDetached?: boolean },
 	): Promise<Dictionary<OsUpdateActionResult>>;
 	async function startOsUpdate(
 		uuidOrUuids: string | string[],
 		targetOsVersion: string,
+		options?: { runDetached?: boolean },
```
Page-, line 354: You can set the default here and then remove the manual default assignment from below
```suggestion
		options: { runDetached?: boolean } = { runDetached: false },
```

src/models/device.ts
```diff
@@ -350,6 +360,10 @@ const getDeviceModel = function (
 			);
 		}
 
+		if (options == null) {
```
kb2ma, line 363: You can remove this now that you have the default above, right?

jaomaloy: Sorry yes I missed that

kb2ma:
@jaomaloy I've gone through all the code now, and overall it is more compact and looks good. Please update the API comments in `src/models/device.ts` to describe that `startOsUpdate()` is deprecated for for runDetached=false. Also add a comment to `getOsUpdateStatus()` also is deprecated, and will not return a useful status for runDetached=true.

src/models/device.ts
```diff
@@ -368,7 +378,11 @@ const getDeviceModel = function (
 			{ primitive: true, promise: true },
 		);
 
-		const osUpdateHelper = await getOsUpdateHelper();
+		const osUpdateHelper =
```
kb2ma, line 381: Please add an internal comment that gives a short description of v1 vs. v2 so the reader will have an idea of the purpose.

jaomaloy:
> @jaomaloy I've gone through all the code now, and overall it is more compact and looks good. Please update the API comments in `src/models/device.ts` to describe that `startOsUpdate()` is deprecated for for runDetached=false. These marks will update the online API docs to show in strikethrough font.
> 
> Also add a comment to `getOsUpdateStatus()` also is deprecated, and will not return a useful status for runDetached=true.

We will do this at the next step, correct? This PR isn't for deprecation yet but to add the option for true so the rest of our code can use runDetached=true.

kb2ma:
@jaomaloy we want to deprecate now to give users notice of the change and time to adapt.

kb2ma:
Tests look OK, except for what I describe below.

When I test v1, I see these results when starting the update, as expected.
```
> await balena.models.device.startOsUpdate("43c932f91a2e305a13f6b692faf39230", "2.99.27")
{
  status: 'in_progress',
  lastRun: 1729563480785,
  parameters: { target_version: '2.99.27' },
  action: 'resinhup'
}
> await balena.models.device.getOsUpdateStatus("43c932f91a2e305a13f6b692faf39230")
{
  status: 'in_progress',
  lastRun: 1729563480785,
  parameters: { target_version: '2.99.27' },
  action: 'resinhup'
}
```

However, when running v2, I see these results when starting the update. I actually would expect to see similar results -- basically whatever is in the action server's ACTION_STATUS dictionary.
```
> await balena.models.device.startOsUpdate("43c932f91a2e305a13f6b692faf39230", "4.0.26+rev1", {runDetached: true})
{
  status: 'triggered',
  lastRun: 1729569204298,
  parameters: { target_version: '4.0.26+rev1' },
  action: 'resinhup'
}
> await balena.models.device.getOsUpdateStatus("43c932f91a2e305a13f6b692faf39230")
{ status: 'error' }
```

Looking into the error, I see this in the log for balena-proxy -- generating an error just looking up the device. Of course we don't support getOsUpdateStatus() when running in detached mode. However, it should not generate an error either. Do you get the same results? My concern is that there may be some code or rules in pinejs that are sensitive to the status of the device.
```
Oct 22 04:28:24 d90c519 2c27e73cdc26[1631]: [40938.112701] actions-backend[7773]: ::ffff:127.0.0.1 GET actions.devices.d90c5192ae585eaed21f1f48258c954c.kb2ma.balena-dev.com /v1/43c932f91a2e305a13f6b692faf39230/resinhup 400 12 10.968ms curl/7.81.0
Oct 22 04:28:24 d90c519 2c27e73cdc26[1631]: [40938.112886] proxy[7757]: ::ffff:192.168.1.100 GET actions.devices.d90c5192ae585eaed21f1f48258c954c.kb2ma.balena-dev.com /v1/43c932f91a2e305a13f6b692faf39230/resinhup 400 12 15.058ms curl/7.81.0
Oct 22 04:28:24 d90c519 2c27e73cdc26[1631]: [40938.113011] actions-backend[7773]: error: Error: Actions Backend error: -  StatusError
Oct 22 04:28:24 d90c519 2c27e73cdc26[1631]: [40938.113076] actions-backend[7773]:     at PinejsClientRequest._request (/usr/src/app/node_modules/pinejs-client-request/request.ts:143:10)
Oct 22 04:28:24 d90c519 2c27e73cdc26[1631]: [40938.113122] actions-backend[7773]:     at process.processTicksAndRejections (node:internal/process/task_queues:105:5)
Oct 22 04:28:24 d90c519 2c27e73cdc26[1631]: [40938.113166] actions-backend[7773]:     at async PinejsClientRequest.callWithRetry (/usr/src/app/node_modules/pinejs-client-core/src/index.ts:1173:11)
Oct 22 04:28:24 d90c519 2c27e73cdc26[1631]: [40938.113205] actions-backend[7773]:     at async PinejsClientRequest.get (/usr/src/app/node_modules/pinejs-client-core/src/index.ts:1308:18)
Oct 22 04:28:24 d90c519 2c27e73cdc26[1631]: [40938.113245] actions-backend[7773]:     at async getDeviceInfo (/usr/src/app/src/common/api-utils.ts:146:17)
Oct 22 04:28:24 d90c519 2c27e73cdc26[1631]: [40938.113283] actions-backend[7773]:     at async <anonymous> (/usr/src/app/src/services/actions-backend/app.ts:689:23)
```

kb2ma:
There is another problem. If I run a HUP in non-detached mode, and then run a second HUP in detached mode, the second HUP actually runs in non-detached mode.

1st HUP:
```
> await balena.models.device.getOsUpdateStatus("447c7aeda6702cac332d48bc65d78797")
{ status: 'idle', action: 'resinhup', lastRun: null }
> await balena.models.device.startOsUpdate("447c7aeda6702cac332d48bc65d78797", "2.99.27", {runDetached: false})
{
  status: 'in_progress',
  lastRun: 1729575207485,
  parameters: { target_version: '2.99.27' },
  action: 'resinhup'
}
> await balena.models.device.getOsUpdateStatus("447c7aeda6702cac332d48bc65d78797")
{
  status: 'in_progress',
  lastRun: 1729575207485,
  parameters: { target_version: '2.99.27' },
  action: 'resinhup'
}
> await balena.models.device.getOsUpdateStatus("447c7aeda6702cac332d48bc65d78797")
{
  status: 'done',
  action: 'resinhup',
  stdout: '[upgrade-2.x.sh][000000000][INFO]Loading info from config.json\n',
  error: 'failed to get digest sha256:cfe5adead644da98672552df619e3758fe4e8773abe02f54b95bf01cd2fd71f4: open /mnt/sysroot/inactive/balena/image/overlay2/imagedb/content/sha256/cfe5adead644da98672552df619e3758fe4e8773abe02f54b95bf01cd2fd71f4: no such file or directory\n' +
    "WARNING: The requested image's platform (linux/amd64) does not match the detected host platform (linux/arm64/v8) and no specific platform was requested\n" +
    "WARNING: The requested image's platform (linux/amd64) does not match the detected host platform (linux/arm64/v8) and no specific platform was requested\n" +
    "WARNING: The requested image's platform (linux/amd64) does not match the detected host platform (linux/arm64/v8) and no specific platform was requested\n" +
    'Running as unit: run-update-supervisor.service\n' +
    'Finished with result: success\n' +
    'Main processes terminated with: code=exited/status=0\n' +
    'Service runtime: 9.969s\n',
  lastRun: 1729575327902,
  parameters: { target_version: '2.99.27' }
}
```
2nd HUP. Notice status should be 'triggered'.
```
> await balena.models.device.startOsUpdate("447c7aeda6702cac332d48bc65d78797", "4.0.26", {runDetached: true})
{
  status: 'in_progress',
  lastRun: 1729575680882,
  parameters: { target_version: '4.0.26' },
  action: 'resinhup'
}
> await balena.models.device.getOsUpdateStatus("447c7aeda6702cac332d48bc65d78797")
{
  status: 'in_progress',
  lastRun: 1729575680882,
  parameters: { target_version: '4.0.26' },
  action: 'resinhup'
}
```

Proxy log shows the second attempt did indeed call the /v1 endpoint:
```
Oct 22 05:41:20 d90c519 2c27e73cdc26[1631]: [45314.541747] actions-backend[7773]: ::ffff:127.0.0.1 POST actions.devices.d90c5192ae585eaed21f1f48258c954c.kb2ma.balena-dev.com /v1/447c7aeda6702cac332d48bc65d78797/resinhup 200 109 63.712ms node-fetch/1.0 (+https://github.com/bitinn/node-fetch)
```

jaomaloy:
> Tests look OK, except for what I describe below.
> 
> When I test v1, I see these results when starting the update, as expected.
> 
> ```
> > await balena.models.device.startOsUpdate("43c932f91a2e305a13f6b692faf39230", "2.99.27")
> {
>   status: 'in_progress',
>   lastRun: 1729563480785,
>   parameters: { target_version: '2.99.27' },
>   action: 'resinhup'
> }
> > await balena.models.device.getOsUpdateStatus("43c932f91a2e305a13f6b692faf39230")
> {
>   status: 'in_progress',
>   lastRun: 1729563480785,
>   parameters: { target_version: '2.99.27' },
>   action: 'resinhup'
> }
> ```
> 
> However, when running v2, I see these results when starting the update. I actually would expect to see similar results -- basically whatever is in the action server's ACTION_STATUS dictionary.
> 
> ```
> > await balena.models.device.startOsUpdate("43c932f91a2e305a13f6b692faf39230", "4.0.26+rev1", {runDetached: true})
> {
>   status: 'triggered',
>   lastRun: 1729569204298,
>   parameters: { target_version: '4.0.26+rev1' },
>   action: 'resinhup'
> }
> > await balena.models.device.getOsUpdateStatus("43c932f91a2e305a13f6b692faf39230")
> { status: 'error' }
> ```
> 
> Looking into the error, I see this in the log for balena-proxy -- generating an error just looking up the device. Of course we don't support getOsUpdateStatus() when running in detached mode. However, it should not generate an error either. Do you get the same results? My concern is that there may be some code or rules in pinejs that are sensitive to the status of the device.
> 
> ```
> Oct 22 04:28:24 d90c519 2c27e73cdc26[1631]: [40938.112701] actions-backend[7773]: ::ffff:127.0.0.1 GET actions.devices.d90c5192ae585eaed21f1f48258c954c.kb2ma.balena-dev.com /v1/43c932f91a2e305a13f6b692faf39230/resinhup 400 12 10.968ms curl/7.81.0
> Oct 22 04:28:24 d90c519 2c27e73cdc26[1631]: [40938.112886] proxy[7757]: ::ffff:192.168.1.100 GET actions.devices.d90c5192ae585eaed21f1f48258c954c.kb2ma.balena-dev.com /v1/43c932f91a2e305a13f6b692faf39230/resinhup 400 12 15.058ms curl/7.81.0
> Oct 22 04:28:24 d90c519 2c27e73cdc26[1631]: [40938.113011] actions-backend[7773]: error: Error: Actions Backend error: -  StatusError
> Oct 22 04:28:24 d90c519 2c27e73cdc26[1631]: [40938.113076] actions-backend[7773]:     at PinejsClientRequest._request (/usr/src/app/node_modules/pinejs-client-request/request.ts:143:10)
> Oct 22 04:28:24 d90c519 2c27e73cdc26[1631]: [40938.113122] actions-backend[7773]:     at process.processTicksAndRejections (node:internal/process/task_queues:105:5)
> Oct 22 04:28:24 d90c519 2c27e73cdc26[1631]: [40938.113166] actions-backend[7773]:     at async PinejsClientRequest.callWithRetry (/usr/src/app/node_modules/pinejs-client-core/src/index.ts:1173:11)
> Oct 22 04:28:24 d90c519 2c27e73cdc26[1631]: [40938.113205] actions-backend[7773]:     at async PinejsClientRequest.get (/usr/src/app/node_modules/pinejs-client-core/src/index.ts:1308:18)
> Oct 22 04:28:24 d90c519 2c27e73cdc26[1631]: [40938.113245] actions-backend[7773]:     at async getDeviceInfo (/usr/src/app/src/common/api-utils.ts:146:17)
> Oct 22 04:28:24 d90c519 2c27e73cdc26[1631]: [40938.113283] actions-backend[7773]:     at async <anonymous> (/usr/src/app/src/services/actions-backend/app.ts:689:23)
> ```

This is likely to be because if we getOsUpdateStatus with a v2 running, the result is forced to 'error'. I could just make this return new error type instead?
```
		getOsUpdateStatus: async (uuid: string): Promise<OsUpdateActionResult> => {
			try {
				const osUpdateHelper = await getOsUpdateHelper('v1');
				const result = await osUpdateHelper.getOsUpdateStatus?.(uuid);

				if (result === undefined) {
					return { status: 'error' } as OsUpdateActionResult;
				}

				return result;
			} catch (err) {
				if (err.statusCode !== 400) {
					throw err;
				}

				// as an attempt to reduce the requests for this method
				// check whether the device exists only when the request rejects
				// so that it's rejected with the appropriate BalenaDeviceNotFound error
				await exports.get(uuid, { $select: 'id' });
				// if the device exists, then re-throw the original error
				throw err;
			}
		},
```

kb2ma:
It's good that the SDK code is catching this, but I don't think we need a separate error. My concern is that the error occurs on line 689 in balena-proxy `src/services/actions-backend/app.ts`, where the resinhup GET is just looking up device info.

For v2, the resinhup GET should still return either 'triggered' or 'idle'. They are not really useful, but I don't understand why an error would be generated in this case. I'm concerned that backend modules like the UI will use the SDK as well, and also will get this odd error.

Is there something in the pinejs rules or SQL that is failing because the device is in an unexpected state? @Page- , any insights here?

src/models/device.ts
```diff
@@ -2273,7 +2292,7 @@ const getDeviceModel = function (
 		startOsUpdate,
 
 		/**
-		 * @summary Get the OS update status of a device
+		 * @summary (Deprecated) Get the OS update status of a device. This will no longer return a useful status for runDetached=true updates.
```
kb2ma, line 2295: Use the @deprecated [tag](https://www.typescriptlang.org/docs/handbook/jsdoc-supported-types.html#deprecated). You can add additional text to it.

src/models/device.ts
```diff
@@ -149,12 +149,17 @@ const getDeviceModel = function (
 		}),
 	);
 
-	const getOsUpdateHelper = once(async () => {
+	const getOsUpdateHelper = async (deviceActionsApiVersion: 'v1' | 'v2') => {
```
kb2ma, line 152: If we remove the call to `once`, it doesn't make sense to retain getOsUpdateHelper because it doesn't really provide any value. The problem is that we need to dynamically pass the API version when actually running the startOsUpdate() or getOsUpdateStatus() functions. Well, we already pass in uuid dynamically to those functions, so let's pass apiVersion the same way.

So to retain the `once`, remove the API version from util.device-actions.os-update.getOsUpdateHelper(). Instead add apiVersion as a parameter to the startOsUpdate() function in that module, just like uuid. Modify the callers of that function to pass in the API version at that time. 

You don't need to add an apiVersion to the getOsUpdateStatus() function in util.device-actions.os-update because there is not a v2, and the SDK user does not specify `runDetached`. Actually, I think this change means the src.models.getOsUpdateStatus() doesn't change at all in this PR!

This way we keep the runtime performance the same at essentially no extra cost. Every time those functions are called, they end up calling DeviceActionService startAction() and getActionStatus() which dynamically compute the URL for the action server endpoint anyway.

Please create a new commit for this change so we can easily review the diff. We can squash later.

src/models/device.ts
```diff
@@ -2262,6 +2270,10 @@ const getDeviceModel = function (
 		 * Unsupported (unpublished) version will result in rejection.
 		 * The version **must** be the exact version number, a "prod" variant and greater than the one running on the device.
 		 * To resolve the semver-compatible range use `balena.model.os.getMaxSatisfyingVersion`.
+		 * @param {Object} [options] - options
+		 * @param {Boolean} [options.runDetached] - run the update in detached mode.
+		 * Default behaviour is runDetached=false but is DEPRECATED and will be removed in a future release. Use runDetached=true
```
kb2ma, line 2275: Nice comment!

kb2ma:
LGTM, please rebase and squash commits. @Page- do you want to take another look?

kb2ma:
When I do a local build, the DOCUMENTATION.md file is updated for the deprecations. I expect this update should be included with the PR. Also, please confirm that the documentation for getOsUpdateStatus() is accurate. I'm not sure it works to put the @summary and @deprecated on a single line.

kb2ma (changes requested):
Need to resolve comment above on docs.

jaomaloy:
> When I do a local build, the DOCUMENTATION.md file is updated for the deprecations. I expect this update should be included with the PR. Also, please confirm that the documentation for getOsUpdateStatus() is accurate. I'm not sure it works to put the @summary and @deprecated on a single line.

yep, @deprecated needed a new line.

PR #1450, opened 2024-10-22
Remove non-existent properties from model typings

Page-:
Change-type: patch

Resolves: # 
HQ: <url> 
See: <url> 
Depends-on: <url> 

---
##### Contributor checklist

- [ ] Includes tests
- [ ] Includes typings

### DIFF
--- a/typing_tests/pine-typed-result.ts
+++ b/typing_tests/pine-typed-result.ts
@@ -223,7 +223,6 @@ export let aStringOrUndefined: string | undefined;
 	aNumber = result.id;
 	aString = result.device_name;
 	aNumber = result.actor[0].id;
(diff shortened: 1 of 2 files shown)


typing_tests/pine-typed-result.ts
```diff
@@ -223,7 +223,6 @@ export let aStringOrUndefined: string | undefined;
 	aNumber = result.id;
 	aString = result.device_name;
 	aNumber = result.actor[0].id;
-	aString = result.actor[0].created_at;
```
Page-, line 226: Yes this is a change, but the real type will always be `undefined` since it doesn't actually exist in any response..

thgreasi (approved):

PR #1453, opened 2024-10-24
Typings: fix some id typings to the correct typing

Page-:
Change-type: patch

Resolves: # 
HQ: <url> 
See: <url> 
Depends-on: <url> 

---
##### Contributor checklist

- [ ] Includes tests
- [ ] Includes typings

### DIFF
--- a/src/models/application-membership.ts
+++ b/src/models/application-membership.ts
@@ -53,7 +53,7 @@ const getApplicationMembershipModel = function (
 		const role = await pine.get({
 			resource: 'application_membership_role',
 			id: {
-				name: roleName,
+				name: roleName as ApplicationMembershipRoles,
 			},
 			options: {
 				$select: 'id',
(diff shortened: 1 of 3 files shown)


src/models/application-membership.ts
```diff
@@ -53,7 +53,7 @@ const getApplicationMembershipModel = function (
 		const role = await pine.get({
 			resource: 'application_membership_role',
 			id: {
-				name: roleName,
+				name: roleName as ApplicationMembershipRoles,
```
Page-, line 56: These 2 casts are necessary because the type being passed in is `string` but that doesn't match the custom typing meaning the resource lookup fails and the returned `id` becomes `string | number | undefined` rather than `number | undefined` which then fails later as it passes the id to a model that doesn't allow a string. Doing this cast means it can resolve the resource properly and continue to infer the `number | undefined` response

thgreasi (approved):

PR #1495, opened 2024-11-15
Update renovate config

thgreasi:
Resolves: # 
HQ: <url> 
See: <url> 
Depends-on: <url> 

---
##### Contributor checklist

- [ ] Includes tests
- [ ] Includes typings

### DIFF
--- a/.github/renovate.json
+++ b/.github/renovate.json
@@ -1,11 +1,13 @@
 {
 	"extends": ["github>balena-io/renovate-config"],
 	"prConcurrentLimit": 1,
+	"prHourlyLimit": 1,
 	"rebasewhen": "automerge",
 	"packageRules": [
 		{
 			"matchUpdateTypes": ["major", "minor", "patch", "digest"],
 			"automerge": false,
+			"prHourlyLimit": 1,
 			"rebasewhen": "never",
 			"labels": ["renovate", "dependencies", "{{depType}}", "major"]
 		}
--- a/.github/renovate.json
+++ b/.github/renovate.json
@@ -2,14 +2,5 @@
 	"extends": ["github>balena-io/renovate-config"],
 	"prConcurrentLimit": 1,
 	"prHourlyLimit": 1,
-	"rebasewhen": "automerge",
-	"packageRules": [
-		{
-			"matchUpdateTypes": ["major", "minor", "patch", "digest"],
-			"automerge": false,
-			"prHourlyLimit": 1,
-			"rebasewhen": "never",
-			"labels": ["renovate", "dependencies", "{{depType}}", "major"]
-		}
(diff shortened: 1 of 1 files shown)


.github/renovate.json
```diff
@@ -1,11 +1,18 @@
 {
 	"extends": ["github>balena-io/renovate-config"],
-	"prConcurrentLimit": 1,
+	"prConcurrentLimit": 3,
+	"prHourlyLimit": 1,
+	"rebasewhen": "automerge",
 	"packageRules": [
 		{
-			"matchUpdateTypes": ["major", "minor", "patch", "digest"],
+			"groupName": "only-manually-bumped-packages",
```
klutchell, line 8: You don't need a groupName if enabled == false.

.github/renovate.json
```diff
@@ -1,11 +1,18 @@
...
+			"matchPackageNames": ["balena-hup-action-utils"],
+			"matchUpdateTypes": [
+				"major",
+				"minor"
+			],
 			"automerge": false,
```
klutchell, line 14: You don't need an automerge setting if enabled == false.

.github/renovate.json
```diff
@@ -1,11 +1,18 @@
...
-			"labels": ["renovate", "dependencies", "{{depType}}", "major"]
+			"enabled": false
```
klutchell, line 15: Do you really want renovate to never open a PR for this, or would you like it to open a PR for convenience but never rebase it and disable automerge?

thgreasi: If it's a major or a minor, which has a dependency of the proxy being bumped & deployed first, I think it makes sense to avoid opening the PR completely. I expect that who-ever does such a minor or major, will anyway have to bump it everywhere as part of shipping their project.

klutchell: ack, as long as you're okay with creating more PRs by hand

.github/renovate.json
```diff
@@ -1,11 +1,16 @@
 {
 	"extends": ["github>balena-io/renovate-config"],
-	"prConcurrentLimit": 1,
+	"prConcurrentLimit": 3,
```
klutchell, line 3: Why are you increasing this?

thgreasi: B/c I expect that `prHourlyLimit: 1` might help enough and having just 1 starving PR sounds like renovate will always be blocked. I can start w/ leaving `prConcurrentLimit: 1` as it was and only increase it later if you prefer.
Yea let me do that...

.github/renovate.json
thgreasi: Like I have always done :)

klutchell (approved):

PR #1498, opened 2024-11-20
Pin to target OS release in startOsUpdate and skip action-based HUP if OS >=vTODO

cywang117:
Change-type: minor

See: https://balena.fibery.io/Work/Improvement/API-and-SDK-changes-for-supervisor-hostOS-updates-2348
Change-type: minor
---
##### Contributor checklist

- [ ] Includes tests
- [ ] Includes typings


src/models/device.ts
```diff
@@ -107,6 +107,7 @@ export type DeviceMetrics = Pick<
 >;
 
 const DEFAULT_DAYS_OF_REQUESTED_HISTORY = 7;
+const SUPERVISOR_MANAGED_HUP_MIN_OS_VERSION = 'vTODO';
```
kb2ma, line 110: Also need to compare ESR releases?

PR #1499, opened 2024-11-21
Ignore host app service installs for device service env vars

otaviojacobi:
Change-type: patch

Resolves: # 
HQ: <url> 
See: <url> 
Depends-on: <url> 

---
##### Contributor checklist

- [ ] Includes tests
- [ ] Includes typings


src/models/device.ts
```diff
@@ -2984,31 +2997,25 @@ const getDeviceModel = function (
 			): Promise<void> {
 				value = String(value);
 
-				let deviceFilter;
-				if (isId(uuidOrId)) {
```
thgreasi, line 2988: It would be great if we could preserve this optimization, since I think the UI relies on it to make less requests.
Other than that LGTM :+1:

src/models/device.ts
```diff
@@ -2911,7 +2911,19 @@ const getDeviceModel = function (
 				serviceNameOrId: string | number,
 				key: string,
 			): Promise<string | undefined> {
-				const { id: deviceId } = await exports.get(uuidOrId, { $select: 'id' });
+				const deviceOptions = {
+					$select: 'id',
+					$expand: { belongs_to__application: { $select: 'id' } },
+				} as const;
```
thgreasi, line 2917:
```suggestion
				const deviceOptions = {
					$select: ['id', 'belongs_to__application'],
				} as const;
```

PR #1505, opened 2024-11-26
Major v21

myarmolinsky:
Drops support for OS versions <= 2.14.0
Requires open-balena-api 34.1.0

Depends-on: https://github.com/balena-io/open-balena-api/pull/1890
Change-type: major

---
##### Contributor checklist

- [ ] Includes tests
- [ ] Includes typings

### DIFF
--- a/src/models/api-key.ts
+++ b/src/models/api-key.ts
@@ -38,39 +38,49 @@ const getApiKeysModel = function (
 		 *
 		 * @description This method registers a new api key for the current user with the name given.
 		 *
-		 * @param {String} name - the API key name
-		 * @param {String} [description=null] - the API key description
+		 * @param {Object} createApiKeyParams - an object containing the parameters for the creation of an API key
+		 * @param {String} createApiKeyParams.name - the API key name
+		 * @param {String} createApiKeyParams.expiryDate - the API key expiry date
+		 * @param {String} [createApiKeyParams.description=null] - the API key description
 		 *
 		 * @fulfil {String} - API key
 		 * @returns {Promise}
 		 *
 		 * @example
-		 * balena.models.apiKey.create(apiKeyName).then(function(apiKey) {
+		 * balena.models.apiKey.create({name: apiKeyName, expiryDate: 2030-10-12}).then(function(apiKey) {
 		 * 	console.log(apiKey);
 		 * });
 		 *
 		 * @example
-		 * balena.models.apiKey.create(apiKeyName, apiKeyDescription).then(function(apiKey) {
+		 * balena.models.apiKey.create({name: apiKeyName, expiryDate: 2030-10-12, description: apiKeyDescription}).then(function(apiKey) {
 		 * 	console.log(apiKey);
 		 * });
 		 */
-		async create(
-			name: string,
-			description: string | null = null,
-			expiryDate: string | null = null,
-		): Promise<string> {
+		async create({
+			name,
+			expiryDate,
+			description = null,
+		}: {
+			name: string;
+			expiryDate: string | null;
+			description?: string | null;
+		}): Promise<string> {
+			if (expiryDate === undefined) {
+				throw new errors.BalenaInvalidParameterError(
+					'createApiKeyParams.expiryDate',
+					expiryDate,
+				);
+			}
 			const apiKeyBody: {
 				name: string;
+				expiryDate: string | null;
 				description?: string | null;
-				expiryDate?: string | null;
 			} = {
 				name,
+				expiryDate,
 			};
 			if (typeof description === 'string' && !!description) {
 				apiKeyBody.description = description;
-			}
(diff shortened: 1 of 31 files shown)


typing_tests/pine-options.ts
```diff
@@ -989,5 +988,6 @@ export const string$OrderbyDesc: BalenaSdk.PineOptions<BalenaSdk.Application> =
 
 export const string$OrderbyNestedProp: BalenaSdk.PineOptions<BalenaSdk.Organization> =
 	{
+		// @ts-expect-error test case
```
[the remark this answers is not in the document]
myarmolinsky: Dropped this commit altogether. It was never supposed to be a ts-expect-error, and the change that led to this cannot be achieved at the moment.

src/models/application.ts
```diff
@@ -818,62 +805,19 @@ const getApplicationModel = function (
...
-		 * balena.models.application.generateApiKey('myorganization/myapp').then(function(apiKey) {
-		 * 	console.log(apiKey);
-		 * });
-		 *
-		 * @example
-		 * balena.models.application.generateApiKey(123).then(function(apiKey) {
-		 * 	console.log(apiKey);
-		 * });
-		 */
-		generateApiKey: async (
```
thgreasi, line 863: Should mention that this drops support for balenaOS <2.7.8

src/models/api-key.ts
```diff
@@ -39,39 +39,38 @@ const getApiKeysModel = function (
...
 		 *
 		 * @example
-		 * balena.models.apiKey.create(apiKeyName, apiKeyDescription).then(function(apiKey) {
+		 * balena.models.apiKey.create(apiKeyName, 2030-10-12, apiKeyDescription).then(function(apiKey) {
 		 * 	console.log(apiKey);
 		 * });
 		 */
 		async create(
 			name: string,
+			expiryDate: string | null,
```
thgreasi, line 60: Let's make the api key methods accept an object since they got lengthy and re-ordering could lead to accidental user errors.

Then we might want to throw `if (!('expiryDate' in options))`.

src/models/application.ts
```diff
@@ -882,55 +826,52 @@ const getApplicationModel = function (
...
 		 *
 		 * @example
-		 * balena.models.application.generateProvisioningKey(123, 'api key name', 'api key long description', '2030-01-01T00:00:00Z').then(function(key) {
+		 * balena.models.application.generateProvisioningKey(123, '2030-10-12', 'api key name', 'api key long description', '2030-01-01T00:00:00Z').then(function(key) {
 		 * 	console.log(key);
 		 * });
 		 */
 		generateProvisioningKey: async (
 			slugOrUuidOrId: string | number,
+			keyExpiryDate: string | null,
```
thgreasi, line 852: Should we change this to accept an object with all the parameters?

src/models/organization-membership.ts
```diff
@@ -241,58 +239,6 @@ const getOrganizationMembershipModel = function (
...
-		 * @param {Object} options - membership creation parameters
-		 * @param {String|Number} options.organization - organization handle (string), or id (number)
-		 * @param {String} options.username - the username of the balena user that will become a member
-		 * @param {String} [options.roleName="member"] - the role name to be granted to the membership
-		 *
-		 * @fulfil {Object} - organization membership
-		 * @returns {Promise}
-		 *
-		 * @example
-		 * balena.models.organization.membership.create({ organization: "myorg", username: "user123", roleName: "member" }).then(function(membership) {
```
thgreasi, line 264: How about changing the commit message to v ?
```
Drop deprecated method `organization-membership.create` in favor of `organization.invite.create`
```

src/models/application-membership.ts
```diff
@@ -225,7 +225,7 @@ const getApplicationMembershipModel = function (
 		 * @function
 		 * @memberof balena.models.application.membership
 		 *
-		 * @description This method adds a user to an application by their username.
+		 * @description This method adds a user to an application by their username if they are a member of the organization.
```
thgreasi, line 228: How about also adding a check in the function's implementation so that we enforce it right away?
You can also leave a `TODO` to remove that check once we add that check in the API.

This though would block you bumping the SDK in the UI until after the deadline date that we will announce... Was this the reason that you didn't add a runtime check?

src/models/organization.ts
```diff
@@ -186,18 +176,13 @@ const getOrganizationModel = function (
...
-			}
-			throw err;
-		}
+		const id = (
+			await sdkInstance.models.organization.get(handleOrId, { $select: 'id' })
+		).id;
+		await pine.delete<BalenaSdk.Organization>({
+			resource: 'organization',
+			id,
+		});
```
Page-, line 185: This could be combined into a single API call, but it isn't necessary to do that in a major

Page- (approved):

thgreasi (approved):

PR #1520, opened 2025-01-12
Add methods for the team model

JSReds:
Change-type: minor

Resolves: # 
HQ: <url> 
See: <url> 
Depends-on: <url> 

---
##### Contributor checklist

- [ ] Includes tests
- [ ] Includes typings

### DIFF
--- a/src/models/team-application-access.ts
+++ b/src/models/team-application-access.ts
@@ -1,0 +1,241 @@
+/*
+Copyright 2020 Balena
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+     http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+import type * as BalenaSdk from '..';
+import type { InjectedDependenciesParam } from '..';
+import * as errors from 'balena-errors';
+import { mergePineOptions } from '../util';
+
+const getTeamApplicationAccessModel = function (
+	deps: InjectedDependenciesParam,
+) {
+	const { pine, sdkInstance } = deps;
+
+	const getRoleId = async (roleName: BalenaSdk.ApplicationMembershipRoles) => {
+		const role = await pine.get({
+			resource: 'application_membership_role',
+			id: {
+				name: roleName,
+			},
+			options: {
+				$select: 'id',
+			},
+		});
+		if (!role) {
+			throw new errors.BalenaApplicationMembershipRoleNotFound(roleName);
+		}
+		return role.id;
+	};
+
+	/**
+	 * @summary Get all team applications access
+	 * @name getAllByTeam
+	 * @public
+	 * @function
+	 * @memberof balena.models.team.applicationAccess
+	 *
+	 * @description This method get all team application access.
+	 *
+	 * @param {Number} teamId - Required: the team id.
+	 * @param {Object} [options={}] - extra pine options to use
+	 *
+	 * @fulfil {Object[]} - team application access
+	 * @returns {Promise}
+	 *
+	 * @example
+	 * balena.models.team.applicationAccess.getAllByTeam(1239948).then(function(teamApplicationAccesses) {
+	 * 	console.log(teamApplicationAccesses);
+	 * });
+	 */
+	const getAllByTeam = async function (
+		teamId: number,
+		options: BalenaSdk.PineOptions<BalenaSdk.TeamApplicationAccess> = {},
+	): Promise<BalenaSdk.TeamApplicationAccess[]> {
+		const team = await sdkInstance.models.team.get(teamId, { $select: 'id' });
+
+		return sdkInstance.pine.get({
+			resource: 'team_application_access',
+			options: mergePineOptions(
+				{
+					$filter: {
+						team: team.id,
+					},
+				},
+				options,
+			),
+		});
+	};
+
+	/**
+	 * @summary Get team applications access
+	 * @name get
+	 * @public
+	 * @function
+	 * @memberof balena.models.team.applicationAccess
+	 *
+	 * @description This method get specific team application access.
+	 *
+	 * @param {Number} teamApplicationAccessId - Required: the team application access id.
+	 * @param {Object} [options={}] - extra pine options to use
+	 *
+	 * @fulfil {Object} - TeamApplicationAccess
+	 * @returns {Promise}
+	 *
+	 * @example
+	 * balena.models.team.applicationAccess.get(1239948).then(function(teamApplicationAccess) {
+	 * 	console.log(teamApplicationAccess);
+	 * });
+	 */
+	const get = async function (
+		teamApplicationAccessId: number,
+		options: BalenaSdk.PineOptions<BalenaSdk.TeamApplicationAccess> = {},
+	): Promise<BalenaSdk.TeamApplicationAccess | undefined> {
+		const teamApplicationAccess = await sdkInstance.pine.get({
+			resource: 'team_application_access',
+			id: teamApplicationAccessId,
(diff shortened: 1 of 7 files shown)


JSReds:
TODO:

Add getAll and assignToFleet

src/models/team.ts
```diff
@@ -0,0 +1,249 @@
...
+	 *   console.log(team);
+	 * });
+	 */
+	const create = async function (
+		organizationSlugOrId: string | number,
+		name: string,
+	): Promise<PinePostResult<Team>> {
+		const orgId = isId(organizationSlugOrId)
+			? organizationSlugOrId
+			: (await sdkInstance.models.organization.get(organizationSlugOrId))?.id;
```
thgreasi, line 57: I would be fine if we always did a GET request to confirm that the org exists and give a more helpful error in case it doesn't.
The `isId` in most cases is an optimization for methods that the UI uses in batch actions.

src/models/team.ts
```diff
@@ -0,0 +1,249 @@
...
+
+		if (orgId == null) {
+			throw new errors.BalenaOrganizationNotFound(organizationSlugOrId);
+		}
+
+		const teams = await sdkInstance.models.team.getAllByOrganization(orgId);
+
+		if (teams.some((t) => t.name === name)) {
```
thgreasi, line 65: How about using a `$filter` in `team.getAllByOrganization(orgId);` so that we don't have to fetch all teams of the org?
FWIW Given that the API does block this, I would be fine even if we didn't have this check.

src/models/team.ts
```diff
@@ -0,0 +1,249 @@
...
+	 * balena.models.team.get(123).then(function(team) {
+	 * 	console.log(team);
+	 * });
+	 */
+	const get = async function (
+		teamId: number,
+		options: BalenaSdk.PineOptions<BalenaSdk.Team> = {},
+	): Promise<BalenaSdk.Team> {
+		if (teamId == null) {
+			throw new errors.BalenaInvalidParameterError('id', teamId);
```
thgreasi, line 145:
```suggestion
			throw new errors.BalenaInvalidParameterError('teamId', teamId);
```

src/models/team.ts
```diff
@@ -0,0 +1,249 @@
...
+	 * @example
+	 * balena.models.team.getAllByOrganization('MyOrganizationHandle').then(function(teams) {
+	 * 	console.log(teams);
+	 * });
+	 */
+	const getAllByOrganization = function (
+		organizationSlugOrId: string | number,
+		options: BalenaSdk.PineOptions<BalenaSdk.Team> = {},
+	): Promise<BalenaSdk.Team[]> {
+		return pine.get({
```
thgreasi, line 107: * Mark it as async
* then add a throw in case organizationSlugOrId is unexpected

src/models/team.ts
```diff
@@ -0,0 +1,249 @@
...
+	 * balena.models.team.rename(123, 'MyNewTeamName');
+	 */
+	const rename = async function (
+		teamId: number,
+		newTeamName: string,
+	): Promise<void> {
+		const team = await get(teamId, {
+			$select: 'id',
+			$expand: { belongs_to__organization: { $select: 'id' } },
+		});
```
thgreasi, line 180: Let's avoid the JOIN :)
```suggestion
		const team = await get(teamId, {
			$select: ['id', 'belongs_to__organization'],
		});
```

src/models/team.ts
```diff
@@ -0,0 +1,249 @@
...
+		if (team == null) {
+			throw new Error(`Team not found: ${teamId}`);
+		}
+
+		const orgId =
+			Array.isArray(team.belongs_to__organization) &&
+			team.belongs_to__organization[0] &&
+			'id' in team.belongs_to__organization[0]
+				? team.belongs_to__organization[0].id
+				: undefined;
```
thgreasi, line 191: If we do
```
		const teamOptions = {
			$select: ['id', 'belongs_to__organization'],
		} satisfies PineOptions<Team>;
		const team = (await get(teamId, teamOptions)) as PineTypedResult<Team, typeof teamOptions>;
```
then this will be fully typed and we should be able to do just:
```suggestion
		const orgId = team.belongs_to__organization.__id;
```

src/models/team.ts
```diff
@@ -0,0 +1,249 @@
...
+		if (orgId == null) {
+			throw new Error(`Team does not belong to any organization: ${teamId}`);
+		}
+		const teams = await getAllByOrganization(orgId);
+
+		if (teams.some((t) => t.name === newTeamName)) {
+			throw new Error(
+				`A team with this name already exists in the organization. Organization: ${orgId}, Name: ${newTeamName}`,
+			);
+		}
```
thgreasi, line 202: You can probably also expand the teams w/ a $filter in the original pine query so that we do just 1 request.

src/models/team.ts
```diff
@@ -0,0 +1,249 @@
...
+	const remove = async function (teamId: number): Promise<void> {
+		try {
+			await pine.delete<BalenaSdk.Organization>({
+				resource: 'team',
+				id: teamId,
+			});
+		} catch (err) {
+			if (isNotFoundResponse(err)) {
+				throw new Error(`Team not found: ${teamId}`);
+			}
```
thgreasi, line 235: afaik the API will just return 200 when the team (or any OData DELETE to a resource that) doesn't exist, so let's remove the `try catch`
```suggestion
```

src/models/team.ts
JSReds: 🤦

src/models/team.ts
JSReds: do you mean by adding a pine.get(resource: 'organization', ...) or by checking if the argument is of type string | num ber ?

src/models/team-application-access.ts
```diff
@@ -0,0 +1,239 @@
...
+	 * @fulfil {Object} - TeamApplicationAccess
+	 * @returns {Promise}
+	 *
+	 * @example
+	 * balena.models.team.update(123, 'developer').then(function(teamApplicationAccess) {
+	 * 	console.log(teamApplicationAccess);
+	 * });
+	 */
+	const update = async function (
+		teamApplicationAccessId: string | number,
```
JSReds, line 188: Regarding the internal server error: the same issue applies here for `teamApplicationAccessId`

src/models/team.ts
thgreasi: I'm fine with either :)  IE
* Checking for number/string and throwing invalid parameter error
* doing just a `sdk.modelsorganization.get()`

src/models/team-application-access.ts
```diff
@@ -0,0 +1,239 @@
...
+	 *
+	 * @example
+	 * balena.models.team.applicationAccess.add(1239948, 456789, 'observer').then(function(teamApplicationAccess) {
+	 * 	console.log(teamApplicationAccess);
+	 * });
+	 */
+	const add = async function (
+		teamId: number,
+		applicationIdOrSlug: number | string,
+		roleName: BalenaSdk.ApplicationMembershipRoles,
```
JSReds, line 144: I'm not entirely sure, but do you think it's worth updating this method to accept one of the following options?

For example:

```
applicationIdOrSlug: number | string | number[] | string[]  
roleName: BalenaSdk.ApplicationMembershipRoles | BalenaSdk.ApplicationMembershipRoles[]  
```

Or alternatively:

```
applicationSlugOrIdAndRole: { idOrSlug: number | string, roleName: BalenaSdk.ApplicationMembershipRoles }[]  
```
(just consider the idea not names that I gave as an example)

And then wrapping it in a batch function from utils. What do you think?

tests/integration/models/team.spec.ts
```diff
@@ -0,0 +1,146 @@
...
+					);
+					ctx.newTeam2 = team;
+					expect(team).to.have.property('name', teamName);
+				});
+			});
+			describe('balena.models.team.rename()', function () {
+				it('should be rejected if the team does not exists', function () {
+					const promise = balena.models.team.rename(
+						999999,
+						`new_rename_${TEST_TEAM_NAME}`,
```
thgreasi, line 70:
```suggestion
						`${TEST_TEAM_NAME}_new_rename`,
```

thgreasi (approved):

PR #1534, opened 2025-05-05
os: Improve the checks against empty string release tag values

thgreasi:
Change-type: patch

Resolves: # 
HQ: <url> 
See: <url> 
Depends-on: <url> 

---
##### Contributor checklist

- [ ] Includes tests
- [ ] Includes typings

### DIFF
--- a/src/models/os.ts
+++ b/src/models/os.ts
@@ -134,7 +134,7 @@ const archCompatibilityMap: Partial<Dictionary<string[]>> = {
 const tagsToDictionary = (
 	tags: Array<Pick<ResourceTagBase, 'tag_key' | 'value'>>,
 ): Partial<Dictionary<string>> => {
-	const result: Dictionary<string> = {};
+	const result: Dictionary<string> = Object.create(null);
 	for (const { tag_key, value } of tags) {
 		result[tag_key] = value;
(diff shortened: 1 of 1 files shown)


src/models/os.ts
```diff
@@ -134,7 +134,7 @@ const archCompatibilityMap: Partial<Dictionary<string[]>> = {
 const tagsToDictionary = (
 	tags: Array<Pick<ResourceTagBase, 'tag_key' | 'value'>>,
 ): Partial<Dictionary<string>> => {
-	const result: Dictionary<string> = {};
+	const result: Dictionary<string> = Object.create(null);
```
myarmolinsky, line 137: What is the reason for this?

thgreasi: This is a more appropriate way to create a dictionary object since it doesn't include any prototypical chain.
This is more important in cases that you don't trust the values that will be used as keys, since they could be overriding properties defined in the protoypical chain, eg `__proto__`,  `toString` etc, which could be a security issues.
In this case it's not that much of an issue since we do control the hostApp tags, but I decided to go ahead with the change anyway.

myarmolinsky (approved):

PR #1548, opened 2025-06-02
Extend the retryRateLimitedRequests sdk option to also accept a function

thgreasi:
Change-type: minor


tests/integration/setup.ts
```diff
@@ -14,7 +14,8 @@ export let balenaSdkExports: typeof BalenaSdk;
 export const sdkOpts: BalenaSdk.SdkOptions = {
 	isBrowser: IS_BROWSER,
 	requestBatchingChunkSize: 5,
-	retryRateLimitedRequests: true,
+	// Retry ratelimited requests only if the ratelimiting is for less than 60s
+	retryRateLimitedRequests: (retryAfterMs) => retryAfterMs < 60_000,
```
thgreasi, line 18: We atm often get ratelimited for ages in some tests (eg invites) and the sdk tests looked like they were stalled/blocked for hours...
This should surface such cases right sooner and make clear what the issue is.

JSReds: the underscore is a typo i guess ?

PR #1551, opened 2025-06-11
Bump pinejs client core

otaviojacobi:
Resolves: # 
HQ: <url> 
See: <url> 
Depends-on: <url> 

---
##### Contributor checklist

- [ ] Includes tests
- [ ] Includes typings


src/index.ts
```diff
@@ -23,7 +23,8 @@ import type {
 import { globalEnv } from './util/global-env';
 import type { BalenaRequestError } from 'balena-errors';
 
-export type { Pine, PineStrict } from './pine';
+export type { PineClient } from './pine';
```
myarmolinsky, line 26:
```suggestion
export type { PineClient };
```

tests/integration/models/os.spec.ts
```diff
@@ -416,6 +417,7 @@ describe('OS model', function () {
 				expect(versionInfos[0]).to.have.nested.property(
 					'belongs_to__application[0].id',
 				);
+
```
myarmolinsky, line 420: Remove line breaks for 1 less file diff?

src/models/application-membership.ts
```diff
@@ -15,46 +15,44 @@ limitations under the License.
...
 import { mergePineOptions } from '../util';
+import type {
+	ODataOptionsWithoutCount,
+	OptionsToResponse,
+	ResourceAlternateKey,
+} from 'pinejs-client-core';
+import type { PickDeferred } from '@balena/abstract-sql-to-typescript';
 
 const RESOURCE = 'user_application_membership';
+type ApplicationMembership = BalenaModel[typeof RESOURCE];
```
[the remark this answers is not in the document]
otaviojacobi: Yes, I was just trying to reduce the changes/make it more readable as IMHO `ApplicationMembership` is way more readable than `UserIsMemberOfApplication`

src/models/api-key.ts
```diff
@@ -181,15 +190,21 @@ const getApiKeysModel = function (
...
 			slugOrUuidOrId: string | number,
-			options: BalenaSdk.PineOptions<BalenaSdk.ApiKey> = {},
-		): Promise<BalenaSdk.ApiKey[]> {
+			options: ODataOptionsWithoutCount<ApiKey['Read']> = {},
+		): Promise<Array<ApiKey['Read']>> {
 			const appOpts = {
-				$select: 'actor' as const,
-			};
+				$select: 'actor',
+			} as const;
```
myarmolinsky, line 197: Double-checking: is moving the `as const` to the whole opts instead of just the `$select` intentional?

src/models/application.ts
```diff
@@ -1179,19 +1188,16 @@ const getApplicationModel = function (
...
-				slugOrUuidOrId,
-				appOptions,
-			)) as PineTypedResult<Application, typeof appOptions>;
-			const body: SubmitBody<Application> = {
+			const application = await exports.get(slugOrUuidOrId, appOptions);
+			const body: Partial<Application['Write']> = {
 				should_track_latest_release: true,
 			};
-			const latestRelease = application.owns__release[0];
+			const latestRelease = application.owns__release?.[0];
```
myarmolinsky, line 1200: Why would `owns__release` come back undefined? Wouldn't it come back as an empty array at worst?

src/models/api-key.ts
otaviojacobi: Yes, this is a pattern we can use everywhere and simplify the thinking of where we should put stuff. It is also what we do on the api, so we just `as const` the whole options object everytime and things should just work

src/models/application.ts
otaviojacobi: No according to the api typings this can be an `Array<Release['Read']>` or not-existant (notice the `?` on the typing here): https://github.com/balena-io/balena-sdk/pull/1551/files#diff-fdfc0089fcbfd0be6aa662d8bffad07eddc414657cb2a66fd6f742b1752099e3R341

src/models/os.ts
```diff
@@ -382,16 +396,23 @@ const getOsModel = function (
...
 		deviceType: string,
 		options: TP & { includeDraft?: boolean },
-	): Promise<Array<ExtendedPineTypedResult<Release, OsVersion, TP>>>;
-	async function getAvailableOsVersions<TP extends PineOptions<Release>>(
-		deviceTypes: string[],
-		options: TP & { includeDraft?: boolean },
 	): Promise<
-		Dictionary<Array<ExtendedPineTypedResult<Release, OsVersion, TP>>>
+		// any[]
+		// Array<OptionsToResponse<Release['Read'], NonNullable<TP>, undefined>[number] & OsVersion>
```
myarmolinsky, line 406: comment 🚓

.prettierignore
```diff
@@ -1 +0,0 @@
-typing_tests
```
thgreasi, line 1: Isn't balena-lint complaining any more

src/auth.ts
```diff
@@ -456,6 +456,7 @@ const getAuth = function (
 			resource: 'user',
 			id,
 			body: {
+				// @ts-expect-error this actually exists on the model but it is only on the update allow list
```
thgreasi, line 459: Is that v right?
```suggestion
				// @ts-expect-error this actually exists but isn't part of the generated model but it is only on the update allow list and the allowlist atm needs it to also be in the Read as well
```
I would argue that we should augment the Write allowlist if it's easy, to avoid the confusion in case any consumer tries to do the same request manually.

src/models/application.ts
thgreasi: > No according to the api typings this can be an Array<Release['Read']> or not-existant (notice the ? on the typing here): https://github.com/balena-io/balena-sdk/pull/1551/files#diff-fdfc0089fcbfd0be6aa662d8bffad07eddc414657cb2a66fd6f742b1752099e3R341

I would say that it's now required b/c `as PineTypedResult<Application, typeof appOptions>;` was removed.

I would prefer it if we did v and avoid the `?.` as Matt says
````
const application = (await exports.get(slugOrUuidOrId, appOptions)) as NonNullable<
	OptionsToResponse<
		Application['Read'],
		typeof appOptions,
		typeof slugOrUuidOrId
	>
>
```
which I guess this is going to go away w/ your other PR that fully types

src/auth.ts
otaviojacobi: @Page- is against touching the allowlist iirc but I would prefer to defer this to afterwards

src/models/device-type.ts
```diff
@@ -434,8 +439,9 @@ const getDeviceTypeModel = function (deps: InjectedDependenciesParam) {
 		getInstructions: async (
 			deviceTypeSlugOrContract: string | Contract,
 		): Promise<Record<'Linux' | 'MacOS' | 'Windows', string[]> | string[]> => {
-			let contract: DeviceType['contract'];
+			let contract: Contract;
```
thgreasi, line 442:
```suggestion
			let contract: Contract | null = null;
```
since the `contract` JSON field returned by getBySlugOrName is nullable.
Otherwise eslint/TS might later start warning that the `?.` in `contract?.partials` isn't necessary.

src/models/application-invite.ts
```diff
@@ -157,14 +156,18 @@ const getApplicationInviteModel = function (
...
 					: undefined,
 			]);
-			type ApplicationInviteBase = Omit<ApplicationInvite, 'invitee'>;
+			type ApplicationInviteBase = Omit<
+				ApplicationInvite['Write'],
+				'invitee' | 'application_membership_role'
+			>;
 			type ApplicationInvitePostBody = ApplicationInviteBase & {
 				invitee: string;
+				application_membership_role?: ApplicationInvite['Write']['is_invited_to__application'];
```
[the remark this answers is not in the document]
otaviojacobi: Indded this is much simpler. I did the same for the organization-invite. Thanks

src/models/device-type.ts
```diff
@@ -253,8 +256,8 @@ const getDeviceTypeModel = function (deps: InjectedDependenciesParam) {
 		 * })
 		 */
 		async getAllSupported(
-			options?: PineOptions<DeviceType>,
-		): Promise<DeviceType[]> {
+			options?: ODataOptionsWithoutCount<DeviceType['Read']>,
+		): Promise<Array<DeviceType['Read']>> {
```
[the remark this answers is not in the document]
otaviojacobi: @thgreasi this won't go well with the generated typings commit (follow up). I think the real response is for the api to have knowledge of the Contract typings and generate it properly :)

src/models/application.ts
otaviojacobi: There you have it then folks :) I could avoid the `?` and no `PineTypedResult` as the entire function call now is typed

src/models/application.ts
```diff
@@ -455,40 +451,51 @@ const getApplicationModel = function (
...
-			)) as Application & {
-				owns__device: Array<DeviceWithServiceDetails<CurrentServiceWithCommit>>;
-			};
+			)) as unknown as NonNullable<
+				OptionsToResponse<
+					Application['Read'],
+					typeof baseOptions,
+					typeof slugOrUuidOrId
+				>
+			>;
```
thgreasi, line 476: I thought that if we did the cast upfront, then we could avoid the ts-expect-errors, but I was getting max call stack excheeded errors when I tried it :(
```suggestion
			)) as unknown as NonNullable<
				OptionsToResponse<
					Application['Read'],
					MergePineOptions<Application['Read'], typeof baseOptions, T>,
					typeof slugOrUuidOrId
				>
			> & {
				owns__device: Array<
					(typeof app)['owns__device'][number] & {
						current_services: Record<string, CurrentServiceWithCommit[]>;
					}
				>;
			};
```

src/models/application.ts
```diff
@@ -455,40 +451,51 @@ const getApplicationModel = function (
...
+
 			if (app.owns__device) {
```
thgreasi, line 478: btw I think that this `if` was just to make TS happy, so feel free to remove it.

src/models/credit-bundle.ts
```diff
@@ -96,9 +96,9 @@ const getCreditBundleModel = function ({
 			organization: string | number,
 			featureId: number,
 			creditsToPurchase: number,
-		): Promise<PinePostResult<CreditBundle>> => {
+		) => {
 			const orgId = await getOrgId(organization);
-			const body: SubmitBody<CreditBundle> = {
+			const body: CreditBundle['Write'] = {
```
thgreasi, line 101: optional: feel free to inline such cases, instead of having to explicitly define their type

src/models/device-type.ts
```diff
@@ -146,20 +150,18 @@ const getDeviceTypeModel = function (deps: InjectedDependenciesParam) {
...
-			options ??= {};
-
+			options?: T,
+		) {
 			if (idOrSlug == null) {
 				throw new errors.BalenaInvalidDeviceType(idOrSlug);
 			}
 
-			let deviceType: DeviceType | undefined;
+			let deviceType: OptionsToResponse<DeviceType['Read'], T, typeof idOrSlug>;
```
thgreasi, line 161: I don't know why, but it seems that this needs the `| undefined`

Image

src/models/device-type.ts
```diff
@@ -178,8 +180,8 @@ const getDeviceTypeModel = function (deps: InjectedDependenciesParam) {
 							},
 							options,
 						),
-					)
-				)[0];
+					)) as OptionsToResponse<DeviceType['Read'], T, undefined>
+				)[0] as OptionsToResponse<DeviceType['Read'], T, typeof idOrSlug>;
```
thgreasi, line 184: Am I right that this looks like a bug in pinejs-client typings?
In which case maybe better avoid the double cast and
// @ts-expect-error this should have been working, maybe it's a pinejs-client but or we are missing something 
deviceType = $deviceType

src/models/device.ts
```diff
@@ -2422,12 +2409,9 @@ const getDeviceModel = function (
...
 			 * });
 			 */
-			async getAllByApplication(
-				slugOrUuidOrId: string | number,
-				options?: PineOptions<DeviceVariable>,
-			): Promise<DeviceVariable[]> {
-				options ??= {};
-
+			async getAllByApplication<
+				T extends ODataOptionsWithoutCount<DeviceVariableModels['Read']>,
```
thgreasi, line 2413: We should probably use the exact type, since this is specifically for device config vars and not for any type that matches the DeviceVariableModels. It should be simpler & faster as well.
```suggestion
				T extends ODataOptionsWithoutCount<DeviceConfigVariable['Read']>,
```

src/models/device.ts
```diff
@@ -2579,12 +2563,9 @@ const getDeviceModel = function (
...
```
thgreasi, line 2567: like above
```suggestion
				T extends ODataOptionsWithoutCount<DeviceEnvironmentVariable['Read']>,
```

src/models/device.ts
```diff
@@ -91,8 +86,21 @@ export * as DeviceState from '../types/device-state';
 export type { DeviceOverallStatus as OverallStatus } from '../types/device-overall-status';
 export type { SupervisorStatus } from './device.supervisor-api.partial';
 
+type DeviceVariableModels = BalenaModel[{
```
thgreasi, line 89: Let me raise that even though this looks fancy, the standard pattern would either be defining the structure

src/util/dependent-resource.ts
```diff
@@ -20,58 +20,70 @@ key-value resources directly attached to a parent (e.g. tags, config variables).
...
-}
+type DependentResource = {
+	[K in StringKeyof<BalenaModel>]: 'Read' extends keyof BalenaModel[K]
+		? BalenaModel[K]['Read'] extends { id: number; value: string }
+			? BalenaModel[K]['Write'] extends { value: string }
+				? K
+				: never
+			: never
+		: never;
+}[StringKeyof<BalenaModel>];
```
thgreasi, line 40: I guess this can be simplified to a single structural check. Not sure whether it's faster, but it looks easier to read imo
```suggestion
type DependentResource = {
	[K in StringKeyof<BalenaModel>]: BalenaModel[K] extends {
		'Read': {
			id: number;
			value: string;
		},
		'Write': { value: string; }
	}
		? K
		: never
}[StringKeyof<BalenaModel>];
```

src/util/dependent-resource.ts
```diff
@@ -82,16 +94,17 @@ export function buildDependentResource<T extends DependentResource>(
...
 				options: {
 					$select: 'value',
 					$filter: {
 						[parentResourceName]: id,
 						[resourceKeyField]: key,
 					},
-				},
-			});
+				} as ODataOptionsWithoutCount<BalenaModel[T]['Read']>,
+			})) as unknown as Array<{ value: string } | undefined>;
```
thgreasi, line 106: This can reduce the casts v to the very minimum:
```suggestion
			const [result] = (await pine.get({
				resource: resourceName as DependentResource,
				options: {
					$select: 'value',
					$filter: {
						[parentResourceName]: id,
						[resourceKeyField]: key,
					} as ODataOptionsWithoutCount<BalenaModel[DependentResource]['Read']>['$filter'],
				},
			}));
```
the `as DependentResource` is avoiding the `T can be any type that extends DependentResource`, but it shouldn't matter since atm DependentResource is a union of very specific strings.

thgreasi: somehow it looks like this is also working with `satisfies`... but I don't trust that it's correct tbh

src/util/device-service-details.ts
```diff
@@ -112,9 +109,13 @@ function getSingleInstallSummary(
 export const generateCurrentServiceDetails = <
 	TCurrentService extends CurrentService = CurrentService,
 >(
-	rawDevice: Device,
+	rawDevice: OptionsToResponse<
```
thgreasi, line 112: Given that now always expand the `is_provided_by__release` and you also fully type the provided object, we can also get rid of the TCurrentService 🎉 , so these two functions can be simplified to v
```ts
function getSingleInstallSummary(
	rawData: OptionsToResponse<
		ImageInstall['Read'],
		typeof getCurrentServiceDetailsPineExpand.image_install,
		undefined
	>[number],
) {
	const image = rawData.installs__image[0];
	const service = image.is_a_build_of__service[0];

	const release = rawData.is_provided_by__release[0];
	const releaseInfo = {
		commit: release?.commit,
		raw_version: release?.raw_version,
		release_id: release?.id,
	};

	const result: CurrentServiceWithCommit & WithServiceName = {
		...rawData,
		service_id: service.id,
		// add this extra property to make grouping the services easier
		service_name: service.service_name,
		image_id: image.id,
		...releaseInfo,
	};

	if ('installs__image' in result) {
		delete result.installs__image;
	}
	if ('is_provided_by__release' in result) {
		delete result.is_provided_by__release;
	}

	return result;
}

export const generateCurrentServiceDetails = (
	rawDevice: OptionsToResponse<
		Device['Read'],
		{ $expand: typeof getCurrentServiceDetailsPineExpand },
		undefined
	>[number],
): DeviceWithServiceDetails<CurrentServiceWithCommit> => {
	const installs = rawDevice.image_install.map(getSingleInstallSummary);

	// Essentially a groupBy(installs, 'service_name')
	// but try making it a bit faster for the sake of large fleets.
	// Uses Object.create(null) so that there are no inherited properties
	// which could match service names
	const currentServicesGroupedByName: Record<string, CurrentServiceWithCommit[]> =
		Object.create(null);
	for (const containerWithServiceName of installs) {
		const { service_name } = containerWithServiceName;
		let serviceContainerGroup: CurrentServiceWithCommit[];
		if (currentServicesGroupedByName[service_name] == null) {
			serviceContainerGroup = [];
			currentServicesGroupedByName[service_name] = serviceContainerGroup;
		} else {
			serviceContainerGroup = currentServicesGroupedByName[service_name];
		}

		const container: CurrentServiceWithCommit & Partial<WithServiceName> =
			containerWithServiceName;

		// remove the extra property that we added for the grouping
		delete container.service_name;
		serviceContainerGroup.push(container);
	}

	for (const serviceName in currentServicesGroupedByName) {
		if (currentServicesGroupedByName[serviceName]) {
			currentServicesGroupedByName[serviceName].sort((a, b) => {
				return b.install_date.localeCompare(a.install_date);
			});
		}
	}

	const device = rawDevice as DeviceWithServiceDetails<CurrentServiceWithCommit>;
	device.current_services = currentServicesGroupedByName;
	return device;
};
```

thgreasi: Updated the src a bit

tests/integration/models/device-type.spec.ts
```diff
@@ -159,7 +160,10 @@ describe('Device Type model', function () {
 					{ $select: 'contract' },
 				);
 				assertExists(contract);
-				const result = await balena.models.deviceType.getInstructions(contract);
+				// @ts-expect-error - parsed contract will be a Contract
+				const $contract = contract as Contract;
```
thgreasi, line 164: This is going to become super frustrating while using these methods....
Maybe we should make `getInstructions` access `Contract | typeof DeviceType['Read']['contract']`  just to make it a bit easier, and have it throw in case it doesn't look like a Contract.

tests/integration/models/release.spec.ts
```diff
@@ -263,7 +263,7 @@ describe('Release Model', function () {
 		});
 
 		describe(`given ${uniquePropertyNames.length} draft releases`, function () {
-			const testReleaseByField: Dictionary<BalenaSdk.Release> = {};
+			const testReleaseByField: Dictionary<BalenaSdk.Release['Read']> = {};
```
thgreasi, line 266: If you do v
```suggestion
			const testReleaseByField: Dictionary<PickDeferred<BalenaSdk.Release['Read'],
				| 'id'
				| 'commit'
				| 'raw_version'
				| 'belongs_to__application'
			>> = {};
```
then we can drop the other two `as BalenaSdk.Release['Read']` casts that were added.

src/auth.ts
Page-: Yes, I think having something you can write but not read is a real anti-pattern and I think the correct answer is they should not exist, since in practice they're actions and not really writes to the model (they might trigger writes to the model as part of executing the action, but the action is the desired outcome because you can't observe the write itself..)

src/models/device.ts
```diff
@@ -3135,22 +3119,30 @@ const getDeviceModel = function (
...
+				T extends ODataOptionsWithoutCount<DeviceHistory['Read']>,
+			>(
 				slugOrUuidOrId: string | number,
 				{
 					fromDate = subDays(new Date(), DEFAULT_DAYS_OF_REQUESTED_HISTORY),
 					toDate,
 					...options
-				}: PineOptions<DeviceHistory> & { fromDate?: Date; toDate?: Date } = {},
-			): Promise<DeviceHistory[]> {
+				}: T & {
```
Page-, line 3130: It seems like it would make sense to separate the pinejs-client options from the balena-sdk options into two separate parameters, it would make the typing/code cleaner and when using it I think it would also make it clearer that `fromDate`/`toDate` are not things that exist in the model or are expected to potentially pass through to pinejs-client

(and elsewhere that follows the pattern)

src/models/device.ts
```diff
@@ -630,7 +612,9 @@ const getDeviceModel = function (
 			if (device == null) {
 				throw new errors.BalenaDeviceNotFound(uuidOrId);
 			}
-			return device as Device;
+			return device as NonNullable<
```
Page-, line 615: I think it would make sense to not have this cast (which can hide issues) and instead change the above to:
```ts
			let device; NonNullable<
				OptionsToResponse<Device['Read'], T, typeof uuidOrId>
			>;
```
so that attempts to assign something that doesn't match to `device` will fail at the root cause

src/models/device.ts
```diff
@@ -621,7 +603,7 @@ const getDeviceModel = function (
 						},
 						options,
 					),
-				});
+				})) as OptionsToResponse<Device['Read'], T, undefined>;
```
Page-, line 606: Is this cast necessary?

src/models/application.ts
```diff
@@ -385,7 +380,7 @@ const getApplicationModel = function (
 						},
 						options,
 					),
-				});
+				})) as OptionsToResponse<Application['Read'], T, undefined>;
```
Page-, line 383: Is this cast necessary?

src/models/application-membership.ts
```diff
@@ -270,34 +262,45 @@ const getApplicationMembershipModel = function (
 						},
 					},
 				},
-			} satisfies PineOptions<Application>;
+			} as const;
+
 			const [{ id, organization }, roleId] = await Promise.all([
 				getApplication(application, appOptions) as Promise<
```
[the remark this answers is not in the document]
otaviojacobi: Yes, without it the infered type of $select is string and we miss the whole type resolution

src/models/device.ts
otaviojacobi: Well I think we can go with 
```
let device: OptionsToResponse<Device['Read'], T, typeof uuidOrId> | OptionsToResponse<Device['Read'], T, undefined>[number];

```
as for the particular case here these are not mutually exchangeable

src/util/request-batching.ts
```diff
@@ -161,12 +166,11 @@ export function batchResourceOperationFactory<
 							? ['uuid']
 							: []),
 						...(groupByNavigationPoperty ? [groupByNavigationPoperty] : []),
-					] as Array<PineSelectableProps<T>>,
-					$filter: resourceFilter as PineFilter<T>,
+					] as ODataOptionsWithoutCount<T>['$select'],
```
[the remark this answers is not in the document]
otaviojacobi: I think it can be even better. if I go with `['uuid'] as const` the whole tree satisfies nd I don't need any casting

src/models/team-application-access.ts
```diff
@@ -185,8 +183,8 @@ const getTeamApplicationAccessModel = function (
 	 */
 	const update = async function (
 		teamApplicationAccessId: number,
-		roleName: BalenaSdk.ApplicationMembershipRoles,
-	): Promise<void> {
+		roleName: string,
```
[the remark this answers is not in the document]
otaviojacobi: Solved here and for Team & OrganizationMemberships aswell

src/models/service.ts
```diff
@@ -90,20 +97,20 @@ const getServiceModel = ({
...
 				$select: 'id',
 				$expand: {
-					service: options,
+					service: options ?? {},
 				},
 			},
 		);
-		return service!;
+		// Need to manually cast to help typescript with the `options ?? {}` which could be problematic if null
+		return service as OptionsToResponse<Service['Read'], T, undefined>;
```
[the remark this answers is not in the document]
otaviojacobi: See above

src/models/release.ts
```diff
@@ -349,7 +352,7 @@ const getReleaseModel = function (
 				},
 				options,
 			),
-		);
+		)) as OptionsToResponse<Release['Read'], T, undefined>;
```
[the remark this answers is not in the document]
otaviojacobi: Yes, for the same reason of the reply https://github.com/balena-io/balena-sdk/pull/1551#discussion_r2183726411

src/models/release.ts
```diff
@@ -256,7 +261,7 @@ const getReleaseModel = function (
 
 		release.images = images
 			.map(function ({ is_a_build_of__service, ...imageData }) {
-				const image: BalenaSdk.ReleaseWithImageDetails['images'][number] = {
+				const image: ReleaseWithImageDetails['images'][number] = {
```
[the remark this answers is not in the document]
otaviojacobi: In fact, the whole type assertion here can be removed

src/models/release.ts
```diff
@@ -213,40 +213,45 @@ const getReleaseModel = function (
...
-		} satisfies BalenaSdk.PineOptions<BalenaSdk.Release>;
+		} as const;
 
 		const rawRelease = (await get(
 			commitOrIdOrRawVersion,
-			mergePineOptions(baseReleaseOptions, options.release),
-		)) as PineTypedResult<Release, typeof baseReleaseOptions>;
-		const release = rawRelease as BalenaSdk.ReleaseWithImageDetails;
+			mergePineOptions(baseReleaseOptions, options.release ?? {}),
+		)) as NonNullable<
```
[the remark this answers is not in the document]
otaviojacobi: Yes, for the same reason of the reply https://github.com/balena-io/balena-sdk/pull/1551#discussion_r2183726411

src/models/os.ts
```diff
@@ -1068,16 +1095,13 @@ const getOsModel = function (
 						{ revision: 'desc' },
 					],
 				},
-				options,
+				options ?? {},
```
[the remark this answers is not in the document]
otaviojacobi: duplicate

src/models/application.ts
```diff
@@ -394,7 +389,9 @@ const getApplicationModel = function (
 			if (application == null) {
 				throw new errors.BalenaApplicationNotFound(slugOrUuidOrId);
 			}
-			return application;
+			return application as NonNullable<
+				OptionsToResponse<Application['Read'], T, typeof slugOrUuidOrId>
+			>;
```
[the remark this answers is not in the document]
otaviojacobi: Did the same resolution as for device

src/util/index.ts
```diff
@@ -168,18 +308,18 @@ const mergeExpandOptions = <T>(
...
+): ResourceExpand<T> => {
 	if (expandOption == null) {
 		return {};
 	}
 
 	if (typeof expandOption === 'string') {
 		return {
 			[expandOption]: {},
-		} as Pine.ResourceExpand<T>;
+		} as ResourceExpand<T>;
```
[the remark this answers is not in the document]
otaviojacobi: This, sadly, does not satisfies as typescript will narrow down the `[expandOption: string]`. Even if I explictely cast something like

```
return {
			[expandOption as ExpandableStringKeyOf<T>]: {},
		}
``` 
(which Eslint correctly complains that is a unnecessary assertion) TS will continue narrowing it down to `string` which makes it not satisfiable

src/models/os.ts
```diff
@@ -17,28 +17,27 @@ limitations under the License.
...
+import type {
+	ODataOptionsWithoutCount,
+	OptionsToResponse,
+} from 'pinejs-client-core';
+
+type ResourceTagBase = {
+	id: number;
+	tag_key: string;
+	value: string;
+};
```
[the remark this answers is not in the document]
otaviojacobi: I like it. It is also copied in two places, I will put this into a shared typing for both with the new definition

src/models/service.ts
```diff
@@ -207,7 +213,7 @@ const getServiceModel = ({
 									},
 								},
 							},
-							$orderby: 'name asc',
+							$orderby: { name: 'asc' },
```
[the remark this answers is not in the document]
otaviojacobi: done

src/models/device.ts
otaviojacobi: done (follow up commit I will pull up when finishing rebasing)

src/models/release.ts
```diff
@@ -213,40 +213,45 @@ const getReleaseModel = function (
...
 			},
-		} satisfies BalenaSdk.PineOptions<BalenaSdk.Release>;
+		} as const;
 
 		const rawRelease = (await get(
 			commitOrIdOrRawVersion,
-			mergePineOptions(baseReleaseOptions, options.release),
-		)) as PineTypedResult<Release, typeof baseReleaseOptions>;
-		const release = rawRelease as BalenaSdk.ReleaseWithImageDetails;
+			mergePineOptions(baseReleaseOptions, options.release ?? {}),
```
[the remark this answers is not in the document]
otaviojacobi: @Page-  These runtime changes can be replaced with 

```
mergePineOptions(baseReleaseOptions, options.release) as typeof baseReleaseOptions
```

and equivalent from image.

The reasoning for that is that in order to have the inference run while the `T` generic is still not resolved by `MergePineOptions`, we need this cast so that the return type is known and can be used to run the inner logic.

This particular function is also an exception of most of the other ones as it is deleting stuff (`release_image`) and doing custom property (e.g. renaming `is_created_by__user` into `user`) so inference here is not the most important case AFAICT

src/models/release.ts
```diff
@@ -213,40 +213,45 @@ const getReleaseModel = function (
...
 			},
-		} satisfies BalenaSdk.PineOptions<BalenaSdk.Image>;
+		} as const;
 
 		const baseReleaseOptions = {
 			$expand: {
 				release_image: {
 					$expand: {
-						image: mergePineOptions(baseImageOptions, options.image),
+						image: mergePineOptions(baseImageOptions, options.image ?? {}),
```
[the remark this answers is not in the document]
otaviojacobi: See above

src/models/application.ts
```diff
@@ -455,40 +450,39 @@ const getApplicationModel = function (
...
+		>(slugOrUuidOrId: string | number, options?: T) {
+			const baseOptions = {
+				$expand: {
+					owns__device: {
+						$expand: getCurrentServiceDetailsPineExpand,
+					},
 				},
+			} as const;
+
+			const serviceOptions: typeof baseOptions = mergePineOptions(
```
otaviojacobi, line 464: This is needed as the current `MergePineOptions` cannot infer while the generic T is not defined that the return type will indeed have all the properties needed to resolve the query, but it knows that it satisfies so we can simply declare the typing so it works withing function scope

typing_tests/pine-typed-result.ts
```diff
@@ -1,106 +1,98 @@
...
-	aString = result.device_name;
-	aNumber = result.belongs_to__application.__id;
+	aNumberOrUndefined = result.is_of__device_type?.__id;
 	aNumberOrUndefined = result.is_pinned_on__release?.__id;
 
 	// @ts-expect-error test case
 	aNumber = result.is_pinned_on__release.__id;
 	// @ts-expect-error test case
-	aAny = result.device_tag;
+	anArray = result.device_tag;
```
thgreasi, line 36:
```suggestion
	// TODO: Ideally the prop shouldn't be there at all
	anUndefined = result.device_tag;
```

thgreasi, line 36:
```suggestion
	// TODO: Ideally the prop shouldn't exist at all
	anUndefined = result.device_tag;
```

typing_tests/pine-typed-result.ts
```diff
@@ -167,31 +174,35 @@ export let aStringOrUndefined: string | undefined;
...
 	const result: deviceOptionsExpandNavigationResourceString = {} as any;
 
-	aNumber = result.belongs_to__application[0].id;
+	aNumber = result.is_of__device_type[0].id;
 	aNumber = result.id;
 	aString = result.device_name;
-
-	// @ts-expect-error test case
-	aAny = result.belongs_to__application[1];
+	aAny = result.is_of__device_type[1];
```
thgreasi, line 192:
```suggestion
	// TODO: TS should complain about is_of__device_type[1] since the prop should ideally be a single item tuple
	aNumber = result.is_of__device_type[1].id;
```

typing_tests/pine-typed-result.ts
```diff
@@ -473,23 +506,25 @@ export let aStringOrUndefined: string | undefined;
...
+			number
+		>
 	>;
 
 	const result: deviceOptionsExpandNavigationResourceString = {} as any;
 
 	aNumber = result.id;
 	// Errors, since it could be an OptionalNavigationResource
-	// @ts-expect-error test case
 	aStringOrUndefined = result.is_pinned_on__release[0].commit;
```
thgreasi, line 528: Same as earlier, add a todo that this is nullable and should complain since we are not using `?.[0]`

typing_tests/pine-options.ts
```diff
@@ -1,545 +1,586 @@
...
-export const unkown$expandProps: BalenaSdk.PineOptions<BalenaSdk.Application> =
-	{
-		$expand: {
-			// @ts-expect-error test case
+export const unkownODataPropInside$expand: BalenaSdk.Pine.ODataOptions<
+	BalenaSdk.Application['Read']
+> = {
+	$expand: {
+		owns__device: {
+			// // @ts-expect-error test case - new pine client is not able to detect invalid double expand
```
thgreasi, line 51:
```suggestion
			// TODO this should ideally be failing but atm pine client is not able to detect invalid double expand
```

typing_tests/pine-options.ts
```diff
@@ -1,545 +1,586 @@
...
+> = {
+	// @ts-expect-error test case
+	$orderby: 'id ascending',
+};
 
-export const string$OrderbyInvalidDirectionCase: BalenaSdk.PineOptions<BalenaSdk.Application> =
-	{
-		// @ts-expect-error test case
-		$orderby: { id: 'Asc' },
-	};
```
thgreasi, line 514: Should we also add these two back, to make sure that the `dir` is checked?
If of course they are not failing, we can add then w/ a todo instead of a ts-expect-error
```
export const object$OrderbyInvalidDirection: BalenaSdk.PineOptions<BalenaSdk.Application> =
	{
		// @ts-expect-error test case
		$orderby: { id: 'ascending' },
	};


export const object$OrderbyInvalidDirectionCase: BalenaSdk.PineOptions<BalenaSdk.Application> =
	{
		// @ts-expect-error test case
		$orderby: { id: 'Asc' },
	};
```

typing_tests/pine-options.ts
```diff
@@ -564,186 +607,216 @@ export const appOptionsEValid5: BalenaSdk.PineOptions<BalenaSdk.Application> = {
...
+	| 'has_direct_access_to__application'
+	| 'user_profile';
 
 export const userExpandablePropsTest: Equals<
-	BalenaSdk.PineExpandableProps<BalenaSdk.User>,
+	BalenaSdk.Pine.ExpandableStringKeyOf<BalenaSdk.User['Read']>,
 	UserExpandablePropsExpectation
 > = EqualsTrue;
 
-export const deviceIsRunningReleaseAssociatedResourceType: Equals<
```
thgreasi, line 629: :(
I guess it's too hard to replicate this test w/ the new typings

typing_tests/pine-options.ts
otaviojacobi: They are there... it is just github diff missleading you

typing_tests/pine-options.ts
```diff
@@ -976,18 +1052,30 @@ export const anyObjectOptionsValid1: BalenaSdk.PineOptions<AnyObject> = {
...
 
-export const string$OrderbyNestedProp: BalenaSdk.PineOptions<BalenaSdk.Organization> =
-	{
-		$orderby: 'application_type/slug asc',
-	};
+export const string$OrderbyStringProp: BalenaSdk.Pine.ODataOptions<
+	BalenaSdk.Organization['Read']
+> = {
+	// @ts-expect-error - prefer object notation for orderby
+	$orderby: 'app_name asc',
```
thgreasi, line 1080:
```suggestion
	$orderby: 'handle asc',
```

typing_tests/pine-options.ts
```diff
@@ -976,18 +1052,30 @@ export const anyObjectOptionsValid1: BalenaSdk.PineOptions<AnyObject> = {
...
+> = {
 	$orderby: { id: 'asc' },
 };
 
-export const string$OrderbyDesc: BalenaSdk.PineOptions<BalenaSdk.Application> =
-	{
-		$orderby: { id: 'asc' },
-	};
+export const string$OrderbyNestedProp: BalenaSdk.Pine.ODataOptions<
+	BalenaSdk.Organization['Read']
```
thgreasi, line 1070:
```suggestion
	BalenaSdk.Application['Read']
```

typing_tests/pine-options.ts
```diff
@@ -976,18 +1052,30 @@ export const anyObjectOptionsValid1: BalenaSdk.PineOptions<AnyObject> = {
...
+	// @ts-expect-error - prefer object notation for orderby
+	$orderby: 'application_type/slug asc',
+};
 
-export const string$OrderbyNestedProp: BalenaSdk.PineOptions<BalenaSdk.Organization> =
-	{
-		$orderby: 'application_type/slug asc',
-	};
+export const string$OrderbyStringProp: BalenaSdk.Pine.ODataOptions<
+	BalenaSdk.Organization['Read']
```
thgreasi, line 1077:
```suggestion
	BalenaSdk.Application['Read']
```

typing_tests/pine-params.ts
```diff
@@ -211,13 +157,13 @@ await (async () => {
 	});
 	aNumber = result.id;
 	aString = result.device_name;
-	aNumber = result.belongs_to__application.__id;
+	aNumberOrUndefined = result.belongs_to__application?.__id;
```
thgreasi, line 160: Can we do the device type thing like in the other file?
```suggestion
	aNumber = result.is_of__device_type.__id;
```

typing_tests/pine-params.ts
```diff
@@ -211,13 +157,13 @@ await (async () => {
...
 	aNumberOrUndefined = result.is_pinned_on__release?.__id;
 
 	// @ts-expect-error test case
 	aNumber = result.is_pinned_on__release.__id;
 	// @ts-expect-error test case
-	aAny = result.device_tag;
+	anArray = result.device_tag;
```
thgreasi, line 166: Like in the other file 🙏 
```suggestion
	// TODO: Ideally the prop shouldn't be there at all
	anUndefined = result.device_tag;
```

typing_tests/pine-params.ts
```diff
@@ -607,12 +543,17 @@ await (async () => {
 				},
 			},
 			...(maybeService != null && {
-				is_a_build_of__service: {
-					$any: {
-						$alias: 'iabos',
-						$expr: {
-							iabos: {
-								service_name: maybeService,
+				$expand: {
```
thgreasi, line 546: I think the intention of the test (which I realize I failed to do) was to have it as a conditional filter like:
```
			$filter: {
				release_image: {
					// ...
				},
				...(maybeService != null && {
					is_a_build_of__service: {
						$any: {
							$alias: 'iabos',
							$expr: {
								iabos: {
									service_name: maybeService,
								},
							},
						},
					},
				}),
			},
```
Great find 👍

typing_tests/pine-typed-result.ts
```diff
@@ -1,106 +1,99 @@
...
 // $select
 
 {
-	type deviceOptionsNoProps = PineClient.TypedResult<
-		BalenaSdk.Device,
-		undefined
+	type deviceOptionsNoProps = NonNullable<
+		BalenaSdk.Pine.OptionsToResponse<
+			BalenaSdk.Device['Read'],
+			{}, // eslint-disable-line @typescript-eslint/no-empty-object-type
```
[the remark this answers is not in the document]
otaviojacobi: @thgreasi This is the only file with it, what do you think of just disabling this rule for this file? IMO 
`is_pinned_on__release: {};` reads easier than `is_pinned_on__release: emptyObject;`

typing_tests/merge-pine-options.ts
```diff
@@ -0,0 +1,497 @@
...
+	aStringOrNull = noProps[0].description;
+	aStringOrNull = noProps[0].name;
+	aNumber = noProps[0].is_of__actor.__id;
+
+	const narrowDownSelectString = await sdk.models.apiKey.getAll({
+		$select: 'id',
+	});
+	aNumber = narrowDownSelectString[0].id;
+	// @ts-expect-error - test case
+	aString = narrowDownSelectString[0].created_at;
```
thgreasi, line 33: For the cases that you want to make sure that the error is originating from the right side of the expression, you might better use `aAny = ...` (or a `console.log` I guess, but I think I prefer the aAny)

typing_tests/merge-pine-options.ts
```diff
@@ -0,0 +1,497 @@
...
+	// @ts-expect-error - test case
+	aStringOrNull = narrowDownSelectArrayWithNavigation[0].name;
+})();
+
+// $expand properties
+await (async () => {
+	const noExpand = await sdk.models.application.getAll();
+	aNumberOrUndefined = noExpand[0].is_for__device_type?.__id;
+	aNumberOrUndefined = noExpand[0].should_be_running__release?.__id;
+	anUndefined = noExpand[0].owns__device;
```
thgreasi, line 91: 🙏  TODO ideally this prop shouldn't be here at all? 🙏

typing_tests/merge-pine-options.ts
```diff
@@ -0,0 +1,497 @@
...
+	});
+
+	// @ts-expect-error - test case
+	aNumber = objectExpandWithSelect[0].is_for__device_type[0].id;
+	aString = objectExpandWithSelect[0].is_for__device_type[0].name;
+	aString = objectExpandWithSelect[0].is_for__device_type[0].slug;
+
+	aNumberOrUndefined =
+		objectExpandWithSelect[0].should_be_running__release?.__id;
+	aNumber = objectExpandWithSelect[0].owns__device[0].id;
```
thgreasi, line 128: NIT: Maybe better use a random other prop than `id` (eg `name`), even though we know there isn't any special handling for `id`.

typing_tests/merge-pine-options.ts
```diff
@@ -0,0 +1,502 @@
...
+							$expand: {
+								installs__image: {
+									$select: 'is_stored_at__image_location',
+								},
+							},
+						},
+					},
+				},
+				owns__release: {
+					$select: 'id',
```
thgreasi, line 210: Let's also add v since it's in the select
```suggestion

aString = appWithNestedExpand.app_name
```

typing_tests/merge-pine-options.ts
```diff
@@ -0,0 +1,502 @@
...
+						$select: ['username'],
+					},
+				},
+			},
+		},
+	});
+	aNumber =
+		// @ts-expect-error - test case
+		objectExpandNestedWithSelect[0].owns__device[0].belongs_to__user[0].id;
+
```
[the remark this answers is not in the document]
otaviojacobi: I kept both as 

```
	// @ts-expect-error - test case
	aString = apps[0].service_name;
	// @ts-expect-error - test case
	aString = apps[0].app_name;
```

typing_tests/merge-pine-options.ts
otaviojacobi: I don't think it makes sense to change the prop as you said yourself nothing is special for `id` (at least on the current approach)

typing_tests/merge-pine-options.ts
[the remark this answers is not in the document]
thgreasi: Oh I guess it should be
```
aString = apps[0].app_name;
```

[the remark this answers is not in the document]
otaviojacobi: @thgreasi no? I think this is correct as app_name is not selected

typing_tests/pine-params.ts
thgreasi: @otaviojacobi kind reminder to update this one as well

typing_tests/merge-pine-options.ts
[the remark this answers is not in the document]
otaviojacobi: I see. Done.

PR #1553, opened 2025-06-16
Run CI on supported node version (20)

otaviojacobi:
Change-type: patch

Resolves: # 
HQ: <url> 
See: <url> 
Depends-on: <url> 

---
##### Contributor checklist

- [ ] Includes tests
- [ ] Includes typings


thgreasi:
fwiw I would have preferred if we did this along w/ the major that drops support for node18, b/c atm we do mention that we support node18 in package.json, but we are no longer testing for it

PR #1559, opened 2025-07-11
Support Supervisor managed HUP

kb2ma:
Alternate implementation of #1498.

Adds variable for cutoff version.


src/models/device.ts
```diff
@@ -377,36 +387,91 @@ const getDeviceModel = function (
 			groupByNavigationPoperty: 'is_of__device_type',
 			fn: async (devices, deviceTypeId) => {
 				const dt = await getDeviceType(deviceTypeId);
+				const toPinOSRelease = new Map<string, number>();
```
thgreasi, line 390: It will probably be better to make this a of of releaseId to deviceUuids.
```suggestion
				const deviceUuidsByTargetReleaseId = new Map<number, string[]>();
```

kb2ma:
Obsolete

PR #1563, opened 2025-08-08
Add release_asset

otaviojacobi:
Change-type: minor

Resolves: # 
HQ: <url> 
See: <url> 
Depends-on: <url> 

---
##### Contributor checklist

- [ ] Includes tests
- [ ] Includes typings


src/models/release-asset.ts
```diff
@@ -0,0 +1,468 @@
...
+			await pine.delete({
+				resource: 'release_asset',
+				id: await getId(id),
+			});
+		},
+	};
+	return exports;
+};
+
+export default getReleaseAssetModel;
```
thgreasi, line 468: Any reason you decided to go w/ default exports rather than named exports like in the other helpers?

src/models/release-asset.ts
```diff
@@ -0,0 +1,468 @@
...
+			ETag,
+		};
+	};
+
+	const uploadParts = async (
+		asset: File | string,
+		parts: UploadPart[],
+		uploadParams: Required<Pick<UploadParams, 'chunkSize' | 'parallelUploads'>>,
+	) => {
+		const limit = pLimit(uploadParams.parallelUploads);
```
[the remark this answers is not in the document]
thgreasi: Also, is there any reason that you didn't choose to use `p-map`?
It seems that it would fix better in this case, since you don't seem to need the extra flexibility that p-limit offers.

src/models/release-asset.ts
```diff
@@ -0,0 +1,468 @@
...
+import type { ReleaseRawVersionApplicationPair } from './release';
+import pLimit from 'p-limit';
+import { getType } from 'mime';
+
+const getMimeType = (filePath: string): string => {
+	return getType(filePath) ?? 'application/octet-stream';
+};
+
+const getFileBasename = (filePath: string): string => {
+	return filePath.split('/').pop() ?? filePath.split('\\').pop() ?? filePath;
```
thgreasi, line 21: Optional: What do you think about v ?
```suggestion
	return filePath.match(/\/([^/]+)$/)?.[1] ?? filePath.match(/\\([^\\]+)$/)?.[1] ?? filePath;
```

src/models/release-asset.ts
```diff
@@ -0,0 +1,468 @@
...
+type ReleaseAssetId =
+	| ReleaseAsset['Read']['id']
+	| {
+			asset_key: ReleaseAsset['Read']['asset_key'];
+			release: string | number | ReleaseRawVersionApplicationPair;
+	  };
+
+const MINIMUM_MULTIPART_SIZE = 5 * 1024 * 1024; // 5MB
+
+const DEFAULT_MULTIPART_CHUNK_SIZE = 5 * 1024 * 1024; // 5MB
```
thgreasi, line 61: Heads-up that these atm are `MiB` and for MB you could just do
```suggestion
const MINIMUM_MULTIPART_SIZE = 5_000_000; // 5MB

const DEFAULT_MULTIPART_CHUNK_SIZE = 5_000_000; // 5MB
```
No sure which one of the two we intend to be using

src/models/release-asset.ts
```diff
@@ -0,0 +1,468 @@
...
+				},
+			} satisfies CommitData,
+		});
+	};
+
+	const createMultipartReleaseAsset = async (
+		createParams: WriteReleaseAssetParams,
+		uploadParams: Required<Pick<UploadParams, 'chunkSize' | 'parallelUploads'>>,
+	) => {
+		const releaseAssetId = await getOrCreateReleaseAsset(createParams);
```
thgreasi, line 256: I think that you should be able to do:
```suggestion
		const releaseAssetId = (await pine.getOrCreate({
			resource: 'release_asset',
			id: { release, asset_key },
			options: {
				$select: 'id',
			},
			body: {
				asset: null,
			},
		}))!.id;
```

src/models/release-asset.ts
```diff
@@ -0,0 +1,468 @@
...
+		const fs = await import('fs/promises');
+		const fd = await fs.open(filePath, 'r');
+		try {
+			const buffer = Buffer.alloc(length);
+			const { bytesRead } = await fd.read(buffer, 0, length, offset);
+			return buffer.subarray(0, bytesRead);
+		} finally {
+			await fd.close();
+		}
+	};
```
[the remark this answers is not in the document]
thgreasi: It seems that `getFileSize` & `readFileChunk` are only needed when  `typeof asset === 'string'`, right?
So effectively `typeof asset === 'string'` is not supported in browsers afaict, right?
Moreover `typeof asset === 'string'` is the only case where we use `getMimeType`, so we can move the `mime` package import in the node specific `asset-helpers.ts` and keep the browser bundle size small, right?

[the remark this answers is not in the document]
thgreasi: eg:
```js
		async create(
			createParams: WriteReleaseAssetParams,
			{
				chunkSize = DEFAULT_MULTIPART_CHUNK_SIZE,
				parallelUploads = DEFAULT_MULTIPART_PARALLEL_UPLOAD,
			}: UploadParams = {},
		) {
			// ...

			if (typeof asset === 'string') {
				const assetsHelper = (
					require('./util/assets-helper') as typeof import('./util/assets-helper')
				).readFileChunk(opts);
				size = await assetsHelper.getFileSize(asset);
				if (size <= MINIMUM_MULTIPART_SIZE) {
					normalizedParams = {
						...restParams,
						asset: await assetsHelper.readFileChunkToFile(asset, 0, size),
					};
				}
			} else {
				// ...
			}
```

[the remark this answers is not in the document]
thgreasi: Let me know if you prefer me to push it in a separate commit.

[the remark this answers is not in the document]
otaviojacobi: @thgreasi 
> typeof asset === 'string' is not supported in browsers afaict, right?

Correct. This is basically so that for large file uploads (e.g. host apps) we can use node streaming processing for chunking and parallel chunk uploads. The stream handler from File api is way less powerfull (e.g. does not allow us to simply load into memory chunks of the whole file)

>so we can move the mime package import in the node specific asset-helpers.ts and keep the browser bundle size small, right?

Correct, I can update it to what you mention, no problem

src/models/release-asset.ts
otaviojacobi: I think it is (unnecesarily) harder to read :) but I don't mind changing if you see great improvements

src/models/release-asset.ts
otaviojacobi: Yeah, 5MiB is what we want (altough the naming with the `i` is new for me)

src/models/release-asset.ts
```diff
@@ -0,0 +1,388 @@
...
+						resource: 'release_asset',
+						body: normalizedParams!,
+					});
+				} else {
+					return await createMultipartReleaseAsset(uploadParams, {
+						chunkSize,
+						parallelUploads,
+					});
+				}
+			}
```
thgreasi, line 375: How about isolating a bit the sections to make it shallower?
The following should be pretty much the exact same amount of lines (ignoring the extra line break) and only repeat the `return await exports.get(existingReleaseAsset.id);` two times.
```suggestion
			if (existingReleaseAsset != null && !overwrite) {
				throw new ReleaseAssetAlreadyExists(
					restParams.release,
					restParams.asset_key,
				);
			}

			if (size <= MINIMUM_MULTIPART_SIZE) {
				// POST body upload upload
				if (existingReleaseAsset != null) {
					await pine.patch({
						id: existingReleaseAsset.id,
						resource: 'release_asset',
						body: { asset: normalizedParams!.asset },
					});
					return await exports.get(existingReleaseAsset.id);
				}
				return await pine.post({
					resource: 'release_asset',
					body: normalizedParams!,
				});
			}

			// Multipart upload
			if (existingReleaseAsset != null) {
				await uploadMultipartReleaseAsset(existingReleaseAsset.id, asset, {
					chunkSize,
					parallelUploads,
				});
				return await exports.get(existingReleaseAsset.id);
			}
			return await createMultipartReleaseAsset(uploadParams, {
				chunkSize,
				parallelUploads,
			});
```

tests/integration/models/release-asset.spec.ts
```diff
@@ -0,0 +1,586 @@
...
+				);
+
+				expect(overwrittenAsset).to.be.an('object');
+				expect(overwrittenAsset.asset_key).to.equal('overwrite-test.txt');
+				expect(overwrittenAsset.id).to.equal(firstAsset.id);
+			});
+		});
+
+		if (!IS_BROWSER) {
+			describe('balena.models.release.asset.upload() with file paths', function () {
```
thgreasi, line 200:
```suggestion
		describe('balena.models.release.asset.upload() with file paths', function () {
			if (IS_BROWSER) {
				it('should not be supported in browsers', async function () {
					await expectError(async () => {
						await balena.models.release.asset.upload({
							release: release.id,
							asset_key: 'browser-file-path.txt',
							asset: '/fake/file/path.txt',
						});
					}, 'File path uploads are not supported in the browser');
				});
				return;
			} else {
				// ...
			}
		});
```

src/models/release-asset.ts
```diff
@@ -0,0 +1,444 @@
...
+			},
+			{ concurrency: uploadParams.parallelUploads },
+		);
+	};
+
+	const uploadMultipartReleaseAsset = async (
+		releaseAssetId: number,
+		asset: File | string,
+		uploadParams: Required<Pick<UploadParams, 'chunkSize' | 'parallelUploads'>>,
+		onUploadProgress?: (progress: UploadProgressInfo) => void | Promise<void>,
```
thgreasi, line 186: NIT: since this is only called internally and we want the sdk to pass that parameter in all calls
```suggestion
		onUploadProgress: undefined | (progress: UploadProgressInfo) => (void | Promise<void>),
```

src/models/release-asset.ts
```diff
@@ -0,0 +1,444 @@
...
+			ETag,
+		};
+	};
+
+	const uploadParts = async (
+		asset: File | string,
+		parts: UploadPart[],
+		uploadParams: Required<Pick<UploadParams, 'chunkSize' | 'parallelUploads'>>,
+		totalSize: number,
+		onUploadProgress?: (progress: UploadProgressInfo) => void | Promise<void>,
```
thgreasi, line 161: Like below
> NIT: since this is only called internally and we want the sdk to pass that parameter in all calls

```suggestion
		onUploadProgress: undefined | (progress: UploadProgressInfo) => void | Promise<void>),
```

src/models/release-asset.ts
```diff
@@ -0,0 +1,446 @@
...
+type ReleaseAssetBeginUpload = {
+	asset: {
+		uuid: string;
+		uploadParts: UploadPart[];
+	};
+};
+
+type UploadProgressInfo = {
+	total: number;
+	received: number;
```
thgreasi, line 41: @otaviojacobi Should this be `uploaded`/`sent` instead of `received`?
PS: I initially fat fingered `received` in my comment, but edited right away to `uploaded` once I realize that I had pasted the wrong thing. I'm sorry for the confusion.

thgreasi (approved):
Drafted it to give you a chance to check my comment (which might be my fault causing confusion).
👏

src/models/release-asset.ts
otaviojacobi: @thgreasi I interpreted as `received` by the server which IMO makes sense.... I don't mind changing

PR #1566, opened 2025-08-27
Add release assets docs

otaviojacobi:
See: https://balena.fibery.io/Work/Project/Make-release-assets-publicly-available-and-maintained-1177
Change-type: patch

Resolves: # 
HQ: <url> 
See: <url> 
Depends-on: <url> 

---
##### Contributor checklist

- [ ] Includes tests
- [ ] Includes typings

### DIFF
--- a/src/models/release.ts
+++ b/src/models/release.ts
@@ -729,6 +729,204 @@ const getReleaseModel = function (
 		remove: tagsModel.remove,
 	};
 
+	/**
+	 * @namespace asset
+	 * @memberof balena.models.release
+	 */
+	const asset = {
+		/**
+		 * @summary Get all release assets for a release
+		 * @name getAllByRelease
+		 * @public
+		 * @function
+		 * @memberof balena.models.release.asset
+		 *
+		 * @param {String|Number|Object} commitOrIdOrRawVersion - release commit (string) or id (number) or an object with the unique `application` (number or string) & `rawVersion` (string) pair of the release
+		 * @param {Object} [options={}] - extra pine options to use
+		 * @fulfil {Object[]} - release assets
+		 * @returns {Promise}
+		 *
+		 * @example
+		 * balena.models.release.asset.getAllByRelease(123).then(function(assets) {
+		 * 	console.log(assets);
+		 * });
+		 *
+		 * @example
+		 * balena.models.release.asset.getAllByRelease('7cf02a6').then(function(assets) {
+		 * 	console.log(assets);
+		 * });
+		 *
+		 * @example
+		 * balena.models.release.asset.getAllByRelease({ application: 456, raw_version: '1.2.3' }).then(function(assets) {
+		 * 	console.log(assets);
+		 * });
+		 */
+		getAllByRelease: assetsModel.getAllByRelease,
+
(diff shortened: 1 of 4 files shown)


src/models/release.ts
```diff
@@ -729,6 +729,153 @@ const getReleaseModel = function (
...
+
+		/**
+		 * @summary Upload a release asset
+		 * @name upload
+		 * @public
+		 * @function
+		 * @memberof balena.models.release.asset
+		 *
+		 * @param {Object} uploadParams - upload parameters
+		 * @param {String|File} uploadParams.asset - asset file path (Node.js) or File object (browser)
```
thgreasi, line 822: Isn't the File approach working on both browsers & node?
Maybe b/c of that benefit, File should be the primarily advertised way of using this method
```suggestion
		 * @param {String|File} uploadParams.asset - asset file path (Node.js) or File object (Node.js & browser)
```

src/models/release.ts
```diff
@@ -729,6 +729,153 @@ const getReleaseModel = function (
...
+		 * @param {Object} [options={}] - upload options
+		 * @param {Number} [options.chunkSize=5242880] - chunk size for multipart uploads (5MiB default)
+		 * @param {Number} [options.parallelUploads=5] - number of parallel uploads for multipart
+		 * @param {Boolean} [options.overwrite=false] - whether to overwrite existing asset
+		 * @param {Function} [options.onUploadProgress] - callback for upload progress
+		 * @fulfil {Object} - uploaded release asset
+		 * @returns {Promise}
+		 *
+		 * @example
+		 * // Upload from file path (Node.js)
```
thgreasi, line 834: I guess adding a File example might not be as easy/need too much extra code, right?

src/models/release.ts
```diff
@@ -729,6 +729,153 @@ const getReleaseModel = function (
...
+		 * @fulfil {Object[]} - release assets
+		 * @returns {Promise}
+		 *
+		 * @example
+		 * balena.models.release.asset.getAllByRelease(123).then(function(assets) {
+		 * 	console.log(assets);
+		 * });
+		 *
+		 * @example
+		 * balena.models.release.asset.getAllByRelease('7cf02a6').then(function(assets) {
```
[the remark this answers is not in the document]
otaviojacobi: I added this one, there is one with the commit based already :)

src/models/release.ts
otaviojacobi: Yes :) good catch

src/models/release.ts
```diff
@@ -729,6 +729,204 @@ const getReleaseModel = function (
...
+		 * 	asset_key: file.name,
+		 * 	release: 123
+		 * }).then(function(asset) {
+		 * 	console.log('File uploaded from browser:', asset);
+		 * });
+		 *
+		 * @example
+		 * // Upload with overwrite option
+		 * balena.models.release.asset.upload({
+		 * 	asset: '/path/to/logo.png',
```
thgreasi, line 897: Let's put the two examples that are using a string path next to each other since they are node only

thgreasi (approved):

PR #1578, opened 2025-11-11
os.getSupportedOsUpdateVersions: Add support for including ESR versions

thgreasi:
Change-type: minor
See: https://balena.fibery.io/Work/Project/2009

### DIFF
--- a/tests/integration/models/os.spec.ts
+++ b/tests/integration/models/os.spec.ts
@@ -1028,6 +1028,73 @@ describe('OS model', function () {
 				expect(versions).to.deep.equal(sortedVersions);
 			});
 
+			it('should only include non-ESR OS releases by default', async () => {
+				const { current, recommended, versions } =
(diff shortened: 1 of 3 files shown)


tests/integration/models/os.spec.ts
```diff
@@ -1028,6 +1028,57 @@ describe('OS model', function () {
...
+				expect(bSemver.gt(recommended, '2000.0.0')).to.be.true;
+				expect(versions.filter((v) => bSemver.lt(v, '2000.0.0'))).to.deep.equal(
+					[],
+				);
+				expect(
+					versions.filter((v) => bSemver.gt(v, '2000.0.0')),
+				).to.have.length.greaterThan(0);
+			});
+
+			it('should include both default & ESR OS releases when the respective flag is used', async () => {
```
myarmolinsky: nit: 
```suggestion
			it('should include both default & ESR OS releases when osType is null', async () => {
```

Issue #1576, opened 2025-10-15
Should rename device.setSupervisorRelease to device.pinToSupervisorRelease


thgreasi:
Should change `device.setSupervisorRelease()` -> `device.pinToSupervisorRelease()`
That's for consistency w/ `device.pinToRelease`

PR #1580, opened 2025-11-11
Release balena-sdk v23

thgreasi:
Resolves: #1576
Resolves: #1583

Change-type: major
See: https://balena.fibery.io/Work/Project/2196
See: https://balena.fibery.io/Work/Project/1704

### DIFF
--- a/src/models/device.ts
+++ b/src/models/device.ts
@@ -1767,7 +1767,7 @@ const getDeviceModel = function (
 		 * @see {@link balena.models.device.get} for an example on selecting the `overall_progress` field.
 		 *
 		 * @param {String|Number} uuidOrId - device uuid (string) or id (number)
-		 * @fulfil {Number|Null} - device progress
+		 * @fulfil {Number|null} - device progress
 		 * @returns {Promise}
 		 *
 		 * @example
--- a/src/models/os.ts
+++ b/src/models/os.ts
@@ -903,7 +903,7 @@ const getOsModel = function (
 	 * @param {String} currentVersion - semver-compatible version for the starting OS version
 	 * @param {Object} [options] - Extra options to filter the OS releases by
 	 * @param {Boolean} [options.includeDraft=false] - Whether pre-releases should be included in the results
-	 * @param {String|Null} [options.osType=Null] - Can be one of 'default', 'esr' or null which includes all types
+	 * @param {String|null} [options.osType=null] - Can be one of 'default', 'esr' or null which includes all types
 	 * @fulfil {Object[]|Object} - An array of OsVersion objects when a single device type slug is provided,
 	 * or a dictionary of OsVersion objects by device type slug when an array of device type slugs is provided.
 	 * @fulfil {Object} - the versions information, of the following structure:
--- a/DOCUMENTATION.md
+++ b/DOCUMENTATION.md
@@ -3804,7 +3804,7 @@ in case that you need to retrieve more device fields than just the progress.
 **Kind**: static method of [<code>device</code>](#balena.models.device)  
 **Summary**: Get the progress of a device  
 **Access**: public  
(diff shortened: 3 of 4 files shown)


kb2ma:
I have browsed through the commits related to Supervisor managed OS updates. It looks as expected. Next I will try using it.

src/models/os.ts
```diff
@@ -921,7 +902,7 @@ const getOsModel = function (
 	 * @param {String} currentVersion - semver-compatible version for the starting OS version
 	 * @param {Object} [options] - Extra options to filter the OS releases by
 	 * @param {Boolean} [options.includeDraft=false] - Whether pre-releases should be included in the results
-	 * @param {Boolean|Null} [options.osType='default'] - Can be one of 'default', 'esr' or null to include all types
+	 * @param {Boolean|Null} [options.osType=Null] - Can be one of 'default', 'esr' or null which includes all types
```
kb2ma: Type should be {String|Null}

src/models/device.ts
```diff
@@ -2255,28 +2070,112 @@ const getDeviceModel = function (
 		},
 
 		/**
-		 * @summary Start an OS update on a device
-		 * @name startOsUpdate
+		 * @summary Mark a specific device to be updated to a particular OS release
+		 * @name pinToOSRelease
```
kb2ma: Use CamelCase like other functions. So "Os" rather than "OS". Also address other instances in this function definition.

src/models/device.ts
```diff
@@ -2255,28 +2070,112 @@ const getDeviceModel = function (
...
 		 * @public
 		 * @function
 		 * @memberof balena.models.device
 		 *
-		 * @param {String|String[]} uuidOrUuids - full device uuid or array of full uuids
-		 * @param {String} targetOsVersion - semver-compatible version for the target device
+		 * @param {String|String[]|Number|Number[]} uuidOrIdOrArray - device uuid (string) or id (number) or array of full uuids or ids
+		 * @param {String} osVersionOrId - the raw version of a OS release (string) or id (number)
 		 * Unsupported (unpublished) version will result in rejection.
 		 * The version **must** be the exact version number, a "prod" variant and greater than the one running on the device.
```
kb2ma: We'll need to update this description from "greater than" -> "greater than or equal to" to allow for pin back to current.

The _checkOsUpdateTarget() -> balena-hup-action-utils getHUPActionType() check already is OK though ([code](https://github.com/balena-io-modules/balena-hup-action-utils/blob/a8e1ad89145d45f7462f856a3b645b12ad7881b8/lib/index.ts#L146)).

kb2ma:
When I run `device.getWithServiceDetails()`, the response shows the individual app services:
```
...
  is_undervolted: false,
  current_services_by_app: [Object: null prototype] {
    'ken8/otel-netio-amd64': [Object: null prototype] {
      'node-exporter': [Array],
      'otel-collector': [Array],
      'netio-logger': [Array]
    }
  }
```

However, when I run `application.getWithDeviceServiceDetails()`, I don't see the individual services for each device. Is this difference to be expected?
```
...
      is_undervolted: false,
      current_services_by_app: [Object: null prototype]
    }
],
...
```

package.json
```diff
@@ -123,7 +123,7 @@
     "abortcontroller-polyfill": "^1.7.1",
     "balena-auth": "^6.0.1",
     "balena-errors": "^4.9.0",
-    "balena-hup-action-utils": "~6.2.0",
+    "balena-hup-action-utils": "~7.0.0",
```
otaviojacobi: Not sure about what this impacts/changes but I think @kb2ma is reviewing this part?

kb2ma: @otaviojacobi That work rejects old versions of balenaOS, including pre-v2.x. It was merged with balena-io-modules/balena-hup-action-utils#52 .

PR #1586, opened 2025-12-30
Block OS Upgrades from OS < 2.14.0 and target OS < 2.16.0

thgreasi:
This is not a major since we have already
dropped support for balenaOS < 2.14.0
in balena-sdk v21.0.0 with:
https://github.com/balena-io/balena-sdk/pull/1505/changes/6e284001e5a06cf71d8456837a7f6ab8681d0d5b

Depends-on: https://github.com/balena-io-modules/balena-hup-action-utils/pull/52
Change-type: patch
See: https://balena.fibery.io/Work/Project/2183
See: https://balena.fibery.io/Work/Project/2033
See: https://github.com/balena-os/balenahup/pull/422

Resolves: # 
HQ: <url> 
See: <url> 
Depends-on: <url> 

---
##### Contributor checklist

- [ ] Includes tests
- [ ] Includes typings

### DIFF
--- a/tests/integration/models/device.spec.ts
+++ b/tests/integration/models/device.spec.ts
@@ -3630,7 +3630,8 @@ describe('Device Model', function () {
 			it('should throw when the device is offline', () => {
 				[
 					['Resin OS 1.21.0', '', '1.28.0'],
-					['Resin OS 1.30.1', '', '2.5.0+rev1'],
+					['Resin OS 1.30.1', '', '2.16.0+rev1'],
+					['Resin OS 2.14.0+rev1', '', '2.16.0+rev1'],
 					['balenaOS 2.26.0+rev1', 'prod', '2.29.2+rev1.prod'],
 				].forEach(function ([osVersion, osVariant, targetOsVersion]) {
 					return expect(() => {
@@ -3693,7 +3694,7 @@ describe('Device Model', function () {
 					['balenaOS 2.26.0+rev1', 'dev'],
 					['balenaOS 2.28.0+rev1', 'dev'],
 				].forEach(function ([osVersion, osVariant]) {
-					return expect(() => {
+					expect(() => {
 						_checkOsUpdateTarget(
 							{
 								uuid,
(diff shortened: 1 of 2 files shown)


thgreasi:
All your comments have valid points about dropping tests for legacy & soon to be announced as deprecated OS versions, but all these changes are not related to the changes the the hup-action-utils bump bring, so they should better be in a separate PR, and imo after we officially announce the deprecation of these OS versions.


> Questions about some tests in `device.spec.ts`.
> The test below looks obsolete since it checks for host OS version < 2.12.
https://github.com/balena-io/balena-sdk/blob/8b3e2d27f7aa5ad52eb3c382565b189d5ad3d39e/tests/integration/models/device.spec.ts#L2914

This is a pre-existing test which captures how the API responds in such requests. Since it's not related to the HUP changes, I think we should leave it as for now.

> The validity of this test is marginal. I understand it's checking for a dev version, but it's using an invalid OS version. The same is true for the Supervisor version, which must be >= v7, correct?
https://github.com/balena-io/balena-sdk/blob/8b3e2d27f7aa5ad52eb3c382565b189d5ad3d39e/tests/integration/models/device.spec.ts#L3442

The API still supports enabling local mode for such an old device, so I'm in favor of capturing the current behavior in an sdk test.
PS: Resin OS 2.7.8+rev1 is a valid OS version afaik and we officially also haven't yet announced it to be EOL.

> Here is another example, checking OS update target, where one of the source/target pairs is for OS v1.
https://github.com/balena-io/balena-sdk/blob/8b3e2d27f7aa5ad52eb3c382565b189d5ad3d39e/tests/integration/models/device.spec.ts#L3630

I would argue that it doesn't hurt testing what error the user would be getting when passing an offline device that's attempts an unsupported OS version.

> The test below really is checking for 'v1', not '<v1.30.1'. Is it necessary to run this test for beaglebone-black, and the previous test for RPi3?
https://github.com/balena-io/balena-sdk/blob/8b3e2d27f7aa5ad52eb3c382565b189d5ad3d39e/tests/integration/models/device.spec.ts#L3778

You are right, we no longer need to test for both DTs. Removed the bbb test.



> Questions about some tests in `os.spec.ts`.
> The test below should reject for 'v1' not '<= v1.2.0'.
https://github.com/balena-io/balena-sdk/blob/8b3e2d27f7aa5ad52eb3c382565b189d5ad3d39e/tests/integration/models/os.spec.ts#L1276

This error originates from the API, so until we announce that we dropped support for legacy OS versions and actively block it in the API, I would be in favor of keeping this test as is.
https://github.com/balena-io/open-balena-api/blob/master/src/features/device-config/device-config.ts#L40-L44

> The test below is to configure a v1 image. That's obsolete, correct?
https://github.com/balena-io/balena-sdk/blob/8b3e2d27f7aa5ad52eb3c382565b189d5ad3d39e/tests/integration/models/os.spec.ts#L1294

Like above I would argue to keep it as is and change/drop it after the API drops support for it, especially since this is unrelated to this PR.

PR #1630, opened 2026-06-16
Switch from custom `Dictionary` type to built in `Record` typing

Page-:
This reduces the number of custom typings we maintain/need to know about

Change-type: patch

Resolves: # 
HQ: <url> 
See: <url> 
Depends-on: <url> 

---
##### Contributor checklist

- [ ] Includes tests
- [ ] Includes typings

### DIFF
--- a/src/models/balenaos-contract.ts
+++ b/src/models/balenaos-contract.ts
@@ -1,4 +1,3 @@
-import type { Dictionary } from '../../typings/utils';
 import type { Contract } from '../types/contract';
 
 // Hardcoded host OS contract, this should be moved to the Yocto build process with meta-balena.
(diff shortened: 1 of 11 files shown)


src/models/balenaos-contract.ts
```diff
@@ -43,7 +42,7 @@ export const BalenaOS: Contract = {
 			`To provision {{deviceType.name}}, follow the instructions using our [Jetson Flash tool](https://github.com/balena-os/jetson-flash/blob/master/docs/{{deviceType.slug}}.md) to make the process more streamlined.`,
 		],
 		usbMassStorage: [
-			`{{#each deviceType.partials.instructions}}{{{this}}} 
+			`{{#each deviceType.partials.instructions}}{{{this}}}
```
otaviojacobi, line 45: Is this a linter change?

Page-: My editor automatically removes trailing spaces.. (a common thing to do) It shouldn't be necessary surely?

thgreasi: I think that this is necessary for adding a space between the sentences that #each iterates over.
I'm guessing that w/o it the result would be something like:
> This is sentence A.This is sentence B.This is C.

thgreasi: nvm, I tested it and it's fine

thgreasi (approved):

